PR #7, opened 2025-02-20
Adds publishing to pypi

elena-kolevska:
Publishing will require adding the PyPI publishing token under `secrets.PYPI_UPLOAD_PASS`.

I think we should be safe to remove the setup.py and setup.cfg files now that we have a pyproject.toml.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,7 +35,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install setuptools wheel twine build
-      - name: Build and publish Dapr Durable Task Python
+      - name: Build and publish Dapr Agents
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,43 @@
...
+          fetch-depth: 0
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.11
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install setuptools wheel twine build
+      - name: Build and publish Dapr Durable Task Python
```
yaron2: Change to Dapr Agents

PR #8, opened 2025-02-21
Add Dapr's Conversation API as LLM inference provider

yaron2:
^ Title

### DIFF
--- a/dapr_agents/llm/dapr/chat.py
+++ b/dapr_agents/llm/dapr/chat.py
@@ -49,9 +49,6 @@ class DaprChatClient(DaprInferenceClientBase, ChatClientBase):
         # Generate the prompt template from the Prompty instance
         prompt_template = Prompty.to_prompt_template(prompty_instance)
 
-        # Extract the model configuration from Prompty
-        model_config = prompty_instance.model
-
         # Initialize the DaprChatClient based on the Prompty model configuration
         return cls.model_validate({
             'timeout': timeout,
--- a/dapr_agents/llm/dapr/client.py
+++ b/dapr_agents/llm/dapr/client.py
@@ -61,7 +61,6 @@ class DaprInferenceClientBase(LLMClientBase):
         """
         Initializes and returns the Dapr Inference client.
         """
-        config: DaprInferenceClientConfig = self.config
         return DaprInferenceClient()
     
     @classmethod


yaron2:
> In the meantime I opened a PR with Conversation API support in the python SDK ([dapr/python-sdk#787](https://github.com/dapr/python-sdk/pull/787)). If we can get this approved soon I believe it will be easier to just use DaprClient, it handles all kinds of things for us out of the box (tls, retries, healthcheck, all supported env variables...).

Can we get this is for 1.15?

elena-kolevska:
Yes, it's last moment, but we're still on time. The PR is ready for review (with full test coverage and docs). I'd like to have at least one more person give it a review and get to an approval.

yaron2:
Updated with the usage of the Dapr Client

dapr_agents/llm/dapr/client.py
```diff
@@ -0,0 +1,87 @@
...
+        """
+        Returns the appropriate configuration for the Dapr Conversation API.
+        """
+        return DaprInferenceClientConfig()
+
+    def get_client(self) -> DaprInferenceClient:
+        """
+        Initializes and returns the Dapr Inference client.
+        """
+        config: DaprInferenceClientConfig = self.config
```
elena-kolevska: The config var doesn't seem to be used.

dapr_agents/llm/dapr/chat.py
```diff
@@ -0,0 +1,155 @@
...
+            DaprChatClient: An instance of DaprChatClient configured with the model settings from the Prompty source.
+        """
+        # Load the Prompty instance from the provided source
+        prompty_instance = Prompty.load(prompty_source)
+
+        # Generate the prompt template from the Prompty instance
+        prompt_template = Prompty.to_prompt_template(prompty_instance)
+
+        # Extract the model configuration from Prompty
+        model_config = prompty_instance.model
```
elena-kolevska: Not used

dapr_agents/llm/dapr/client.py
yaron2: Fixed

dapr_agents/llm/dapr/chat.py
yaron2: Fixed

PR #14, opened 2025-02-28
Quickstarts and fixes

bibryam:
- Initial version of quickstarts
- Added automatic validation with Mechanical markdown

Can run validation locally with `make validate-quickstarts`. This is still not set up to run in CI because it needs an OpenAI token.


quickstarts/01-hello-world/03_reason_act.py
```diff
@@ -0,0 +1,23 @@
+from floki import tool, ReActAgent
```
msfussell:
```suggestion
from dapr_agents import tool, ReActAgent
```

msfussell: All of these should change.

quickstarts/01-hello-world/README.md
```diff
@@ -0,0 +1,196 @@
+# Hello World with Dapr Agents
+
+This quickstart provides a hands-on introduction to Dapr Agents through simple examples. You'll learn the fundamentals of working with LLMs, creating basic agents, implementing the ReAct pattern, and setting up simple workflows - all in less than 20 lines of code per example.
+
+## Prerequisites
+
+- Python 3.10 (recommended)
+- pip package manager
+- OpenAI API key
```
msfussell, line 9: It would be nice to say you can have other LLMs keys rather than just OpenAI. Given the Conversation API integration this should be possible

quickstarts/01-hello-world/03_reason_act.py
bibryam: I'm waiting for some fixes to make the quickstarts work, then I'll replace old references

quickstarts/01-hello-world/README.md
elena-kolevska: @bibryam I was planning on adding this. Let me know if you prefer doing it yourself.

yaron2:
@bibryam we need the commits to have verified signatures

PR #21, opened 2025-03-05
Fixes roundrobin workflow orchestrator and aligns all three orchestrators

elena-kolevska:


cookbook/workflows/multi_agents/basic_lotr_agents_as_actors/config.yaml
```diff
@@ -9,4 +9,4 @@ spec:
       endpointAddress: http://localhost:9411/api/v2/spans
   features:
     - name: SchedulerReminders
-      enabled: false
\ No newline at end of file
+      enabled: true
```
elena-kolevska, line 12: This is the default in Dapr now, and much more performant, so it would be better to stick with it.
We can still keep this file for a while, in case someone is running the example with dapr 1.14 (when this option was needed to use reminders with scheduler).

yaron2:
@elena-kolevska commits need to be verified

PR #33, opened 2025-03-11
Remove sandbox refs

marcduiker:


docs/home/installation.md
```diff
@@ -14,13 +14,13 @@ pip install dapr-agents
 ### Remotely from GitHub
 
 ```bash
-pip install git+https://github.com/dapr-sandbox/dapr-agents.git
+pip install git+https://github.com/dapr/dapr-agents.git
```
elena-kolevska: We should be able to add installing from pypi now :) https://pypi.org/project/dapr-agents/

PR #51, opened 2025-03-13
fix workflow orchestrator generate_plan activity

famarting:
Testing the dapr workflow quickstart I found this error

```
Activity task #2 failed: Task 'generate_plan' execution failed: issubclass() arg 1 must be a class
```

### DIFF
--- a/dapr_agents/workflow/orchestrators/llm/orchestrator.py
+++ b/dapr_agents/workflow/orchestrators/llm/orchestrator.py
@@ -196,7 +196,7 @@ class LLMOrchestrator(OrchestratorServiceBase):
         return agent_list
     
     @task(description=TASK_PLANNING_PROMPT)
-    async def generate_plan(self, task: str, agents: str, plan_schema: str) -> List[PlanStep]:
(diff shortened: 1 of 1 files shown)


yaron2 (approved):

yaron2:
@Cyb3rWard0g for review

Cyb3rWard0g (commented):
Hey @famarting ! thank you for testing the project and sharing your proposal. I noticed you removed self from the method definition. Since this is an instance method, it should have self as the first parameter, otherwise, it won't have access to instance attributes like self.state or self.timeout.

I don’t think the error you’re seeing is related to this change. Could you share more logs or details on why you think this is the issue? If there’s something specific pointing to this, let’s dig into it!

famarting:
Hey @Cyb3rWard0g I think I'll close this issue, I was testing dapr-agents for the first time against Diagrid Catalyst https://catalyst.diagrid.io/ and I was hitting this issue, but I think there are other issues that were causing this, I still don't know 100%, I'll let you know. For now I think we can close this issue.

PR #54, opened 2025-03-18
Removes additional address arguments for Dapr client init

elena-kolevska:
Relies on env variables instead

### DIFF
--- a/dapr_agents/workflow/agentic.py
+++ b/dapr_agents/workflow/agentic.py
@@ -8,7 +8,6 @@ from typing import Any, Optional, Union, Dict, Type, List
 from fastapi import Depends, Request, Response, HTTPException
 from dapr.clients import DaprClient
 import tempfile
-import json
 import threading
 import inspect
 import logging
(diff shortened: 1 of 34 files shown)


quickstarts/05-multi-agent-workflow-actors/dapr-llm-catalyst.yaml
```diff
@@ -0,0 +1,54 @@
+# https://docs.dapr.io/developing-applications/local-development/multi-app-dapr-run/multi-app-template/#template-properties
+version: 1
+common:
+  resourcesPath: ./components
+  logLevel: info
+  appLogDestination: console
+  daprdLogDestination: console
+
+apps:
+- appId: HobbitApp
```
famarting: I think this needs to be `appID`

elena-kolevska: Ah, yeah, you had mentioned this before, I'll update!

famarting (approved):
lgtm

dapr_agents/workflow/agentic.py
Cyb3rWard0g, line 1: I wonder if we can use the dapr_client from the `DaprPubSub` class which is defined in the `dapt_agents/messaging/dapr.py`. right?

Cyb3rWard0g (approved):
lgtm! Thank you @elena-kolevska !

dapr_agents/workflow/agentic.py
elena-kolevska: They don't depend on the same base class though, no?

Cyb3rWard0g: The `DaprPubSub` brings the messaging capabilities to the WorkflowApp.

```python
class AgenticWorkflowService(WorkflowApp, DaprPubSub):
    """
    A service class for managing agentic workflows, extending `WorkflowApp`.
    Handles agent interactions, workflow execution, messaging, and metadata management.
    """
```

Since you are adding a `dapr_client` as an attribute of the DaprPubSub class, then we could reuse it right?

PR #57, opened 2025-03-21
Feat/mkdocs validator

CasperGN:
Add validation step to ensure that
- upon changes in `docs/**` and `mkdocs.yaml` we trigger
- mkdocs rendering for validating output
- only deploy the docs if files changed and docs rendered valid

We use the same setup to render all `docs/**` towards backstage internally. It serves as a quick good validator that rendering and `nav` works as expected.


yaron2 (approved):
LGTM

Issue #55, opened 2025-03-21
Bug: Task 'generate_plan' execution failed: issubclass() arg 1 must be a class


yaron2:
When running https://github.com/dapr/dapr-agents/blob/main/quickstarts/05-multi-agent-workflow-dapr-workflows/README.md, the following error occurs (Python 3.10):

```
== APP - workflow-llm == 2025-03-21 16:49:16.004 durabletask-client INFO: Instance 'NNN' failed: [TaskFailedError] YYY: Activity task #2 failed: Task 'generate_plan' execution failed: issubclass() arg 1 must be a class
```

yaron2:
History: https://github.com/dapr/dapr-agents/pull/51

CasperGN:
@yaron2 on the latest version I'm unable to reproduce the above error however, I consistently get:
```bash
== APP - ElfApp == INFO:dapr_agents.llm.openai.client.base:Initializing OpenAI client...
== APP - WorkflowApp == Error starting service: 1 validation error for LLMOrchestrator
== APP - WorkflowApp == service_port
== APP - WorkflowApp ==   Field required [type=missing, input_value={'name': 'LLMOrchestrator...ame': 'LLMOrchestrator'}, input_type=dict]
== APP - WorkflowApp ==     For further information visit https://errors.pydantic.dev/2.10/v/missing
...
== APP - ElfApp == Error starting service: 1 validation error for AssistantAgent
== APP - ElfApp == service_port
== APP - ElfApp ==   Field required [type=missing, input_value={'name': 'Legolas', 'role...ervice_name': 'Legolas'}, input_type=dict]
== APP - ElfApp ==     For further information visit https://errors.pydantic.dev/2.10/v/missing
```

For all agents in the [05-multi-agent-workflow-dapr-workflows](https://github.com/dapr/dapr-agents/tree/main/quickstarts/05-multi-agent-workflow-dapr-workflows).

My `.env` file:
```bash
AZURE_OPENAI_ENDPOINT=my_custom_endpoint
OPENAI_API_KEY=my_very_secret_key
```
on the following setup:
```bash
$ python --version
Python 3.10.16
```
```bash
$ pip freeze                     
aiohappyeyeballs==2.6.1
aiohttp==3.11.14
aiosignal==1.3.2
annotated-types==0.7.0
anyio==4.9.0
async-timeout==5.0.1
attrs==25.3.0
azure-core==1.32.0
azure-identity==1.19.0
certifi==2025.1.31
[...]

PR #67, opened 2025-03-28
Fix Structured Output Validation Across Python Versions

Cyb3rWard0g:
This PR improves structured response handling across Python versions by addressing type introspection bugs that impacted task output validation. It ensures correct behavior when return types like List[Model] or List[dict[str, Any]] are used in workflow tasks.

##  Key Fixes
- Replaced `isinstance(..., Iterable)` check with `get_origin(...)` inside `normalize_iterable_format()` for consistent iterable detection.
- Ensured `List[Model]` types are correctly wrapped into generated Pydantic models for validation and schema generation.
- Resolved actual response model with `resolve_response_model()` before validating LLM outputs.
- Hardened `resolve_all_pydantic_models()` by:
- Guarding all `issubclass()` calls with `isinstance(..., type)`
- Catching TypeError to prevent crashes from invalid generic types like Dict or List[dict]

### DIFF
--- a/dapr_agents/llm/utils/response.py
+++ b/dapr_agents/llm/utils/response.py
@@ -1,10 +1,11 @@
-from typing import Union, Dict, Any, Type, Optional, Iterator, Literal, get_args
-from dapr_agents.llm.utils import StreamHandler, StructureHandler
-from dataclasses import is_dataclass, asdict
-from dapr_agents.types import ChatCompletion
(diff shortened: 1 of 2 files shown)


yaron2 (approved):

Cyb3rWard0g:
fix https://github.com/dapr/dapr-agents/issues/55

PR #78, opened 2025-04-11
improve register agent

famarting:
this PR tries to make the initial register agentic system process more seamless and less noisy.

Currently it logs too many errors and agents update the statestore even if they are already registered.

The changes consist of:
- reduce log verbosity, remove some log and change others to debug
- rename function to register_agent and change its signature
- added check to verify if the agent name and metadata already exist in the statestore


Cyb3rWard0g (commented):
LGTM!

famarting:
we need unit or integration tests where we can test all this functions and logic in isolation

yaron2:
> we need unit or integration tests where we can test all this functions and logic in isolation

Integration tests would be the more fitting choice

yaron2 (approved):

PR #72, opened 2025-04-05
Integrate MCP Client + Enable Async Tool Execution for Agent Framework

Cyb3rWard0g:
This PR introduces full support for the [Model Context Protocol (MCP)](https://modelcontextprotocol.io/introduction) via a new MCPClient, and updates the agent tool system to support async-first execution. These changes enable seamless integration with external MCP tool servers and ensure that agents can execute tools over async transports like HTTP, SSE, or subprocess-based stdio.

## ✅ Key Changes
- New MCP Client
  - Added MCPClient with support for stdio and sse transport layers.
  - Dynamically loads tools and prompts from remote MCP servers.
  - Wraps MCP tools as `AgentTool` instances with runtime session management.
  - Provides ephemeral tool execution using async context managers.
  - Includes prompt conversion utilities for internal message format.
- Async Tool Execution Support
  - Updated `AgentTool` to support both sync (run()) and async (arun()) execution paths.
  - Automatically detects async functions and provides flexible error handling.
  - `AgentToolExecutor.run_tool()` now supports seamless execution of both sync/async tools.
  - Improved validation, logging, and fallback logic for mixed tool environments.

PR #69, opened 2025-03-29
Feat/k8s deployment

CasperGN:
## What's in this PR

This PR addresses #63 with an example on how to run the [05-multi-agent-workflow-dapr-workflows](https://github.com/dapr/dapr-agents/tree/main/quickstarts/05-multi-agent-workflow-dapr-workflows) inside Kubernetes.

The PR packs 
1. An install script which covers everything needed to run the setup
2. Manifests for deploying into Kubernetes
3. Readme for documentation of using

_Images are multi-staged with distroless as the runtime layer._

This PR is co-authored with @jgmartinez

PR #82, opened 2025-04-12
Quickstarter: Feat/mcp integration GitHub

CasperGN:
Following #72 this PR builds on top of the Kubernetes based setup in #69 and introduces a single-agent using [GitHub MCP Server](https://github.com/github/github-mcp-server/tree/main).

## Includes

- Markdown linting updates of all README files in quickstarts folders
- Install script for easy run and setup of local `kind` cluster
- Readme with instructions for running without the script
- Simple GitHub Issue Agent that will fetch an issue and propose changes

The intended use-case would be to pick up the proposal with another agent that runs a code-specific model such as `Claude` or similar.

The Instructions for the agent is somewhat elaborate however, I've been unable to gain consistent successful results without being exceptionally explicit. This may change as the the ecosystem matures.

## Prerequisites

This PR requires #69 to be merged as it re-uses some of the files in order to not duplicate too much.


CasperGN:
@Cyb3rWard0g can you see if this somewhat captures the MCP setup albeit a single agent and somewhat trivial?

PR #94, opened 2025-04-22
Agent memory: add support for all Dapr state stores

yaron2:
Currently, the `ConversationDaprStateMemory` only supports Redis with the JSON module, and also requires a pre-built schema.

The current change changes the data structure to support all Dapr state stores.


Cyb3rWard0g (approved):
LGTM 🔥

PR #101, opened 2025-04-24
Fix grpc connection lifecycle with daprclient

yaron2:
When using the Dapr Memory Store, the DaprClient will timeout when unused for a period of time, with the gRPC connection closing on idleness. This change uses the documented short-lived operation programming model.


Cyb3rWard0g (approved):
I tested it locally and it works. It does not time out anymore and it runs for a long time :) Client does not close anymore. 

image

Issue #81, opened 2025-04-11
Dapr Agents calling external tools using Dapr HTTP Endpoint


bibryam:
### Context  
Dapr agents should support calling external tools via HTTP using Dapr's built-in **[Invoke Non-Dapr Endpoints](https://docs.dapr.io/developing-applications/building-blocks/service-invocation/howto-invoke-non-dapr-endpoints/)** capability. This allows agents to benefiting from Dapr’s core features.

By using this pattern, agents gain:
- **Resiliency policies** (e.g. retries, timeouts)
- **Observability** through built-in tracing and metrics
- **Security controls** via endpoint scoping
- **Middleware support** for features like authentication or rate limiting

CasperGN:
@bibryam at the risk of misunderstanding you, I think this is already doable.

**Context**
Assume a conversational `Agent` with `Chainlit`. This agent has the following code:
```python
import os
import chainlit as cl
from dapr_agents import Agent
from dapr_agents.memory import ConversationDaprStateMemory
from dotenv import load_dotenv
from tool import tools
import logging

load_dotenv()
logging.basicConfig(level=logging.INFO)
logger = logging.getLogger(__name__)

agent = Agent(
    name="agent",
    role="Content Expert", 
    goal="Answer questions",
    memory=ConversationDaprStateMemory(
        store_name="agentstatestore", session_id="conversationagent"
    ),
    llm=llm,
    tools=tools,
    instructions = [
        "You are a helpful assistant.",
        "You are a content expert.",
        ...
    ],
)

@cl.on_chat_start
async def main():
    await cl.Message(content="Connected to Chainlit!").send()

@cl.on_message
async def on_message(message: cl.Message):

    result = await agent.run(message.content)
    await cl.Message(content=result).send()
```

With the following `tool.py`:

```python
from dapr_agents import tool
from dapr_agents.types import AgentToolExecutorError
from pydantic import BaseModel, Field
import logging

logging.basicConfig(level=logging.INFO)
logger = logging.getLogger(__name__)

class TriggerHeimdallGenerateSchema(BaseModel):
    prompt: str = Field(description="Prompt to generate Heimdall workload from.")


@tool(args_model=TriggerHeimdallGenerateSchema)
def trigger_heimdall_generate(prompt: str) -> str:
    """Trigger generation of Heimdall workload based off the prompt."""
    import requests
    import time

    task_payload = {"task": prompt}
[...]

PR #107, opened 2025-04-28
Fix/81 dapr http endpoint

CasperGN:
## Calling External tools through Dapr

This PR seeks to address #81 with a stable abstraction of HTTP requests that are easily imported into an agent.

Example

```python
client = DaprHTTPClient(
    http_endpoint="http://echoserver.default.svc.cluster.local:80",
    method="/May",
    headers={"Client": "007", "Endpoint": "http_endpoint"},
)


@tool
def http_post_request(data: str) -> tuple[int, str]:
    """Make an HTTP POST request."""
    response = client.post(
        payload={"task": data},
    )
    return response

@tool
def http_get_request(method: str) -> tuple[int, str]:
    """Make an HTTP GET request."""
    response = client.get(method=method)
    return response

async def main():
    try:
        agent_service = AssistantAgent(
            tools=[http_get_request, http_post_request],
            ...
```

This PR is still WIP.

## TODO

- Address arbitrary inputs where consumers may append `/` after the `http_endpoint` or similar.
- Better logging
- Implement tracing

### DIFF
--- a/dapr_agents/tool/http/client.py
+++ b/dapr_agents/tool/http/client.py
@@ -123,7 +123,7 @@ class DaprHTTPClient(BaseModel):
             url = f"{self._base_url}/{self.dapr_http_endpoint}/method{self.method if method == '' else method}"
         elif self.http_endpoint != "":
             # FQDN URL
-            url = f"{self.http_endpoint}/method{self.method if method == '' else method}"
+            url = f"{self._base_url}/{self.http_endpoint}/method{self.method if method == '' else method}"
         elif endpoint != "":
             # Fallback to default
             url = f"{self._base_url}/{self.endpoint}/method{self.method if method == '' else method}"
--- a/dapr_agents/tool/http/client.py
+++ b/dapr_agents/tool/http/client.py
@@ -1,0 +1,149 @@
+from typing import Dict, Optional, Set, Any, Union
+import logging
+import requests
+
+from pydantic import BaseModel, Field, PrivateAttr
+from dapr_agents.types import ToolError
+from urllib.parse import urlparse
+
+
+logger = logging.getLogger(__name__)
+
+
+class DaprHTTPClient(BaseModel):
+    """
+    Client for sending HTTP requests to Dapr endpoints.
+    """
+
+    dapr_app_id: Optional[str] = Field(
+        default="", description="Optional name of the Dapr App ID to invoke."
+    )
+
+    dapr_http_endpoint: Optional[str] = Field(
+        default="",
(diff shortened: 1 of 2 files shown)


CasperGN:
@yaron2, @Cyb3rWard0g can you take a look at the `otel.py`?

Is this better placed under `dapr_agents.agent.utils`?

jgmartinez:
Looks good! But I would make the open telemetry instrumentation optional. It can be a bit of an overhead and there are other tools out there doing automatic instrumentation and this might not be needed :)

CasperGN:
> Looks good! But I would make the open telemetry instrumentation optional. It can be a bit of an overhead and there are other tools out there doing automatic instrumentation and this might not be needed :)

This is a really good point. I'll add it as ENV variable to set.

CasperGN:
@jgmartinez this works well:
```shell
export DAPR_AGENTS_OTEL_ENABLED=false                   
python3                              
Python 3.10.17 (main, Apr  8 2025, 12:10:59) [Clang 16.0.0 (clang-1600.0.26.6)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> from distutils.util import strtobool
>>> otel_enabled: bool = bool(strtobool(os.getenv("DAPR_AGENTS_OTEL_ENABLED", "True")))
>>> isinstance(otel_enabled, bool)
True
>>> otel_enabled
False
>>> 
```

dapr_agents/tool/http/client.py
```diff
@@ -0,0 +1,222 @@
...
+            )
+            set_logger_provider(otel_logger)
+
+            RequestsInstrumentor().instrument()
+
+        logger.debug("Initializing DaprHTTPClient client")
+
+        super().model_post_init(__context)
+
+    def post(
```
yaron2: Would it be possible to have a single invoke method that accepts the HTTP verb as a parameter instead of individual methods for each verb?

CasperGN: @yaron2 this is a good point. Have a look at fa680b3.

Should I also move the `otel.py` into a more proper location? I would assume under `dapr_agents.agent.utils`?

yaron2: I'd move it for the following import: `dapr_agents.agent.telemetry`

Cyb3rWard0g: agreed :)

CasperGN: `otel.py` moved into `dapr_agents.agent.telemetry`.

Tested with the environment variable to enable/disable OTel traces from the http client.

yaron2 (approved):

PR #112, opened 2025-05-02
Reorganize quickstart examples and added MCP implementation

bibryam:
- Renumber K8s multi-agent workflow example from 07 to 05 as it is an extention of 05
- Add two new MCP client examples:
  - 07: MCP agent with STDIO transport
  - 07: MCP agent with SSE transport
- Maintain consistent structure across README files

https://github.com/dapr/dapr-agents/issues/111

PR #122, opened 2025-05-23
Dapr university preparation

bibryam:
This PR adds several new examples to the Quickstarts. These are required for the Dapr University course on dapr-agents, where we need separate examples per topic for clarity.


bibryam:
This was supposed to have a very few changes, I think it contains changes from https://github.com/dapr/dapr-agents/pull/112 and once 112 is merged, many of the changes here won't be needed

PR #136, opened 2025-06-12
feat: add readme + imgs + todos + cleanup

sicoyle:
I went through the code and removed redundant comments, made stylistic changes, removed a file we don't use (from what I could tell), added missing conditionals in some places, added a readme on the orchestrators + workflow visuals that I created when reading through the workflows :)


dapr_agents/agent/actor/base.py
```diff
@@ -38,16 +38,16 @@ async def _on_activate(self) -> None:
         if not has_state:
             # Initialize state with default values if it doesn't exist
             logger.info(f"Initializing state for {self.actor_id}")
-            self.state = AgentActorState(overall_status=AgentStatus.IDLE)
+            self.agent_state = AgentActorState(overall_status=AgentStatus.IDLE)
```
filintod: wasn't overall_status changed to `status` later? some unit tests would be nice to capture this errors that can happen on refactoring

dapr_agents/workflow/orchestrators/random.py
```diff
@@ -82,6 +83,13 @@ def main_workflow(self, ctx: DaprWorkflowContext, input: TriggerAction):
                 f"Random workflow iteration {iteration + 1} started (Instance ID: {instance_id})."
             )
 
+        # Check Termination Condition
+        if iteration >= self.max_iterations:
```
filintod: isn't this taken care in another place?

sicoyle:
this is ready pls

sicoyle:
ready ready

dapr_agents/workflow/base.py
```diff
@@ -181,6 +182,21 @@ def wrapper(ctx: WorkflowActivityContext, *args, **kwargs):
 
         return wrapper
 
+    # TODO: can workflow discovery simply come from the runtime?
```
Cyb3rWard0g: Yeah, I was trying to support the concept of simply saying

```
@workflow

@task
@task
@task
```
For those that do just want to decorate their functions with the minimal understanding of this is a workflow and this is a task which I then execute in the workflow. :)

dapr_agents/agent/utils/auth.py
Cyb3rWard0g, line 1: Why did we remove this? Wondering since it is a quick util to handle OAuth tokens in auth headers. This can be part of tools. I guess we just expect user to define how their tools will handl auth? which would make sense https://github.com/dapr/dapr-agents/blob/main/dapr_agents/agent/utils/auth.py

sicoyle: I tried searching for it's use within the repo and couldn't find anything so I removed it. I can bring it back no problem if you'd like 🤷‍♀️

dapr_agents/workflow/agentic.py
```diff
@@ -109,6 +110,7 @@ def model_post_init(self, __context: Any) -> None:
         self._text_formatter = ColorTextFormatter()
 
         # Initialize state store client (used for persisting workflow state to Dapr)
+        # Why make a state store client and dapr_client here?
```
Cyb3rWard0g, line 113: mmm I guess here the idea was not to open and close connections for agent state and have a persistent one. However, I believe that with that we have to handle the "closing" of the connection the right way vs using `with daprclient` do X. right?

dapr_agents/workflow/orchestrators/llm/orchestrator.py
```diff
@@ -60,6 +60,8 @@ def model_post_init(self, __context: Any) -> None:
 
     @message_router
     @workflow(name="LLMWorkflow")
+    # TODO: set retry policies on the activities!
```
Cyb3rWard0g, line 63: agreed. we should.

dapr_agents/workflow/orchestrators/llm/orchestrator.py
```diff
@@ -60,6 +60,8 @@ def model_post_init(self, __context: Any) -> None:
...
+    # TODO: utilize prompt verdict value of failed as we do not currently use.
```
Cyb3rWard0g, line 64: I agree. I believe we need to handle exclusively `failed`. We handle anything that is not `continue` or if we reach the max number of iterations, but not necessarily if something fails. https://github.com/dapr/dapr-agents/blob/main/dapr_agents/workflow/orchestrators/llm/orchestrator.py#L245

mmm technically the prompt does not even mention failed, because i guess `failure` needs to come from the Dapr runtime itself right? https://github.com/dapr/dapr-agents/blob/main/dapr_agents/workflow/orchestrators/llm/schemas.py#L75 .

dapr_agents/agent/utils/auth.py
sicoyle: added back

dapr_agents/workflow/agentic.py
sicoyle: correct. I see trade offs both ways. We can leave for now, but yeah I wanted to ask as I combed through things. Thoughts @yaron2 ?

dapr_agents/workflow/orchestrators/llm/orchestrator.py
sicoyle: added link to this discussion in the file

dapr_agents/workflow/agentic.py
yaron2: Correct @Cyb3rWard0g. @sicoyle lets leave it for now

PR #137, opened 2025-06-16
test: init test setup and some tests + dev docs and proper setup with pyproject.toml

sicoyle:
This PR aims to make adding tests/features in future easier.

1. Modernized dependency management through removing `requirements.txt` and `dev-requirements.txt` files in favor of `pyproject.toml` file. This allows us to organize dependencies for main, test, dev, etc, giving us better dep isolation. This should also help to make things easier on new contributors with optional/simplified dependency groups.
2. Enhanced testing setup through running tests in CI now, tests added for the RandomOrchestrator with mocks. Note: we will need future refactoring to improve module structure for better testability.
3. Docs: created dev docs under `docs/development` and updated CI to acct for this.


pyproject.toml
```diff
@@ -13,6 +13,23 @@ license = {file = "LICENSE"}
 authors = [
     {name = "Dapr Authors", email = "dapr@dapr.io"}
 ]
+dependencies = [
```
filintod, line 16: it would be good to sort the list like in https://peps.python.org/pep-0631/

pyproject.toml
```diff
@@ -25,28 +42,23 @@ classifiers = [
...
-    "rich == 13.9.4",
-    "huggingface_hub == 0.30.2",
-    "numpy == 2.2.2",
-    "mcp == 1.7.1",
-    "opentelemetry-distro == 0.53b1",
-    "opentelemetry-exporter-otlp == 1.32.1",
-    "opentelemetry-instrumentation-requests == 0.53b1",
+[project.optional-dependencies]
+test = [
+    "pytest>=7.0.0",
```
filintod, line 47: same of these two,  also first `dev` then `test`

dapr_agents/__init__.py
```diff
@@ -1,6 +1,5 @@
 from dapr_agents.agent import (
     Agent,
-    AgentActor,
```
filintod, line 3: is removing AgentActor part of this PR?

sicoyle: no, i did this when i was trying to add tests while limiting scope of dependencies that i needed. Let me add back

sicoyle:
gimma a little longer on this one pls to make the build happy. I removed ActorAgent in some of the `__init__.py` files bc they give me a lot of issues earlier in adding the tests that I'm adding, and adding ActorAgent back fails the tests. So I need to find a diff workaround for this given the module setup and dependencies.

sicoyle:
I forgot about this build err... give me a little longer to circle back and fix this before we send it pls

Cyb3rWard0g (approved):
LGTM. Thank you @sicoyle !

PR #143, opened 2025-06-30
feat: provide type hints and better DX for new community members/users

sicoyle:
This PR adds a py.typed marker file and updates myproject.toml to include it in package-data. This signals to type checkers that the package provides inline type annotations, following PEP 561. It improves type-checking support for users of the package, and gives better support in IDEs.

For reference:
https://peps.python.org/pep-0561/#packaging-type-information
https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/

I've also added this in the python sdk for dapr so when we are using their packages, we can benefit from this on a consuming side too :) https://github.com/dapr/python-sdk/pull/812

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,6 +57,9 @@ Source = "https://github.com/dapr/python-sdk"
 include = ["dapr_agents"]
 exclude = ["tests"]
 
+[tool.setuptools.package-data]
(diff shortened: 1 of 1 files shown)


Cyb3rWard0g (approved):
Thank you @sicoyle ! 🎉

PR #147, opened 2025-07-02
add dapr client streaming tool calling

filintod:
This require changes in Python SDK that depend on changes on Dapr runtime that depends on Component-Contrib changes.

All these are coming


filintod:
this was depending on a previous PR version upstream. Will close this for now

PR #161, opened 2025-07-23
Refactor agent workflows, orchestrators, and integrations for reliability and modularity

Cyb3rWard0g:
This PR refactors agent workflows, orchestrators, and integrations to improve reliability, modularity, and standards compliance. Key updates include new mixin-based workflow architecture, enhanced error handling, unified message and state schemas, improved memory and vector store backends, and expanded support for LLM and tool APIs. The changes also standardize broadcast messaging, streamline agent orchestration, and update quickstart documentation to demonstrate new features and usage patterns.

### Hugging Face

- Updated Hugging Face integration to use the latest InferenceClient API and improved error handling for chat completions.
- Upgraded huggingface_hub Python package to version 0.33.4 and updated the inference client config schema.
- Changed the quickstart model to HuggingFaceTB/SmolLM3-3B due to deprecation of previous chat completion models.
- Added robust error handling for Hugging Face API responses.

### Agent Class 

- Refactored the Agent class to improve tool execution, error handling, and conversation flow.
- Added a persistent audit log for all tool executions and results, and refactored tool execution to support parallel, bounded concurrency.
- Improved shutdown handling, clarified finish reasons, and updated the conversation loop for better reliability and maintainability.
- Enhanced error reporting and updated the config schema for compatibility with new LLM and tool APIs.
- Introduced the FinishReason enum for clearer handling of LLM response types and agent flow control.
- Improved type annotations and docstrings for better maintainability and developer experience.
- Replaced the legacy process_response method with the more robust execute_tools method for tool execution.

### Agent Base

- Refactored the AgentBase class to centralize prompt template logic, improve attribute handling, and enhance chat history management for agents.
- Centralized prompt template selection and initialization, ensuring agent and LLM client use the same template.
- Improved attribute pre-filling and validation, with warnings for unused agent attributes in prompt templates.
- Enhanced chat history retrieval and formatting, supporting both property and method access.
- Updated message handling to always return copies with trimmed content for safety and consistency.
- Added helper methods and improved type annotations and docstrings for better maintainability and developer experience.

### Chat Completion Helper

- Updated the ChatPromptHelper to support passing tool_calls to AssistantMessage when present in the message data.
- Added logic to detect if the role is "assistant" and tool_calls exist, then include tool_calls in the constructed message.
- Enables richer assistant messages that can reference tool calls directly, improving agent-tool interaction and response handling.

### MessageContent

- Enhanced the MessageContent class to better support tool call metadata and message handling.
- Added a new optional tool_call_id field to track the specific tool call associated with a message.
- Updated post-initialization validation to dynamically remove tool_call_id if it is not set, keeping object representations clean.
- Changed the return type of ChatCompletion.get_message() from Optional[str] to Optional[dict] for richer message content.

### Agentic Workflow Mixin

- Refactored workflow architecture to use mixins for messaging, pub/sub, state management, and service logic, improving modularity and code organization.
- Consolidated related functionality into dedicated mixin classes for easier reuse and maintenance.
- Updated AgenticWorkflow to compose these mixins, streamlining workflow feature integration.

### Durable Agent

- Refactored the DurableAgent to improve workflow state management, tool execution, and message handling for reliability and clarity.
- Introduced new state and message schemas for better workflow tracking and persistence.
- Improved tool execution logic, including parallel tool calls and robust error handling.
- Enhanced workflow entry, assistant, and tool message recording for auditability.
- Updated broadcast and finalization logic to ensure consistent state updates and message persistence.
- Replaced legacy workflow/message classes with unified DurableAgentMessage, DurableAgentToolHistoryEntry, and related schemas for consistent state and message tracking.
- Improved workflow state structure to support richer message history, tool execution auditability, and reliable persistence across agent workflows.

### Workflow Base

- Refactored the workflow base to modularize Dapr integration, improve error handling, and delegate service/state logic to mixins.
- Refactored `WorkflowApp` to remove direct Dapr client dependencies and service invocation methods, shifting integration logic to mixins for modularity.
- Added Dapr availability checks and clearer error messaging to improve developer experience when Dapr is not running.
- Consolidated utility imports and streamlined workflow/task registration logic for maintainability.
- Removed legacy service invocation logic.

### Agentic Workflow

- Refactored AgenticWorkflow to compose messaging, pub/sub, service, and state management features using mixins for modularity.
- Simplified initialization and metadata management, consolidating agent registration and state store logic.
- Improved chat history retrieval and validation, supporting both vector and list memory backends.
- Enhanced FastAPI service integration and workflow execution from HTTP requests.
- Removed legacy direct pub/sub and routing logic, delegating these responsibilities to mixins and utilities.

### Workflow Decorators

- Split Workflow decorators for better organization.

### ConversationDaprStateMemory

- Refactored the memory backend to return full message objects and improved logging for message addition and retrieval.
- Updated logging to include full message details when adding to session state.
- Changed get_messages to return all message fields, supporting richer memory retrieval.

### VectorStoreBase Document Addition Refactor

- Refactored the vector store base to return document IDs and ensure consistent output when adding documents.
- Updated add_documents to return a list of document IDs for added documents.
- Improved result handling for more reliable document tracking.

### Orchestrator Workflow Refactor and BroadcastMessage Standardization

- Refactored LLM, Random, and RoundRobin orchestrators to standardize broadcast message handling, improve decorator imports, and enhance workflow robustness.
- Standardized BroadcastMessage usage for agent broadcasts in Random and RoundRobin orchestrators.
- Unified decorator imports by consolidating task, workflow, and message_router for all orchestrators.
- Improved iteration handling with default values to prevent errors in workflow loops.
- Enhanced workflow event handling by using getattr for safer attribute access in agent responses.
- Added step renumbering and verdict handling for clearer workflow progression in LLM orchestrator.

### DIFF
--- a/quickstarts/03-durable-agent-tool-call/durable_weather_agent.py
+++ b/quickstarts/03-durable-agent-tool-call/durable_weather_agent.py
@@ -9,7 +9,7 @@ async def main():
     load_dotenv()
     logging.basicConfig(level=logging.INFO)
 
-    # 1️⃣ Instantiate your agent (no .as_service())
+    # 1️⃣ Instantiate your agent
     weather_agent = DurableAgent(
         role="Weather Assistant",
         name="Stevie",


yaron2:
@Cyb3rWard0g See failing lints

quickstarts/07-agent-mcp-client-sse/app.py
```diff
@@ -31,6 +31,7 @@ async def main():
             state_key="workflow_state",
             agents_registry_store_name="agentstatestore",
             agents_registry_key="agents_registry",
+            broadcast_topic_name=None,
```
yaron2: Shouldn't this property default to `None`?

Cyb3rWard0g: good question. It is actually set to a topic name `beacon_channel` https://github.com/dapr/dapr-agents/blob/1c832636eb4173dced965debc679ee42702b7cb3/dapr_agents/workflow/agentic.py#L56 . For single Durable Agent execution, if it finds other agents registered, it will try to broadcast the message. I added logic to say, only broadcast if 1) the topic is set -> Topic subscription does not occur and 2) Then, if there are agents registered. This does not affect any existing multi-agent setups. However, for single durable agents without any other agents to collaborate with, user has to set it to `None`.

yaron2: Hmm.. isn't it more predictable behavior that by default agents do not broadcast, but instead that multi-agent setups are configured explicitly?

dapr_agents/__init__.py
```diff
@@ -1,19 +1,40 @@
...
+    "DockerCodeExecutor",
+    "LocalCodeExecutor",
+    "ElevenLabsSpeechClient",
+    "HFHubChatClient",
+    "NVIDIAChatClient",
+    "NVIDIAEmbeddingClient",
+    "OpenAIAudioClient",
+    "OpenAIChatClient",
+    "OpenAIEmbeddingClient",
+    "AgentTool",
```
sicoyle, line 33: I'm assuming Agent and DurableAgent can use AgentTool? We should denote that somewhere 🤔

dapr_agents/__init__.py
```diff
@@ -1,19 +1,40 @@
...
+    "tool",
```
sicoyle, line 34: why is tool lowercased and exported? What is the difference between this and AgentTool above?

dapr_agents/__init__.py
```diff
@@ -1,19 +1,40 @@
...
+    "AgenticWorkflow",
+    "LLMOrchestrator",
+    "RandomOrchestrator",
+    "RoundRobinOrchestrator",
+    "WorkflowApp",
```
sicoyle, line 39: Do we need to expose WorkflowApp?

sicoyle (commented):
Few comments so far. I'm about half way through reviewing :)

dapr_agents/agents/agent/agent.py
```diff
@@ -1,235 +1,279 @@
...
 
 
+class FinishReason(str, Enum):
+    STOP = "stop"
+    LENGTH = "length"
+    CONTENT_FILTER = "content_filter"
+    TOOL_CALLS = "tool_calls"
+    FUNCTION_CALL = "function_call"  # deprecated
+
+
```
sicoyle, line 27: this will be difficult to maintain and will change as we update dependencies. I know this because when I added tool calling support in dapr/dapr and dapr/components-contrib, I had to account for anthropic giving back a different stop reason than some of the other providers. https://github.com/sicoyle/components-contrib/blob/master/tests/conformance/conversation/conversation.go#L106
Is there a way we can make this still bubble back the llm custom finish reason without maintaining our own list on top of that?

dapr_agents/agents/agent/agent.py
```diff
@@ -1,235 +1,279 @@
...
-        Sets the tool choice strategy based on provided tools.
+        Post-initialization hook for Agent.
+        Sets the tool choice strategy based on provided tools and calls the base model setup.
+        Args:
+            __context (Any): Context passed from Pydantic's model initialization.
         """
-        self.tool_choice = self.tool_choice or ("auto" if self.tools else None)
+        # Set tool_choice to 'auto' if tools are provided and not explicitly set
+        if self.tool_choice is None:
+            self.tool_choice = "auto" if self.tools else None
```
sicoyle: can we explicitly set the string none too? Also, there can be a "required" option from what I saw. In Dapr runtime, I set a default no matter what, but from client POV I suppose we should match defaults if we're to set them
https://github.com/sicoyle/dapr/blob/master/pkg/api/universal/conversation.go#L412

sicoyle: I think you're doing good here on not erroring out if it's an unknown string set bc these llm providers will be changing constantly, so you're giving us nice flexibility here 👏

dapr_agents/agents/agent/agent.py
```diff
@@ -1,235 +1,279 @@
...
+        # Process conversation iterations and return the result
         return await self.process_iterations(messages)
 
-    async def process_response(self, tool_calls: List[ToolCall]) -> None:
+    async def execute_tools(self, tool_calls: List[ToolCall]) -> List[ToolMessage]:
         """
-        Asynchronously executes tool calls and appends tool results to memory.
+        Executes a batch of tool calls in parallel, bounded by max_concurrent, using asyncio.gather (Python 3.10+ compatible).
+        Each tool call is executed asynchronously using run_tool, and results are appended to the persistent audit log (tool_history).
+        If any tool call fails, the error is propagated and other tasks continue unless you set return_exceptions=True.
```
sicoyle: do all llm providers support tool calls in parallel? How are we handling the case if it is not supported from the llm? For example, I think I need to explore if I do need to add parallel tool call support in Dapr/dapr (runtime) so yeah curious on this one given my recent tool calling context :)

sicoyle: we can drop the Python 3.10+ compatibility comment too bc thats stated in the root of this project

dapr_agents/agents/agent/agent.py
```diff
@@ -1,235 +1,279 @@
...
-            tool_id = tool_call.id
-            function_args = (
-                tool_call.function.arguments_dict
-            )  # Use the property to get dict
-
-            if not function_name:
-                logger.error(f"Tool call missing function name: {tool_call}")
-                continue
+        # Limiting concurrency to avoid overwhelming downstream systems
+        max_concurrent = 10
```
sicoyle, line 133: can we make this configurable by clients?

sicoyle: looks like it is configurable, can you double check it's not set as an input param before picking 10? If we pick 10 then pls update the comment that says we default to 5 😅  I prefer 5 as the default over 10 personally.

dapr_agents/agents/agent/agent.py
```diff
@@ -1,235 +1,279 @@
...
+        semaphore = asyncio.Semaphore(max_concurrent)
+
+        async def run_and_record(tool_call: ToolCall) -> ToolMessage:
+            """
+            Executes a single tool call, respecting the concurrency limit.
+            Appends the result to the persistent audit log.
+            If the function name is missing, returns a ToolMessage with error status and raises AgentError.
```
sicoyle, line 140: ooo i like the err handling improvements

dapr_agents/agents/agent/agent.py
```diff
@@ -1,235 +1,279 @@
...
+            """
+            async with semaphore:
+                function_name = tool_call.function.name
+                tool_id = tool_call.id
+                function_args = tool_call.function.arguments_dict
+
+                if not function_name:
+                    error_msg = f"Tool call missing function name: {tool_call}"
```
sicoyle, line 148: Eventually we'll need to update this to support other types of tools - just mental note to self as reviewing

dapr_agents/agents/agent/agent.py
```diff
@@ -1,235 +1,279 @@
...
+                try:
+                    logger.debug(
+                        f"Executing {function_name} with arguments {function_args}"
+                    )
+                    result = await self.run_tool(function_name, **function_args)
+                    tool_message = ToolMessage(
+                        tool_call_id=tool_id, name=function_name, content=str(result)
+                    )
+                    # Printing the tool message for visibility
+                    self.text_formatter.print_message(tool_message)
```
sicoyle, line 172: can we make this optional based on like a verbose flag or something? That way the tool message, if huge, wont take up the whole terminal output

dapr_agents/llm/huggingface/client.py
```diff
@@ -17,33 +17,41 @@ class HFHubInferenceClientBase(LLMClientBase):
...
     )
     token: Optional[Union[str, bool]] = Field(
         default=None,
-        description="Hugging Face token. Defaults to the locally saved token if not provided. Pass `False` to disable authentication.",
+        description="Hugging Face access token for authentication. If None, uses the locally saved token. Set to False to skip sending a token. Mutually exclusive with api_key.",
     )
     api_key: Optional[Union[str, bool]] = Field(
-        default=None,
-        description="Alias for `token` for compatibility with OpenAI's client. Cannot be used if `token` is set.",
+        default=None, description="Alias for token. Use only one of token or api_key."
```
sicoyle, line 31: do we check for if these fields are set as mutually exclusive somewhere?

dapr_agents/llm/huggingface/client.py
```diff
@@ -17,33 +17,41 @@ class HFHubInferenceClientBase(LLMClientBase):
...
     )
     base_url: Optional[str] = Field(
         default=None,
-        description="Base URL to run inference. Alias for `model`. Cannot be used if `model` is set.",
+        description="Custom endpoint URL for inference. Used for private deployments or TGI endpoints. Cannot be set if 'model' is a Hub ID.",
     )
-    headers: Optional[Dict[str, str]] = Field(
+    timeout: Optional[float] = Field(
```
sicoyle, line 37: why float instead of int?

dapr_agents/memory/base.py
```diff
@@ -73,3 +75,16 @@ def reset_memory(self):
             This method must be implemented by subclasses.
         """
         pass
+
+    @staticmethod
+    def _convert_to_dict(message: Union[Dict, BaseMessage]) -> Dict:
```
sicoyle, line 80: is there a way we can move away from having the typed BaseMessage and the dict version? Or are both required?

dapr_agents/memory/vectorstore.py
```diff
@@ -85,17 +97,17 @@ def get_messages(
...
         """
         Retrieves messages from the vector store. If a query is provided, it performs a similarity search.
 
         Args:
-            query_embeddings (Optional[List[List[float]]]): The query embeddings for similarity search.
-            k (int): The number of similar results to retrieve.
-            distance_metric (str): The distance metric to use ("l2", "ip", "cosine").
+            query_embeddings (Optional[List[List[float]]], optional): The query embeddings for similarity search. Defaults to None.
+            k (int, optional): The number of similar results to retrieve. Defaults to 4.
+            distance_metric (str, optional): The distance metric to use ("l2", "ip", "cosine"). Defaults to "cosine".
```
sicoyle, line 107: do these defaults get set?

dapr_agents/memory/vectorstore.py
```diff
@@ -105,45 +117,49 @@ def get_messages(
...
                 metadata
                 and "user_message" in metadata
                 and "assistant_message" in metadata
             ):
-                messages.append(UserMessage(metadata["user_message"]))
-                messages.append(AssistantMessage(metadata["assistant_message"]))
+                messages.append({"role": "user", "content": metadata["user_message"]})
+                messages.append(
+                    {"role": "assistant", "content": metadata["assistant_message"]}
+                )
```
sicoyle, line 131: will we eventually have other message types here? I ask bc when I added in tool calling support and designed the conversation api support for this in dapr/dapr, I used openAI as a heavy source of truth/reference which led me to these types of messages:
https://github.com/dapr/dapr/blob/master/dapr/proto/runtime/v1/dapr.proto#L1422
https://github.com/openai/openai-go/blob/main/chatcompletion.go#L1449

dapr_agents/memory/vectorstore.py
```diff
@@ -105,45 +117,49 @@ def get_messages(
...
         """
         Performs a similarity search in the vector store and retrieves the conversation pairs.
 
         Args:
-            query_embeddings (Optional[List[List[float]]]): The query embeddings.
-            k (int): The number of results to return.
-            distance_metric (str): The distance metric to use ("l2", "ip", "cosine").
+            query_embeddings (Optional[List[List[float]]], optional): The query embeddings. Defaults to None.
+            k (int, optional): The number of results to return. Defaults to 4.
+            distance_metric (str, optional): The distance metric to use ("l2", "ip", "cosine"). Defaults to "cosine".
```
sicoyle, line 152: do these defaults get set somewhere?

dapr_agents/memory/vectorstore.py
```diff
@@ -156,21 +172,36 @@ def get_similar_conversation(
...
+                        messages.append(
+                            {
+                                "role": "user",
+                                "content": metadata["user_message"],
+                                "timestamp": timestamp,
+                            }
+                        )
+                        messages.append(
+                            {
+                                "role": "assistant",
```
sicoyle, line 184: how can we support the case of other llm providers having different delineation on role strings?

https://community.openai.com/t/open-ai-chat-roles-has-three-roles-but-langchain-has-more/252294

dapr_agents/storage/vectorstores/__init__.py
```diff
@@ -1,3 +1,5 @@
 from .base import VectorStoreBase
 from .chroma import ChromaVectorStore
 from .postgres import PostgresVectorStore
+
+__all__ = ["VectorStoreBase", "ChromaVectorStore", "PostgresVectorStore"]
```
sicoyle, line 5: doesn't this mean when someone does a piip install on our project that they then bring in all of these dependencies? Do we want that? What if they dont use vector stores at all?

quickstarts/07-agent-mcp-client-sse/app.py
Cyb3rWard0g: Yeah that works. I just did not want to break any existing multi-agent scenarios. I can set it to None by default and let people define the broadcast channel for all multi-agents scenarios.

dapr_agents/types/message.py
```diff
@@ -120,23 +120,25 @@ class ToolCall(BaseModel):
...
     tool_calls: Optional[List[ToolCall]] = None
     function_call: Optional[FunctionCall] = None
+    tool_call_id: Optional[str] = None
 
     @model_validator(mode="after")
     def remove_empty_calls(self):
         attrList = []
         for attribute in self.__dict__:
-            if attribute in ("tool_calls", "function_call"):
+            if attribute in ("tool_calls", "function_call", "tool_call_id"):
```
sicoyle, line 141: i dont think all llm providers set a tool call id... do we acct for that?

dapr_agents/workflow/__init__.py
```diff
@@ -1,5 +1,17 @@
...
+__all__ = [
+    "WorkflowApp",
+    "WorkflowTask",
+    "AgenticWorkflow",
+    "LLMOrchestrator",
+    "RandomOrchestrator",
+    "RoundRobinOrchestrator",
+    "workflow",
+    "task",
+    "message_router",
```
sicoyle, line 16: should the lowercase ones be here?

dapr_agents/__init__.py
Cyb3rWard0g: They both use it right? Since they both inherit from AgentBase https://github.com/dapr/dapr-agents/blob/cyb3rward0g/updating-agentic-workflows/dapr_agents/agents/base.py#L87

dapr_agents/__init__.py
Cyb3rWard0g: tool is the decorator and AgentTool is the class used to handle tools decorated and other tool methods.

dapr_agents/__init__.py
Cyb3rWard0g: I think we should since it is the basic class to build more deterministic workflows with LLM-based activities (task) features. right?

dapr_agents/agents/agent/agent.py
Cyb3rWard0g: mmm good question, how should we handle that? we need a mapping somehow right?

dapr_agents/agents/agent/agent.py
Cyb3rWard0g: All models with function calling should support the concept that if you get more than one tool call, you can execute them in parallel.

dapr_agents/agents/agent/agent.py
Cyb3rWard0g: you mean the LLM client?

Cyb3rWard0g: The max_concurrent is in relation tot he loop taken to process the user input with an LLM, execute tools and produce a response right? so it is done at the agent level. Yeah we can set it to 5 :)

dapr_agents/workflow/mixins/messaging.py
```diff
@@ -0,0 +1,87 @@
...
+        try:
+            agents_metadata = self.get_agents_metadata()
+            if name not in agents_metadata:
+                logger.warning(
+                    f"Target '{name}' is not registered as an agent. Skipping message send."
+                )
+                return
+
+            agent_metadata = agents_metadata[name]
+            logger.info(f"{self.name} sending message to agent '{name}'.")
```
sicoyle, line 75: debug level instead?

dapr_agents/agents/agent/agent.py
Cyb3rWard0g: Yes! I meant to do that one, but I figured I can do that in a separate PR since there is a lot going on already :)

dapr_agents/workflow/mixins/state.py
```diff
@@ -0,0 +1,249 @@
...
+                if not isinstance(state_data, dict):
+                    raise TypeError(
+                        f"Invalid state type retrieved: {type(state_data)}. Expected dict."
+                    )
+                return (
+                    self.validate_state(state_data) if self.state_format else state_data
+                )
+
+            logger.info(
+                f"No existing state found for key '{self.state_key}'. Initializing empty state."
```
sicoyle, line 119: should these be debug level?

quickstarts/03-durable-agent-tool-call/durable_weather_agent.py
```diff
@@ -0,0 +1,34 @@
...
+        message_bus_name="messagepubsub",
+        state_store_name="workflowstatestore",
+        state_key="workflow_state",
+        agents_registry_store_name="agentstatestore",
+        agents_registry_key="agents_registry",
+        broadcast_topic_name=None,
+        tools=tools,
+    )
+    # 2️⃣ Start the agent service
+    await weather_agent.run("What's the weather in Boston?")
```
sicoyle: so durable agent can now run with the run()? Can it still do the .as_service() for a long lived agent? I can't remember if we have it as a starting mechanism or not, but would be nice to also have an as_consumer() where a pubsub event could trigger its start

sicoyle: if it can support both, can you update the comment pls to say it could be optional to instead use the .as_service? or does the run() enable both means of start?

dapr_agents/agents/base.py
```diff
@@ -109,6 +104,22 @@ class AgentBase(BaseModel, ABC):
...
+
+## Name
+Your name is {name}.
+
+## Role
+Your role is {role}.
+
+## Goal
+{goal}.
+{instructions}
```
```suggestion
## Goal
{goal}.

## Instructions
{instructions}.
```

?

dapr_agents/agents/durableagent/agent.py
```diff
@@ -1,35 +1,46 @@
...
 logger = logging.getLogger(__name__)
 
 
+class FinishReason(str, Enum):
+    UNKNOWN = "unknown"
+    STOP = "stop"
+    LENGTH = "length"
+    CONTENT_FILTER = "content_filter"
+    TOOL_CALLS = "tool_calls"
+    FUNCTION_CALL = "function_call"  # deprecated
```
sicoyle, line 41: isn't this elsewhere too? Can we have in a centralized location?

dapr_agents/agents/durableagent/agent.py
```diff
@@ -70,15 +83,17 @@ def set_agent_and_topic_name(cls, values: dict):
...
         # Call AgenticWorkflow's model_post_init first to initialize state store and other dependencies
         super().model_post_init(__context)
 
         # Name of main Workflow
         # TODO: can this be configurable or dynamic? Would that make sense?
         self._workflow_name = "ToolCallingWorkflow"
-        self.tool_choice = self.tool_choice or ("auto" if self.tools else None)
+        # Set tool_choice to 'auto' if tools are provided, otherwise None
+        if self.tool_choice is None:
+            self.tool_choice = "auto" if self.tools else None
```
sicoyle: i guess the regular agent is where i put my original comment on default setting here... can these be moved to the base agent class instead then? so we dont have the same logic like this for both agents separately pls?

dapr_agents/agents/durableagent/agent.py
```diff
@@ -92,24 +107,30 @@ def model_post_init(self, __context: Any) -> None:
...
+        # Prepare input payload for workflow
+        if isinstance(input_data, dict):
+            input_payload = input_data
+        else:
+            input_payload = {"task": input_data}
+
+        # Kick off the workflow and block until it finishes:
+        return await self.run_and_monitor_workflow_async(
+            workflow=self._workflow_name,
+            input=input_payload,
```
sicoyle, line 122: i see now, thank you for implementing this! So now both agent types can be started with .as_service() and run()?

quickstarts/03-durable-agent-tool-call/durable_weather_agent.py
Cyb3rWard0g: oh yes we can run it with as service, etc. Everything that already existed still works. I just wanted to work on your comment from a previous PR. You mentioned that it would be good to be able to also have a RUN command :)

dapr_agents/agents/durableagent/agent.py
```diff
@@ -118,148 +139,129 @@ def tool_calling_workflow(self, ctx: DaprWorkflowContext, message: TriggerAction
...
-            self.tool_history.clear()
-
-        # Step 6: Determine if Workflow Should Continue
+            if tool_calls:
+                if not ctx.is_replaying:
+                    logger.debug(f"Executing {len(tool_calls)} tool call(s)..")
+                # Run the tool calls in parallel
+                parallel = [
+                    ctx.call_activity(self.run_tool, input={"tool_call": tc})
+                    for tc in tool_calls
```
sicoyle, line 223: would there be a case a user wouldn't want tool calls run in parallel? Should this be optional?

dapr_agents/agents/telemetry/__init__.py
```diff
@@ -1 +1,3 @@
 from .otel import DaprAgentsOTel
+
+__all__ = ["DaprAgentsOTel"]
```
sicoyle, line 3: can we make this be DaprAgentsOtel?

dapr_agents/__init__.py
sicoyle: i guess i was commenting moreso on should it be just Tool or DaprAgentTool instead of AgentTool which might confuse folks to think only the Agent class supports this since there is not a DurableAgentTool too, or if its fine and we can just document

quickstarts/07-agent-mcp-client-sse/Stevie_state.json
```diff
@@ -1,29 +1,112 @@
...
         }
-    }
+    },
+    "chat_history": [
+        {
+            "content": "What is the weather in New York?",
+            "role": "user",
+            "id": "b57d8dcf-a84e-4ed9-b6eb-b27360a6a466",
+            "timestamp": "2025-07-23T02:35:58.890010"
+        },
```
sicoyle, line 77: each request to my understanding a user can specify different tools, so eventually in another PR or in future we could add an available_tools field maybe here

quickstarts/07-agent-mcp-client-sse/Stevie_state.json
```diff
@@ -1,29 +1,112 @@
...
+                        "name": "LocalGetWeather",
+                        "arguments": "{\"location\":\"New York\"}"
+                    }
+                }
+            ],
+            "id": "7b508dae-d86a-4cb5-9b84-4bf13f5ef3a3",
+            "timestamp": "2025-07-23T02:35:59.660877"
+        },
+        {
+            "content": "New York: 79F.",
```
sicoyle, line 96: how does this look if the tool call execution result is an object?

sicoyle (commented):
I've gotten through all the files changed now 😅 Overall really liking the changes 👏

dapr_agents/agents/agent/agent.py
sicoyle: do we need to create a list of finish reasons or can we pass this through and say if the finishReason is set then we're done? Where all are we using this field?

dapr_agents/agents/agent/agent.py
sicoyle: ah ok, yeah that makes sense. thanks!

dapr_agents/agents/agent/agent.py
sicoyle: absolutely

dapr_agents/memory/base.py
Cyb3rWard0g: That is a good question. I think the BaseMessage is more so what people can do thins like

```python
UserMessage("Hello")
AssistantMessage("Hello, how can I help you?")
```

Those are based on `BaseMessage`.

dapr_agents/memory/vectorstore.py
Cyb3rWard0g: Ahh yes! I have not considered the developer or other message types there. We should consider all of those you have. Yes. I will add that in this PR.

dapr_agents/types/message.py
Cyb3rWard0g: mmm those following OpenAI schemas for function calling should, but you are right. mmm so the idea with that was to use a base class that we could use to handle multiple types of messages. Technically if that attribute is not set it is set to None by default and removed right?

dapr_agents/workflow/mixins/messaging.py
Cyb3rWard0g: mmm yeah that could be Debug. I use Debug for when we want to show the entire message. I believe I have debug for those messaging classes underneath handling the pub/sub.

dapr_agents/workflow/mixins/state.py
Cyb3rWard0g: Yeah those could be debug. It is hard to figure out which is info and debug tbh. 😅

dapr_agents/agents/durableagent/agent.py
Cyb3rWard0g: We can have it at the AgentBase. I can move it there.

dapr_agents/agents/durableagent/agent.py
Cyb3rWard0g: `DurableAgent` can be `as_service()` and `run()`. However, `Agent` is only `run()`.

dapr_agents/agents/durableagent/agent.py
Cyb3rWard0g: That is a good question, I believe that if a user gets more than 1 tool call, that user would want to run those at the same time. However, maybe the user might want to handle them 1 by 1 for those tools that maybe perform an LLM task which might go against their Token Per Minute (TPM) quota. That is an extreme case of course :).

dapr_agents/agents/telemetry/__init__.py
Cyb3rWard0g: yes, if that makes sense.

quickstarts/07-agent-mcp-client-sse/Stevie_state.json
Cyb3rWard0g: that is a good idea!

quickstarts/07-agent-mcp-client-sse/Stevie_state.json
Cyb3rWard0g: that is a great question! That is something that needs to be handled, IMO, at the tool level. Technically, tools output will alawys be processed by the LLM as a string. However, if the user wants the result of a tool to be processed as an object to do additional things , that should not be handled by the LLM. The LLM should always get a string ;) . If the user decided to return an object, it will always be treated as a string. Like in this example, we return Pydantic Objects. They are processed as a string: https://github.com/dapr/dapr-agents/blob/1c832636eb4173dced965debc679ee42702b7cb3/quickstarts/01-hello-world/03_durable_agent.py

dapr_agents/agents/agent/agent.py
Cyb3rWard0g: thank you for the Python 3.10+ compatibility context :)

dapr_agents/workflow/__init__.py
Cyb3rWard0g: Yes. they are just decorators to define workflows, tasks and workflows that process a message.

dapr_agents/__init__.py
Cyb3rWard0g: ohhh got it! mmm when people define a tool for agents either for Agent or DurableAgent, they use the `@tool` decorator. Users do not need to I guess define or set `AgentTool` when defining tools. we should be fine. right? Or in what context a user would use AgentTool directly?

quickstarts/07-agent-mcp-client-sse/app.py
Cyb3rWard0g: done ✅

dapr_agents/__init__.py
Cyb3rWard0g: I added `This class can be used by the Agent and DurableAgent types to define tools that can be executed` in the `AgentTool` class docstrings 😉

dapr_agents/agents/agent/agent.py
Cyb3rWard0g: I now remember why I set it to None and not `none`. I guess I would have to add some logic since APIs like OpenAI, I get

```
AgentError: Failed during chat generation: Error code: 400 - {'error': {'message': "Invalid value for 'tool_choice': 'tool_choice' is only allowed when 'tools' are specified.", 'type': 'invalid_request_error', 'param': 'tool_choice', 'code': None}}
```

Setting it to None was being removed from the request parameters.

Cyb3rWard0g: centralized `tool_choice` logic in `AgentBase` ✅

dapr_agents/agents/agent/agent.py
Cyb3rWard0g: removed `Python 3.10+ compatibility` context from docstring. ✅

yaron2:
@Cyb3rWard0g I think it'd be great to set `state_key` to `workflow_state` by default and make it an optional parameter

Cyb3rWard0g:
> @Cyb3rWard0g I think it'd be great to set `state_key` to `workflow_state` by default and make it an optional parameter

Would you mind sharing an example or additional details?

yaron2:
> > @Cyb3rWard0g I think it'd be great to set `state_key` to `workflow_state` by default and make it an optional parameter
> 
> Would you mind sharing an example or additional details?

Yes, instead of:

```python
weather_agent = DurableAgent(
            role="Weather Assistant",
            name="Stevie",
            goal="Help humans get weather and location info using smart tools.",
            instructions=[
                "Respond clearly and helpfully to weather-related questions.",
                "Use tools when appropriate to fetch or simulate weather data.",
                "You may sometimes jump after answering the weather question.",
            ],
            tools=tools,
            message_bus_name="messagepubsub",
            state_store_name="workflowstatestore",
            state_key="workflow_state",
            agents_registry_store_name="agentstatestore",
            agents_registry_key="agents_registry",
        )
```

Allow this:

```python
weather_agent = DurableAgent(
            role="Weather Assistant",
            name="Stevie",
            goal="Help humans get weather and location info using smart tools.",
            instructions=[
                "Respond clearly and helpfully to weather-related questions.",
                "Use tools when appropriate to fetch or simulate weather data.",
                "You may sometimes jump after answering the weather question.",
            ],
            tools=tools,
            message_bus_name="messagepubsub",
            state_store_name="workflowstatestore",
            agents_registry_store_name="agentstatestore",
            agents_registry_key="agents_registry",
        )
```

This means the value defaults to `workflow_state` if not provided by the user. This would simplify agent construction.

dapr_agents/agents/agent/agent.py
Cyb3rWard0g: I ended up removing the concept of `finishreason`. In this new PR, to not overwhelm this one 😅, I updated how we process tool calls and direct responses. I standardized the chat response and for each provider `NVIDIA`, `HuggingFace Hub` and `OpenAI`, I can process different ways to capture a direct chat response with plain text and with tools. As you mentioned, some providers have their own `finish reason` to process direct response and response with tools. https://github.com/dapr/dapr-agents/pull/163

dapr_agents/agents/agent/agent.py
Cyb3rWard0g: This should be done at the Agent `Think -> Tool Execution -> Response` loop. If we do it at the `LLM Client` level, it will not be as clean. For example, let's say we have a `Weather Agent` with multiple tools. The user asks for information about the weather in 1 City. This will be

* 1 LLM Call  - ✅ 📡 LLM Chat Completion
* 1 LLM Response (with tools)
* 1 Tool Execution
* 1 Tool Message (Results of execution)
* 1 LLM Response (Summary of the iteration) - ✅ 📡 LLM Chat Completion

So that should be `1` Iteration. Now, if the user asks for the weather in 5 cities, we still have 1 iteration but multiple tools executed. We still do 2 interactions with the LLM. 

* 1 LLM Call  - ✅ 📡 LLM Chat Completion
* 1 LLM Response (with tools)
* 5 Tool Execution
* 5 Tool Message (Results of execution)
* 1 LLM Response (Summary of the iteration) - ✅ 📡 LLM Chat Completion

So it is better to do it per "LLM (Think) -> Tool Use -> LLM (Response)".

dapr_agents/llm/huggingface/client.py
Cyb3rWard0g: Yes we do that at the `model validator` section when initializing the client: https://github.com/dapr/dapr-agents/blob/main/dapr_agents/llm/huggingface/client.py#L50

dapr_agents/llm/huggingface/client.py
Cyb3rWard0g: just following `HuggingFace Hub` Python SDK `InferenceClient`

```python
class InferenceClient:
    """
    Initialize a new Inference Client.

    [`InferenceClient`] aims to provide a unified experience to perform inference. The client can be used
    seamlessly with either the (free) Inference API, self-hosted Inference Endpoints, or third-party Inference Providers.

    Args:
........
        timeout (`float`, `optional`):
            The maximum number of seconds to wait for a response from the server. Defaults to None, meaning it will loop until the server is available.
...........
    """
```

dapr_agents/memory/base.py
Cyb3rWard0g: It is just for flexibility. If a user interacts with those methods to add messages, writing JSON objects like `{"role": "user", "content":"hello"}` might be a little bit annoying. Having the `BaseMessage` style described in my previous comment, makes it easy. Therefore, when dealing with memory, I would expect also a user trying to add `BaseMessage` instances directly.

dapr_agents/memory/vectorstore.py
Cyb3rWard0g: Yes they get set 

```python
def get_messages(
        self,
        query_embeddings: Optional[List[List[float]]] = None,
        k: int = 4,
        distance_metric: str = "cosine",
    ) -> List[Dict[str, Any]]:
        """
        Retrieves messages from the vector store. If a query is provided, it performs a similarity search.

        Args:
            query_embeddings (Optional[List[List[float]]], optional): The query embeddings for similarity search. Defaults to None.
            k (int, optional): The number of similar results to retrieve. Defaults to 4.
            distance_metric (str, optional): The distance metric to use ("l2", "ip", "cosine"). Defaults to "cosine".

        Returns:
            List[Dict[str, Any]]: A list of all stored or similar messages as dictionaries with 'role' and 'content'.
        """
  ```

File: dapr_agents/memory/vectorstore.py (hunk 16 above)
Comment (Cyb3rWard0g) [reply dapr_agents/memory/vectorstore.py:131]: I will add that in another PR to improve our DB integrations (graphs, vector, etc.)

File: dapr_agents/workflow/mixins/state.py (hunk 23 above)
Comment (Cyb3rWard0g) [reply dapr_agents/workflow/mixins/state.py:119]: I will focus on logging in a separate PR. I would like to brainstorm a little bit more the strategy for that based on your experience building SDKs ;) Thank you!

File: dapr_agents/agents/durableagent/agent.py (hunk 26 above)
Comment (Cyb3rWard0g) [reply dapr_agents/agents/durableagent/agent.py:41]: Same as previous comment. I am removing that in a separate PR :) https://github.com/dapr/dapr-agents/pull/163

File: dapr_agents/agents/durableagent/agent.py (hunk 27 above)
Comment (Cyb3rWard0g) [reply dapr_agents/agents/durableagent/agent.py]: ✅ I added it to `AgentBase` here: https://github.com/dapr/dapr-agents/pull/161/commits/94a4e6d976650f2a3d4dc88a941735c88c43971a

File: dapr_agents/agents/telemetry/__init__.py (hunk 30 above)
Comment (Cyb3rWard0g) [reply dapr_agents/agents/telemetry/__init__.py:3]: I will do another PR for Otel. I will modify it then.

**Comment by Cyb3rWard0g** [comment]
> ```python
> workflow_state
> ```

Ahh got it. Actually, `state_key` and `agents_registry_key` already have defaults. https://github.com/dapr/dapr-agents/blob/1c832636eb4173dced965debc679ee42702b7cb3/dapr_agents/workflow/agentic.py#L58-L76

File: quickstarts/03-durable-agent-tool-call/durable_weather_agent.py [hunk 33]
```diff
@@ -0,0 +1,33 @@
...
+from weather_tools import tools
+import asyncio
+import logging
+
+
+async def main():
+    load_dotenv()
+    logging.basicConfig(level=logging.INFO)
+
+    # 1️⃣ Instantiate your agent (no .as_service())
```
sicoyle:
```suggestion
    # 1️⃣ Instantiate your agent
```

Cyb3rWard0g: Thank you @sicoyle :)

PR #168, opened 2025-07-30
Observability Module with OpenTelemetry and Phoenix Integration

Cyb3rWard0g:
This PR introduces a complete observability solution for Dapr Agents, enabling distributed tracing and monitoring through OpenTelemetry with Phoenix UI support. The module provides automatic instrumentation for agents, tools, LLM calls, and workflow executions while maintaining W3C Trace Context standards for distributed tracing across Dapr boundaries. All observability features are optional dependencies that gracefully degrade when not installed.

### New Observability Module (`dapr_agents/observability/`)
- **Automatic Instrumentation**: Zero-code tracing for agents, tools, LLM interactions, and workflows
- **Optional Dependencies**: Clean fallback behavior when observability packages aren't installed  
- **W3C Trace Context**: Standards-compliant context propagation across Dapr Workflow boundaries
- **Phoenix UI Integration**: Rich visualization and analysis through OpenInference semantic conventions

#### Instrumentor (`DaprAgentsInstrumentor`)
- Main entry point for enabling observability
- Automatic discovery and wrapping of key components
- Configurable span processors and exporters

#### Wrapper Classes
- **AgentWrapper**: Traces agent conversations and reasoning flows
- **LLMWrapper**: Captures LLM calls with token usage and message processing
- **ToolWrapper**: Monitors tool executions with input/output tracking
- **WorkflowWrapper**: Traces workflow orchestration and task execution
- **WorkflowTaskWrapper**: Detailed task-level tracing within workflows

#### Context Propagation (`context_propagation.py`)
- W3C Trace Context format support for Dapr serialization
- `extract_otel_context()` and `restore_otel_context()` utilities
- Proper parent-child span relationships across workflow boundaries

#### Message Processing (`message_processors.py`)
- Converts various message formats to OpenInference standard
- Tool schema extraction and serialization
- Token usage tracking and LLM response processing

#### Constants and Utilities
- OpenInference semantic conventions with fallback values
- Availability detection for optional dependencies
- Safe JSON serialization with error handling

### Documentation Updates
- Added observability section to quickstart README
- Docker Compose setup for Phoenix with PostgreSQL
- Step-by-step instrumentation guide
- Troubleshooting and best practices

#### Distributed Tracing
- Complete trace hierarchy from agent conversations to individual tool calls
- Context propagation across async boundaries and Dapr workflows
- Correlation IDs for grouping related operations

#### Performance Monitoring  
- Response times for all operations
- Token usage and cost tracking for LLM calls
- Error rates and failure analysis
- Tool execution performance metrics

#### Rich Visualization
- Phoenix UI compatibility with proper span relationships
- Message flow visualization with input/output content
- Tool schema display and parameter tracking
- Workflow execution timelines

#### Graceful Degradation
- Zero impact when observability packages not installed
- Automatic fallback to no-op implementations
- Clear error messages with installation guidance

#### W3C Trace Context Support
- Proper `traceparent` and `tracestate` header handling
- Compatible with Dapr's serialization mechanisms  
- Maintains trace continuity across workflow restarts

#### OpenInference Compliance
- Standard semantic conventions for AI/ML observability
- Proper message formatting for Phoenix UI
- Tool call tracking with function schemas

#### Optional Dependency Pattern
- Clean import handling with try/except blocks
- Availability flags throughout the codebase
- Helpful error messages guiding users to install extras

## Breaking Changes
None. All observability features are opt-in and don't affect existing functionality.


dapr_agents/observability/constants.py
```diff
@@ -0,0 +1,216 @@
...
+    """
+    if OPENINFERENCE_AVAILABLE and oi_safe_json_dumps:
+        # Use OpenInference implementation if available
+        return oi_safe_json_dumps(obj)
+    else:
+        # Fallback implementation
+        try:
+            return json.dumps(obj, default=str)
+        except Exception:
+            # Last resort - convert to string
```
sicoyle: rm comments pls - this is clear in the code itself

dapr_agents/observability/context_propagation.py
```diff
@@ -0,0 +1,205 @@
...
+        carrier["tracestate"] = ""
+
+    # Add individual components for debugging and verification
+    carrier["trace_id"] = trace_id
+    carrier["span_id"] = span_id
+    carrier["trace_flags"] = flags
+
+    logger.debug(
+        f"🔗 Extracted W3C context: traceparent={carrier.get('traceparent', '')}"
+    )
```
sicoyle, line 103: we can probably rm this

dapr_agents/observability/context_propagation.py
```diff
@@ -0,0 +1,205 @@
...
+
+    # Create span with proper parent-child relationship if context was restored
+    if parent_ctx:
+        # Context restored successfully - create child span maintaining trace hierarchy
+        return tracer.start_as_current_span(
+            span_name, context=parent_ctx, attributes=attributes
+        )
+    else:
+        # Context restoration failed - create root span (fallback behavior)
+        return tracer.start_as_current_span(span_name, attributes=attributes)
```
sicoyle: i think its fine to rm the comments as this is pretty clear in code

dapr_agents/observability/context_storage.py
```diff
@@ -0,0 +1,197 @@
...
+        """
+        with self._lock:
+            return {
+                "stored_instances": len(self._storage),
+                "instance_ids": list(self._storage.keys()),
+            }
+
+
+# Global instance for workflow context storage across the application
+_context_storage = WorkflowContextStorage()
```
sicoyle, line 134: do we really want this global to the entire application? An application can have multiple agents, therefore one agent could access another agents storage here potentially right? I'm not sure we want that. Can the workflow ids as keys be namespaced by agent name maybe to ensure that one agent cannot access another agents storage here?

dapr_agents/observability/instrumentor.py
```diff
@@ -0,0 +1,516 @@
...
+- WorkflowApp.run_and_monitor_workflow_async (AGENT span) - DurableAgent with lifecycle monitoring
+  └── DurableAgent.tool_calling_workflow (CHAIN span) - Workflow orchestration logic
+      ├── WorkflowTask.__call__ (TOOL span) - Tool execution via Dapr Workflow activities
+      └── WorkflowTask.__call__ (LLM span) - LLM calls via Dapr Workflow activities
+
+**Orchestrators (Workflow-based, Fire-and-Forget):**
+- WorkflowApp.run_workflow (AGENT span) - Orchestrator execution without monitoring
+  └── Orchestrator.main_workflow (CHAIN span) - Multi-agent coordination workflow
+      ├── WorkflowTask.__call__ (TASK span) - Workflow orchestration activities
+      └── WorkflowTask.__call__ (LLM span) - LLM calls via Dapr Workflow activities
```
sicoyle, line 27: hmm so durable agent and orchestrators start with WorkflowApp? Is that bc of the class setups? Is there a way to deliniate which orchestrator based on the span syntax at all?
I suppose a follow up too would be maybe if we can unify some of these spans across agent and durableagent :)

dapr_agents/observability/instrumentor.py
```diff
@@ -0,0 +1,516 @@
...
+        # Process iterations wrapper (CHAIN span - processing steps)
+        wrap_function_wrapper(
+            module="dapr_agents.agents.agent.agent",
+            name="Agent.process_iterations",
+            wrapper=ProcessIterationsWrapper(self._tracer),
+        )
+
+        # Tool execution batch wrapper (TOOL span - batch execution)
+        wrap_function_wrapper(
+            module="dapr_agents.agents.agent.agent",
```
sicoyle, line 392: why so many agents in the modules?

dapr_agents/observability/instrumentor.py
```diff
@@ -0,0 +1,516 @@
...
+
+        Instruments DurableAgent workflow execution to create comprehensive AGENT spans
+        that capture the complete workflow lifecycle from start to completion, enabling
+        detailed monitoring of workflow orchestration in Phoenix UI.
+        """
+        # Workflow execution wrapper for DurableAgent (AGENT span with monitoring)
+        # This captures the complete workflow lifecycle from start to completion
+        wrap_function_wrapper(
+            module="dapr_agents.workflow.base",
+            name="WorkflowApp.run_and_monitor_workflow_async",
```
sicoyle, line 416: so another PR enabled the sync start with run() on durable agent so should we specify async here?

dapr_agents/observability/wrappers/llm.py
```diff
@@ -0,0 +1,418 @@
...
+        # Build base span attributes
+        attributes = {
+            OPENINFERENCE_SPAN_KIND: LLM,
+            INPUT_MIME_TYPE: "application/json",
+            OUTPUT_MIME_TYPE: "application/json",
+            "agent.name": agent_name,
+        }
+
+        if model:
+            attributes[LLM_MODEL_NAME] = model
```
sicoyle, line 140: what if a user is running an app with say 3 agents within it and each using a diff model. How will this work if it is a single global const?

sicoyle (commented):
few comments so far. Overall loooking goood and super neat feature add! I'd say I don't think we have to be as verbose in the comments since the code is pretty clear, and in logs I think we can trim down on some and I'm not sure we want emojis in the logs. Runtime Dapr had comments recently saying to avoid emojis in print outs, so maybe we should follow suite there?

dapr_agents/observability/context_storage.py
Cyb3rWard0g: mmm can you elaborate on this? when we run dapr run --app-id `weatherapp` . that is 1 agent right? We use the instance_id / workflow_id here to aggregate logs.

dapr_agents/observability/instrumentor.py
Cyb3rWard0g: Yes, Durable Agent and Orchestrator have (AgenticWorkflow > WorkflowApp) classes underneath. However, they both (by default) start up in a different way. That could be unified in a separate PR and we can adjust the Otel wrappers ;) .

dapr_agents/observability/instrumentor.py
Cyb3rWard0g: oh because we are wrapping specific method of those modules. The Agent class is in the `dapr_agents.agents.agent.agent` module and we are wrapping a specific function from the `Agent` class to generate traces.

dapr_agents/observability/instrumentor.py
Cyb3rWard0g: So the new `run()` for DurableAgent class uses the async https://github.com/dapr/dapr-agents/blob/main/dapr_agents/agents/durableagent/agent.py#L111 and even if someone uses the `sync`, the `sync` is a wrapper around `async` https://github.com/dapr/dapr-agents/blob/main/dapr_agents/workflow/base.py#L551 😅

dapr_agents/observability/wrappers/llm.py
Cyb3rWard0g: That is a good question. I have not tested it with multiple agents and different modules.

dapr_agents/observability/context_storage.py
sicoyle: the dapr run cli cmd will run a single app https://docs.dapr.io/reference/cli/dapr-run/ but a single app in dapr agents world (app python code) can contain multiple agents technically since to some extent they are merely python objects a user can instantiate. 

Alternatively, you can have the agent with the `@task(agent=custom_agent,` syntax. Would we get otel metrics on the agent with the task syntax? 
For example something like this,
```python
# Define simple agents
extractor = Agent(
    name="DestinationExtractor",
    role="Extract destination",
    instructions=["Extract the main city from the user query"]
)

planner = Agent(
    name="PlannerAgent",
    role="Outline planner",
    instructions=["Generate a 3-day outline for the destination"]
)

expander = Agent(
    name="ItineraryAgent",
    role="Itinerary expander",
    instructions=["Expand the outline into a detailed plan"]
)

# Workflow tasks
@task(agent=extractor)
def extract(user_msg: str) -> str:
    pass

@task(agent=planner)
def plan(destination: str) -> str:
    pass

@task(agent=expander)
def expand(outline: str) -> str:
    pass

# Orchestration
@workflow(name="chained_planner_workflow")
def chained_planner_workflow(ctx: DaprWorkflowContext, user_msg: str):
    dest = yield ctx.call_activity(extract, input=user_msg)
    outline = yield ctx.call_activity(plan, input=dest)
    itinerary = yield ctx.call_activity(expand, input=outline)
    return itinerary
```



So yeah in a case such as this I'm not sure how the context sharing would be safe across all agents since they'd be within the same appid.

dapr_agents/observability/instrumentor.py
sicoyle: ah ok. thank you for clearing that up!

dapr_agents/observability/message_processors.py
```diff
@@ -0,0 +1,491 @@
...
+    Uses OpenInference helper functions when available, with fallback
+    to basic attribute structure for compatibility.
+
+    Args:
+        oi_messages: List of messages in OpenInference format
+
+    Returns:
+        Dict[str, Any]: Span attributes for input messages
+    """
+    if OPENINFERENCE_AVAILABLE:
```
sicoyle, line 223: i think this would be cleaner if instead of global var set we used a field on the agent or if this is enabled or yeah just a bool on one of the classes. Would that be doable?

dapr_agents/observability/message_processors.py
```diff
@@ -0,0 +1,491 @@
...
+    attributes: Dict[str, Any] = {}
+
+    if not tools or not isinstance(tools, list):
+        return attributes
+
+    for i, tool in enumerate(tools):
+        try:
+            # Use AgentTool's built-in function call format
+            if hasattr(tool, "to_function_call"):
+                function_call = tool.to_function_call(format_type="openai")
```
sicoyle, line 271: is it the case that every llm provider that we have abides by the openai format?

dapr_agents/observability/message_processors.py
```diff
@@ -0,0 +1,491 @@
...
+
+    Args:
+        oi_message: Single message in OpenInference format
+
+    Returns:
+        Dict[str, Any]: Basic output message attributes
+    """
+    attributes = {}
+    if oi_message:
+        attributes[f"{LLM_OUTPUT_MESSAGES}.0.{MESSAGE_ROLE}"] = oi_message.get(
```
sicoyle, line 407: any way to make these fields on a class too or do we want all of these as global vars?

dapr_agents/observability/utils.py
```diff
@@ -0,0 +1,310 @@
...
+    Returns:
+        Dict[str, Any]: Filtered arguments without self/cls
+
+    Example:
+        >>> strip_method_args({'self': obj, 'param': 'value', 'cls': MyClass})
+        {'param': 'value'}
+    """
+    return {
+        key: value for key, value in arguments.items() if key not in ("self", "cls")
+    }
```
sicoyle, line 102: is smolagents pattern the ideal pattern we're following? why?

dapr_agents/observability/wrappers/tool.py
```diff
@@ -0,0 +1,350 @@
...
+            args (tuple): Positional arguments - typically (tool_name, tool_args)
+            kwargs (dict): Keyword arguments passed to the original method
+
+        Returns:
+            Any: Result from wrapped method execution with span attributes capturing
+                 individual tool execution context, inputs, outputs, and tool metadata
+        """
+        # Check for instrumentation suppression
+        if context_api and context_api.get_value(
+            context_api._SUPPRESS_INSTRUMENTATION_KEY
```
sicoyle, line 236: what does this do?

dapr_agents/observability/wrappers/tool.py
```diff
@@ -0,0 +1,350 @@
...
+                try:
+                    result = await wrapped(*args, **kwargs)
+                    span.set_attribute(OUTPUT_VALUE, str(result))
+                    span.set_attribute(OUTPUT_MIME_TYPE, "application/json")
+                    span.set_status(Status(StatusCode.OK))
+                    return result
+                except Exception as e:
+                    span.set_status(Status(StatusCode.ERROR, str(e)))
+                    span.record_exception(e)
+                    raise
```
sicoyle, line 303: seems like some repeat logic in some of these funcs. Is there a way to reuse some of it through helpers?

dapr_agents/observability/wrappers/workflow.py
```diff
@@ -0,0 +1,371 @@
...
+            kwargs: Keyword arguments
+
+        Returns:
+            str: Workflow instance ID from wrapped method execution
+        """
+        # Check for instrumentation suppression
+        if context_api and context_api.get_value(
+            context_api._SUPPRESS_INSTRUMENTATION_KEY
+        ):
+            return wrapped(*args, **kwargs)
```
sicoyle, line 76: can these be in like a pre_init type python class function similar to how we use the post init validations? would that clean some of the duplicate things like this here and there?

dapr_agents/observability/wrappers/workflow.py
```diff
@@ -0,0 +1,371 @@
...
+
+        return attributes
+
+
+# ============================================================================
+# Workflow Monitor Wrapper
+# ============================================================================
+
+
+class WorkflowMonitorWrapper:
```
sicoyle, line 171: can we do a class per file pls?

dapr_agents/observability/wrappers/workflow.py
```diff
@@ -0,0 +1,371 @@
...
+        # Extract arguments
+        arguments = bind_arguments(wrapped, *args, **kwargs)
+        workflow = arguments.get("workflow")
+
+        # Extract workflow name
+        workflow_name = (
+            workflow
+            if isinstance(workflow, str)
+            else getattr(workflow, "__name__", "unknown_workflow")
+        )
```
sicoyle, line 229: maybe we could have a base wrapper workflow class so this class and the one above dont both need some of these checks?

dapr_agents/tool/http/client.py
```diff
@@ -7,9 +7,20 @@
...
+    from opentelemetry._logs import set_logger_provider
+
+    OPENTELEMETRY_AVAILABLE = True
+except ImportError:
+    # OpenTelemetry is not available - tracing will be disabled
+    # To enable observability features, install: pip install dapr-agents[observability]
+    RequestsInstrumentor = None
+    trace = None
+    set_logger_provider = None
+    OPENTELEMETRY_AVAILABLE = False
```
sicoyle, line 23: maybe we could do something like this instead?
have an `otel_support.py` with the
```
try:
    from opentelemetry.instrumentation.requests import RequestsInstrumentor
    from opentelemetry import trace
    from opentelemetry._logs import set_logger_provider

    OPENTELEMETRY_AVAILABLE = True
except ImportError:
    RequestsInstrumentor = None
    trace = None
    set_logger_provider = None
    OPENTELEMETRY_AVAILABLE = False

```
then do this at the top of these files
```
from . import otel_support

if otel_support.OPENTELEMETRY_AVAILABLE:
    otel_support.RequestsInstrumentor().instrument()

```

that way you can have the core checks in one place that we still have to bring in the import to check, but then all the checks for all these files are not copy/pasted in a few places. I just worry we might miss lines or these imports could get out of sync eventually ya know

pyproject.toml
```diff
@@ -69,6 +65,14 @@ dev = [
     "mypy>=1.15.0,<2.0.0",
     "tox>=4.0.0,<5.0.0",
 ]
+observability = [
```
sicoyle, line 68: thank youuuu!

quickstarts/03-agent-tool-call/README.md
```diff
@@ -240,6 +240,171 @@ if __name__ == "__main__":
...
+First, deploy Phoenix Arize server using Docker Compose with PostgreSQL backend for persistent storage.
+
+#### Prerequisites
+
+- Docker and Docker Compose installed on your system
+- Verify Docker is running: `docker info`
+
+#### Deploy Phoenix with PostgreSQL
+
+1. Create a `docker-compose.yml` file for Phoenix with PostgreSQL backend:
```
sicoyle: instead can you link to the docker compose file you added here. Ppl dont have to create this bc you did it for them :)

quickstarts/03-agent-tool-call/docker-compose.yml
```diff
@@ -0,0 +1,26 @@
+# docker-compose.yml
+services:
+  phoenix:
+    image: arizephoenix/phoenix:latest # Must be greater than 4.0 version to work
```
sicoyle, line 4: pls pin the tag. We should not use latest really ever bc if phoenix upstream breaks things in their latest img tag then we would get the issue. So you can rm the comment and pin to >= v4.0

quickstarts/03-agent-tool-call/weather_agent_tracing.py
```diff
@@ -0,0 +1,39 @@
...
+    # Register Dapr Agents with Phoenix OpenTelemetry
+    tracer_provider = register(
+        project_name="dapr-weather-agents",
+        protocol="http/protobuf",
+    )
+    # Initialize Dapr Agents OpenTelemetry instrumentor
+    instrumentor = DaprAgentsInstrumentor()
+    instrumentor.instrument(tracer_provider=tracer_provider, skip_dep_check=True)
+
+    await AIAgent.run("What is the weather in Virginia, New York and Washington DC?")
```
sicoyle, line 35: very nice and slick

quickstarts/03-durable-agent-tool-call/README.md
```diff
@@ -99,3 +99,181 @@ Durable agents maintain state across runs, enabling workflows that require persi
...
+
+#### Deploy Phoenix with PostgreSQL
+
+1. Create a `docker-compose.yml` file for Phoenix with PostgreSQL backend:
+
+```yaml
+# docker-compose.yml
+services:
+  phoenix:
+    image: arizephoenix/phoenix:latest # Must be version 4.0 or greater
```
sicoyle: pin to v4.0+ pls on img tag

quickstarts/03-durable-agent-tool-call/README.md
```diff
@@ -99,3 +99,181 @@ Durable agents maintain state across runs, enabling workflows that require persi
...
+First, deploy Phoenix Arize server using Docker Compose with PostgreSQL backend for persistent storage.
+
+#### Prerequisites
+
+- Docker and Docker Compose installed on your system
+- Verify Docker is running: `docker info`
+
+#### Deploy Phoenix with PostgreSQL
+
+1. Create a `docker-compose.yml` file for Phoenix with PostgreSQL backend:
```
sicoyle: also just link to your compose instead of saying to create one pls

quickstarts/03-durable-agent-tool-call/README.md
```diff
@@ -99,3 +99,181 @@ Durable agents maintain state across runs, enabling workflows that require persi
...
+
+#### Install Observability Dependencies
+
+Update your `requirements.txt` to include observability features:
+
+```txt
+dapr-agents[observability]>=0.7.1
+python-dotenv
+arize-phoenix
+```
```
sicoyle: this is in the requirements already so can rm  and just have the pip install

quickstarts/03-durable-agent-tool-call/docker-compose.yml
```diff
@@ -0,0 +1,26 @@
+# docker-compose.yml
+services:
+  phoenix:
+    image: arizephoenix/phoenix:latest # Must be greater than 4.0 version to work
```
sicoyle, line 4: pin img tag pls

dapr_agents/observability/context_propagation.py
Cyb3rWard0g: actually for debugging purposes, I think we should keep it.

dapr_agents/observability/context_storage.py
Cyb3rWard0g: OMG! I ran your example (I had to do a minor fix to the agent as a task which was not related to observabiity), and it worked! I loved that it kept everything under One trace starting from the "Workflow". Look:

image

image

Cyb3rWard0g: The Observabiity module was able to trace and connect each actions from the WorkflowApp as the root node. Then 3 tasks were identified and traced as workflow tasks . However, each task spawned an AI Agent. Each agent went through 1 iteration / 1 loop and each loop requested a chat completion. 🔥 Adding @yaron2 ;)

Cyb3rWard0g: The workflow executing successfully too based on the app logs ;) 
image

dapr_agents/observability/wrappers/llm.py
Cyb3rWard0g: I tested this scenario too :) and same everything is captured the right way: 

image

image

image

Cyb3rWard0g: 😉 @yaron2

dapr_agents/observability/message_processors.py
Cyb3rWard0g: mmm I think global makes more sense tbh since we are defining wrappers for all methods used by Agents , Durable Agents, and any agentic workflows. With the previous comments you had about potential issues on having multiple agents under the same app and multi model tasks, everything works as expected.

dapr_agents/observability/wrappers/tool.py
Cyb3rWard0g: It sets `suppress_instrumentation` in the open telemetry context

```
_SUPPRESS_INSTRUMENTATION_KEY = create_key("suppress_instrumentation")
```

that is then used somewhere else to set another key

```
def _instrumented_requests_call(
        method: str, url: str, call_wrapped, get_or_create_headers
    ):
        if context.get_value("suppress_instrumentation") or context.get_value(
            _SUPPRESS_REQUESTS_INSTRUMENTATION_KEY
        ):
            return call_wrapped()
```

And that key ` _SUPPRESS_REQUESTS_INSTRUMENTATION_KEY` is a:

```
# A key to a context variable to avoid creating duplicate spans when instrumenting
# both, Session.request and Session.send, since Session.request calls into Session.send
_SUPPRESS_REQUESTS_INSTRUMENTATION_KEY = "suppress_requests_instrumentation"
```

Some context from OpenTelemetry docs: https://opentelemetry-python-kinvolk.readthedocs.io/en/latest/_modules/opentelemetry/instrumentation/requests.html and where it is set: https://github.com/pexip/os-python-opentelemetry-api/blob/bad159831b8ba321068a4a6b06c282c8737b94a4/src/opentelemetry/context/__init__.py#L171

quickstarts/03-durable-agent-tool-call/README.md
Cyb3rWard0g: got it ! thank you

dapr_agents/observability/utils.py
Cyb3rWard0g: compared to other openinference packages for other frameworks, smolagents is easy to follow to learn how instrumentation / observability is enabled.

dapr_agents/observability/message_processors.py
Cyb3rWard0g: Global vars make sense tbh after testing multiple scenarios ;)

dapr_agents/observability/message_processors.py
Cyb3rWard0g: good question. I need to do more research on that.

PR #169, opened 2025-07-31
refactor: rm check on if dapr is running for durableagent

sicoyle:
We should only warn if dapr is not on localhost. I did not see an env var upstream for us to grab the host https://docs.dapr.io/reference/environment/ 

I know there was a prior proposal on this upstream https://github.com/dapr/proposals/pull/27/files but I don't see all of these reflect in the official docs and see varying things in python sdk vs CLI tests, etc... so for now this PR just "warns" if dapr is not on localhost, and then i saw an openapi spec parser that we can remove since i think i just missed this in removing the openapiagent.


yaron2 (approved):

yaron2:
Lint

PR #177, opened 2025-08-13
fix: speed up deps installed + handle errs in quickstarts better

sicoyle:
Note: I did check / run 99% of every quickstart/requirements files I updated :) 

This PR aims to reduce the amount of time that our dependencies take to install. I was clocking things in the dapr university agents course at 2.5min for a pip install on agents. Because I checked the quickstarts too, I added better err handling, updated with an explicit line on some of them pinning to the cheapest llm for openai which i thought would be good for end users when trying the quickstarts out, and updated our requirements.txt file with a comment we can uncomment to use our local agents dependencies when we're developing on the repo.

I learned that apparently pinning specific versions, esp on larger deps speeds things up so we should try to keep that in mind moving forward, and ensure that if deps are optional that they are not in the root dependencies in our pyproject.toml file 😇 I also went through and moved duplicate deps in our quickstarts that are already in the pyproject toml and moved more to be optional. Also, updating our versions on some deps has helped reduce the timing/size of the imports.

Some searching and testing things locally, I see that using UV locally for my deps takes 13seconds vs the 2.5min for pip install... so UV clearly has some good optimizations.


yaron2:
Conflict

sicoyle:
addressed :)

PR #185, opened 2025-08-27
fix(tracing): enable tracing on durable agent + add quickstart

sicoyle:
I was looking at the different activities that make up the durable agent and diagrammed it out in tldraw, but then realized that I can look at it in the Phoenix UI too. In looking at things in the Phoenix UI, I saw there was a bug on DurableAgent tracing results where the context we not properly passing. I kept getting "No context..." errors for my workflow instance ID. This is what it looked like in the UI
image

Now, the Phoenix UI properly shows the context propagation
image

The issue with the otel context propagation was the workflow would start and then the workflow tasks would execute, but the parent context in my logs said "No context found". This was because we instrumented the application, but the workflow wrappers for the telemetry were not applying and failing silently. This is due to the wrapt library wrapper functions having issues with bounded methods (instance methods) on pydantic models - and also explains why regular agent class had no issues bc regular agent class wrapped regular python class methods, but WorkflowApp class inherits from Pydantic since its a BaseModel class and wrap_function_wrapper doesnt work well with wrapt. I had to register a callback when the workflow runtime actually started so they were available to then be instrumented - in other words, we need lazy instrumentation to ensure workflow classes are fully loaded before applying wrappers.

I also updated the quickstart to include a tracing example for the durable agent as that's what I was testing with, and renamed the regular agent class process_iterations -> conversation so in the UI it will show `agent.conversation()` instead of `agent.process_iterations()`.  I also made some of the strings to use the actual class methods instead for the method wrapping for telemetry so if we update the function names then they are reflected on telemetry side of things as well.

### DIFF
--- a/dapr_agents/agents/agent/agent.py
+++ b/dapr_agents/agents/agent/agent.py
@@ -101,7 +101,7 @@ class Agent(AgentBase):
             )
 
         # Process conversation iterations and return the result
-        return await self.process_iterations(messages)
+        return await self.conversation(messages)
 
     async def execute_tools(self, tool_calls: List[ToolCall]) -> List[ToolMessage]:
         """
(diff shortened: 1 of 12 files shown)


Cyb3rWard0g (approved):
LGTM Thank you @sicoyle !

Cyb3rWard0g:
than kyou @sicoyle ! would you mind taking a look at the Lint and Build errors. Please and thank you.

Issue #196, opened 2025-09-04
Concurrent workflow instances corrupt shared memory session


bibryam:
**Description**
When multiple pubsub messages are sent to a DurableAgent simultaneously, the concurrent workflow instances share the same memory session, leading to corrupted conversation history and OpenAI API failures.

**Current Behavior**

1. Multiple pubsub messages trigger concurrent workflow instances
2. All workflows share the same memory session (session-headless-{uuid})
3. Tool calls and responses from different workflows get mixed up in the shared conversation history
4. OpenAI API rejects malformed message sequences where tool calls don't have corresponding tool responses
5. All workflows fail with Failed during chat generation: Failed to process chat completion

**Possible Solutions that come to mind**
**Option 1:** Sequential Processing
Process messages sequentially, waiting for each workflow to complete before starting the next one.

**Option 2:** Isolated Memory Sessions
Provide a mechanism to start a new memory session for each workflow instance, ensuring conversation history isolation.

**Sample Code** 
https://github.com/diagrid-labs/dapr-agents-catalyst-samples/tree/main/02_durable-agent-headless 

**ErrorLog** attached

bibryam:
[196-log.txt](https://github.com/user-attachments/files/22143289/196-log.txt)

Issue #193, opened 2025-09-02
Durable Agent workflow fails after crash recovery


bibryam:
When the Dapr sidecar is stopped during a tool execution and restarted, the Durable Agent workflow fails to recover and goes into a failed state.

**Code Sample**:

```python
#!/usr/bin/env python3
"""
Stateful Augmented LLM Pattern demonstrates:
1. Memory - remembering user preferences
2. Tool use - accessing external data
3. LLM abstraction
4. Durable execution of tools as workflow actions
"""
import asyncio
import logging
import time  
from typing import List
from pydantic import BaseModel, Field
from dapr_agents import tool, DurableAgent, OpenAIChatClient
from dapr_agents.memory import ConversationDaprStateMemory
from dotenv import load_dotenv


class FlightOption(BaseModel):
    airline: str = Field(description="Airline name")
    price: float = Field(description="Price in USD")


class DestinationSchema(BaseModel):
    destination: str = Field(description="Destination city name")


@tool(args_model=DestinationSchema)
def search_flights(destination: str) -> List[FlightOption]:
    """Search for flights to the specified destination."""
    time.sleep(10)
    return [
        FlightOption(airline="SkyHighAir", price=450.00),
        FlightOption(airline="GlobalWings", price=375.50),
    ]


async def main():
    try:
        travel_planner = DurableAgent(
            name="TravelBuddy",
            role="Travel Planner",
            goal="Help users find flights and remember preferences",
            instructions=[
                "Find flights to destinations",
                "Remember user preferences",
                "Provide clear flight info",
            ],
            tools=[search_flights],
            message_bus_name="message-pubsub",
            state_store_name="execution-state",
            agents_registry_store_name="registry-state",
            memory=ConversationDaprStateMemory(
                store_name="memory-state", session_id="my-unique-session-id"
            ),
            llm=OpenAIChatClient(model="gpt-3.5-turbo"),
        )

        travel_planner.as_service(port=8001)
        await travel_planner.start()
        print("Travel Planner Agent is running")

    except Exception as e:
        print(f"Error starting service: {e}")


if __name__ == "__main__":
    load_dotenv()
    logging.basicConfig(level=logging.INFO)
    asyncio.run(main())
```

**Error Log (Excerpt)**:

```
store to scheduler successfully  app_id=durable-agent-headless instance=b1.lan scope=dapr.runtime.actors.reminders.migration type=log ver=1.15.4
[...]

PR #197, opened 2025-09-04
fix: enable durable agent to keep existing workflow upon 

sicoyle:
This ensures that if someone uses runs a durable agent (starting via .run(), .start(), or pubsub) that we properly maintain workflow_state on the agent to be able to see if we have an instance id already set to use upon agent startup again. Basically, if a user started their app with an agent, and then the underlying workflow started, and then they killed the app, then restarted the app, the durable agent used to always create a new workflow under the hood in addition to ensuring the proper completion of the already started workflow from before. With these changes, we check upon startup if there is an instance id already and check if its status needs updating. The downfall here is with our current workflow_state key setup, this state data can get beefy; however, for now it is fine as this fixes durability, but in a follow up effort I will revisit a better state setup (IE later this week).

I've moved our graceful shutdown logic to a proper class and also added a hook that upon app termination (for whatever reason) we save workflow instance id state.

This PR along the way also showed me that we were maintaining tool call and chat history in global memory, but in the case of the scenario above where I start an agent (and therefore a workflow) and then kill the app and restart it, then the in memory data is lost and so then if the workflow under the hood resumed and another kicks off, then the chat history/tool history was not correct and missed the in memory parts or combining the workflow instance id chat history incorrectly/interleaving, so I've moved this to the workflow store state for now. I've also moved our chat history to be instance specific through the messages field on the instances. Leaving it as it was resulted in OpenAI errors due to then the chat history being invalid as they have certain expectations on conversation flow that this case would then break. This proves that the memory must be persistent, and again, I will evaluate a better solution for this in a follow up PR :) Also, added more info on the openai err bc I had no idea that this state issue was the underlying issue, so we should bubble up a bit more details that I've accounted for here too.

issue: https://github.com/dapr/dapr-agents/issues/193

This actually should also fix https://github.com/dapr/dapr-agents/issues/196

For workflows that get reran they show the agent span gets created and whatever activities that ran before we terminate the app. For example:
image
You can see i started the app and it created an agent span that ran for a bit and then i killed the app so the span was not fully completed before i killed the app. Upon resuming the app, a new workflow is created and i allowed it to run to completion, hence the secondary full agent span.
image

See the uncompleted agent span that is expected
image

See the then completed agent span for me rerunning the app and creating a secondary workflow that ran to completion
image

If you go to the traces section of the UI then you can see the first agent / workflow completion traces here. I think in future maybe we can look to add this back to the initial agent span, but technically this is complete in the sense that the workflow ran to completion and the remaining activities stayed together with the span it picked up on which is what i would expect as with OpenTelemetry spans, when a process is killed, active spans cannot properly close from what i understand so it is an unset trace status as shown (with the full context of the trace activities connected/completing the workflow).
image

Behaviors are documented in the developer docs readme file in the files changed in this PR so we can reference in future.

### DIFF
--- a/quickstarts/01-hello-world-observability/requirements.txt
+++ b/quickstarts/01-hello-world-observability/requirements.txt
@@ -1,4 +1,4 @@
-dapr-agents>=0.8.4
+dapr-agents
 opentelemetry-sdk>=1.25.0
 opentelemetry-exporter-zipkin-json==1.25.0
 opentelemetry-exporter-otlp>=1.25.0
--- a/quickstarts/01-hello-world/requirements.txt
+++ b/quickstarts/01-hello-world/requirements.txt
@@ -1,3 +1,3 @@
-dapr-agents>=0.8.4
+dapr-agents
 # For local development use local changes by commenting out the line above and uncommenting the line below:
 # -e ../../
--- a/quickstarts/02_llm_call_dapr/requirements.txt
+++ b/quickstarts/02_llm_call_dapr/requirements.txt
@@ -1,3 +1,3 @@
-dapr-agents>=0.8.4
+dapr-agents
 # For local development use local changes by commenting out the dapr-agents line above and uncommenting the line below:
 # -e ../../
--- a/quickstarts/02_llm_call_elevenlabs/requirements.txt
+++ b/quickstarts/02_llm_call_elevenlabs/requirements.txt
@@ -1,4 +1,4 @@
-dapr-agents>=0.8.4
+dapr-agents
 elevenlabs
 # For local development use local changes by commenting out the dapr-agentsline above and uncommenting the line below:
 # -e ../../
--- a/quickstarts/02_llm_call_hugging_face/requirements.txt
+++ b/quickstarts/02_llm_call_hugging_face/requirements.txt
@@ -1,3 +1,3 @@
-dapr-agents>=0.8.4
+dapr-agents
 # For local development use local changes by commenting out the dapr-agentsline above and uncommenting the line below:
 # -e ../../
--- a/quickstarts/02_llm_call_nvidia/requirements.txt
+++ b/quickstarts/02_llm_call_nvidia/requirements.txt
@@ -1,4 +1,4 @@
-dapr-agents>=0.8.4
+dapr-agents
 tiktoken==0.9.0
 # For local development use local changes by commenting out the dapr-agents line above and uncommenting the line below:
 # -e ../../
--- a/quickstarts/02_llm_call_open_ai/requirements.txt
+++ b/quickstarts/02_llm_call_open_ai/requirements.txt
@@ -1,4 +1,4 @@
-dapr-agents>=0.8.4
+dapr-agents
 tiktoken==0.9.0
 # For local development use local changes by commenting out the dapr-agents line above and uncommenting the line below:
 # -e ../../
--- a/quickstarts/03-agent-tool-call/requirements.txt
+++ b/quickstarts/03-agent-tool-call/requirements.txt
@@ -1,4 +1,4 @@
-dapr-agents[observability]>=0.8.3
+dapr-agents[observability]
 arize-phoenix>=11.22.0,<12.0.0
 arize-phoenix-otel>=0.12.0,<0.13.0
 opentelemetry-instrumentation>=0.55b1,<0.56b0
--- a/quickstarts/03-durable-agent-multitool-dapr/requirements.txt
+++ b/quickstarts/03-durable-agent-multitool-dapr/requirements.txt
@@ -1,4 +1,4 @@
-dapr-agents>=0.8.4
+dapr-agents
 
 # comment out this line if you are not developing dapr-agents locally
 #-e ../.. # dapr-agents local development
--- a/quickstarts/03-durable-agent-tool-call/requirements.txt
+++ b/quickstarts/03-durable-agent-tool-call/requirements.txt
@@ -1,4 +1,4 @@
-dapr-agents>=0.8.4
+dapr-agents
 arize-phoenix-otel>=0.12.0,<0.13.0
 # For local development use local changes by commenting out the dapr-agents line above and uncommenting the line below:
 # -e ../../
--- a/quickstarts/04-llm-based-workflows/requirements.txt
+++ b/quickstarts/04-llm-based-workflows/requirements.txt
@@ -1,3 +1,3 @@
-dapr-agents>=0.8.4
+dapr-agents
 # For local development use local changes by commenting out the dapr-agents line above and uncommenting the line below:
 # -e ../../
--- a/quickstarts/05-multi-agent-workflows/requirements.txt
+++ b/quickstarts/05-multi-agent-workflows/requirements.txt
@@ -1,3 +1,3 @@
-dapr-agents>=0.8.4
+dapr-agents
 # For local development use local changes by commenting out the dapr-agents line above and uncommenting the line below:
 # -e ../../
--- a/quickstarts/06-document-agent-chainlit/requirements.txt
+++ b/quickstarts/06-document-agent-chainlit/requirements.txt
@@ -1,4 +1,4 @@
-dapr-agents>=0.8.4
+dapr-agents
 chainlit==2.6.8
 unstructured[all-docs]==0.18.11
 # For local development use local changes by commenting out the dapr-agents line above and uncommenting the line below:
--- a/quickstarts/07-agent-mcp-client-sse/requirements.txt
+++ b/quickstarts/07-agent-mcp-client-sse/requirements.txt
@@ -1,4 +1,4 @@
-dapr-agents>=0.8.4
+dapr-agents
 starlette==0.47.2
 websockets>=12.0
 # For local development use local changes by commenting out the dapr-agents line above and uncommenting the line below:
--- a/quickstarts/07-agent-mcp-client-stdio/requirements.txt
+++ b/quickstarts/07-agent-mcp-client-stdio/requirements.txt
@@ -1,4 +1,4 @@
-dapr-agents>=0.8.4
+dapr-agents
 websockets>=12.0
 # For local development use local changes by commenting out the dapr-agents line above and uncommenting the line below:
 # -e ../../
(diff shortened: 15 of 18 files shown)


Cyb3rWard0g:
LGTM @sicoyle 🚀

Cyb3rWard0g (approved):

sicoyle:
actually give me a little longer on this PR. will ping when ready for final review just need to look at final state pieces first

Cyb3rWard0g:
is it possible to simply append the new trace to the existing span? If the last part is runs after the main one, why not appending it to the original one? I am thinking about evaluations here, and I believe that you would need to keep everything together under one span right?

sicoyle:
This is ready. See my PR description for where I had tracing on resumed workflows. That I've since improved in my latest commits; however, it still is not 100% perfect on resumed workflows as they will have the workflow task traces, butttt not the complete agent span which I think is actually technically fine since the spans represent a unit of work during a single process lifetime with a clear start/end, and since a resumed workflow instance is done within an app that was terminated and then resumed I think this is fine. I can create an issue tomorrow to follow up on this if we want, but it has been a very taxing effort to get to this point on tracing for resumed workflows. This is what it looks like now:
image

The bottom is the workflow tasks on the workflow that is interrupted (so you can see it is basically missing the wrapping agent span on the workflow tracing), whereas the full agent span is for the workflow that I kept the app alive for throughout the workflows duration. 

See here where we do get the full workflow activities all linked, it's just with the workflow wrapper class I cannot find a way to add the agent span easily for resumed workflows and I have longer than I would have hoped on it at this point 😅 
image

I think it's basically the case that resumed workflows are bypassing WorkflowMonitorWrapper that does the proper agent span wrapping so a resumed workflow does not get the agent span wrapping, but then trying to add full / long lived agent spans for resumed workflows is difficult because we can't use blocking context managers in the task wrapper, and we can't easily intercept at the workflow level for a resumed workflow. The best thing in the future for this I thinkkkk is to modeify the workflow resumption logic itself to go through the WOrkflowMonitorWrapper class, but that will for sure require deeper changes to the workflow integrations there.

dapr_agents/types/llm.py
```diff
@@ -16,13 +16,6 @@ class ElevenLabsClientConfig(BaseModel):
         None, description="API key to authenticate with the ElevenLabs API."
     )
 
-    @field_validator("*", mode="before")
-    @classmethod
-    def none_to_default(cls, v):
-        if v is None:
-            raise PydanticUseDefault()
-        return v
-
```
sicoyle, line 25: pls note i removed these intentionally to correct an exception i was seeing in the debugger

pyproject.toml
```diff
@@ -37,6 +37,7 @@ dependencies = [
     "mcp>=1.7.1,<2.0.0",
     "python-dotenv>=1.1.1,<2.0.0",
     "posthog<6.0.0",
+    "nltk>=3.8.0,<4.0.0",
```
sicoyle, line 40: this was another thing my debugger caught bc this dep was missing

Cyb3rWard0g: What requires `nltk`?

Cyb3rWard0g: Have you also tested the MCP quickstarts? specifically the `websockets` dependency. When I run the latest version and test MCP, I get:

```
.venv/lib/python3.13/site-packages/dapr_agents/tool/mcp/transport.py:8
      6 from mcp.client.sse import sse_client
      7 from mcp.client.streamable_http import streamablehttp_client
----> 8 from mcp.client.websocket import websocket_client
...
----> 9 from websockets.asyncio.client import connect as ws_connect
     10 from websockets.typing import Subprotocol
     12 import mcp.types as types

ModuleNotFoundError: No module named 'websockets'
```

sicoyle: no, ill test the rest on this PR then. I tested the two hello worlds, 03-durable-agent-tool-call, 04-llm-based-workflows, 05-multi-agent-workflows, 09-agents-as-tasks-in-workflows

I'll circle back on the nltk dep - but I set a dependency on my debugger that showed uncaught exceptions and this was one of the things it complained on... I'll circle back after a few meetings

sicoyle:
i tested a chainlit chat ex that proves session_id info is preserved if i restart my chainlit ui and/or kill my dapr run cmd and then rerun it and ask about the previous wf context
image

I checked every quickstart except nvidia/elevenlab 02-* ones bc i checked the other 02-* ones (and I don't have api keys to those...). You'll see a fix on the streamablehttp quickstart where i added graceful shutdown bc i saw a ton of err lines in terminal upon shutdown, so that's handled gracefully now.


See logs go from
```
                         started                                             
INFO:     Started server process [62104]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
^CINFO:     Shutting down
INFO:     Waiting for application shutdown.
INFO:     Application shutdown complete.
INFO:     Finished server process [62104]
[09/17/25 15:02:51] INFO     StreamableHTTP        streamable_http_manager.py:114
                             session manager                                     
                             shutting down                                       
Traceback (most recent call last):
  File "/opt/homebrew/Cellar/python@3.13/3.13.3/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/opt/homebrew/Cellar/python@3.13/3.13.3/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/base_events.py", line 719, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/Users/samcoyle/go/src/github.com/dapr-agents/quickstarts/07-agent-mcp-client-streamablehttp/server.py", line 102, in serve
    async with session_manager.run():
               ~~~~~~~~~~~~~~~~~~~^^
  File "/opt/homebrew/Cellar/python@3.13/3.13.3/Frameworks/Python.framework/Versions/3.13/lib/python3.13/contextlib.py", line 221, in __aexit__
    await anext(self.gen)
  File "/Users/samcoyle/go/src/github.com/dapr-agents/quickstarts/07-agent-mcp-client-streamablehttp/venvmybranch/lib/python3.13/site-packages/mcp/server/streamable_http_manager.py", line 107, in run
    async with anyio.create_task_group() as tg:
               ~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/samcoyle/go/src/github.com/dapr-agents/quickstarts/07-agent-mcp-client-streamablehttp/venvmybranch/lib/python3.13/site-packages/anyio/_backends/_asyncio.py", line 764, in __aexit__
    await AsyncIOBackend.cancel_shielded_checkpoint()
  File "/Users/samcoyle/go/src/github.com/dapr-agents/quickstarts/07-agent-mcp-client-streamablehttp/venvmybranch/lib/python3.13/site-packages/anyio/_backends/_asyncio.py", line 2356, in cancel_shielded_checkpoint
    await sleep(0)
  File "/opt/homebrew/Cellar/python@3.13/3.13.3/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/tasks.py", line 706, in sleep
    await __sleep0()
  File "/opt/homebrew/Cellar/python@3.13/3.13.3/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/tasks.py", line 700, in __sleep0
    yield
asyncio.exceptions.CancelledError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/samcoyle/go/src/github.com/dapr-agents/quickstarts/07-agent-mcp-client-streamablehttp/server.py", line 192, in <module>
    main()
    ~~~~^^
  File "/Users/samcoyle/go/src/github.com/dapr-agents/quickstarts/07-agent-mcp-client-streamablehttp/server.py", line 188, in main
    start_server(args)
    ~~~~~~~~~~~~^^^^^^
  File "/Users/samcoyle/go/src/github.com/dapr-agents/quickstarts/07-agent-mcp-client-streamablehttp/server.py", line 167, in start_server
    server.run_streamable_http(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^
        host=args.host,
        ^^^^^^^^^^^^^^^
    ...<3 lines>...
        json_response=args.json_response,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/samcoyle/go/src/github.com/dapr-agents/quickstarts/07-agent-mcp-client-streamablehttp/server.py", line 109, in run_streamable_http
    asyncio.run(serve())
    ~~~~~~~~~~~^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.13/3.13.3/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/runners.py", line 195, in run
    return runner.run(main)
           ~~~~~~~~~~^^^^^^
  File "/opt/homebrew/Cellar/python@3.13/3.13.3/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/runners.py", line 123, in run
    raise KeyboardInterrupt()

```
to
```

TO THIS (fix-da-durability) ✗ 
python server.py --server_type streamable-http --port 8000
[09/17/25 15:04:25] INFO     Starting MCP server in streamable HTTP  server.py:91
                             mode (stateless=False,                              
                             json_response=False, resumable=False)               
                    INFO     StreamableHTTP        streamable_http_manager.py:110
                             session manager                                     
                             started                                             
INFO:     Started server process [63586]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
^CINFO:     Shutting down
INFO:     Waiting for application shutdown.
INFO:     Application shutdown complete.
INFO:     Finished server process [63586]
[09/17/25 15:04:27] INFO     Received signal 2, shutting down       server.py:125
                             gracefully...                                       
                    INFO     StreamableHTTP        streamable_http_manager.py:114
                             session manager                                     
                             shutting down              
```
for that quickstart in particular

pyproject.toml
sicoyle: i added the websocket dep to all of the 07-agent-mcp-client-* quickstarts that were missing it so those are all g2g!
Also, the `nltk` pkg is used by the text splitter at `dapr_agents/document/splitter/base.py`.

Issue #198, opened 2025-09-05
DaprBuiltInActorNotFoundRetries in LLMOrchestator app triggered by pubsub


KaiWalter:
I have a sporadically have a problem getting a LLM orchestration running over pubsub.

My one deterministic workflow pusblishes the event:

```
== APP - worker-voice2action == 2025-09-05 14:31:32 INFO [voice2action] Published to pubsub/IntentOrchestrator for workflow_instance_id=test20250905T1431250000.mp3
```

and in that moment the receiving app with the **LLMOrchestrator** starts logging **DaprBuiltInActorNotFoundRetries** 

```
time="2025-09-05T14:31:32.988686817Z" level=warning msg="Error processing operation DaprBuiltInActorNotFoundRetries. Retrying in 1s…" app_id=orchestrator-intent instance=kw-devvm-vm scope=dapr.runtime type=log ver=1.16.0-rc.5
time="2025-09-05T14:31:37.457109353Z" level=info msg="[pubsub/IntentOrchestrator] Trace: 356f5877d3a4cfc8a9a217ac884d04bb Span: 7155649216c31c12/1dcf0255f235203c Time: [2025-09-05 14:31:32.984485984 +0000 UTC m=+60.609647831 ->  2025-09-05 14:31:32.984890097 +0000 UTC m=+60.610051944] Annotations: []" instance=kw-devvm-vm scope=dapr.runtime.trace type=log ver=1.16.0-rc.5
time="2025-09-05T14:31:38.99270072Z" level=warning msg="Error processing operation DaprBuiltInActorNotFoundRetries. Retrying in 1s…" app_id=orchestrator-intent instance=kw-devvm-vm scope=dapr.runtime type=log ver=1.16.0-rc.5
time="2025-09-05T14:31:44.995001882Z" level=warning msg="Error processing operation DaprBuiltInActorNotFoundRetries. Retrying in 1s…" app_id=orchestrator-intent instance=kw-devvm-vm scope=dapr.runtime type=log ver=1.16.0-rc.5
time="2025-09-05T14:31:51.001280147Z" level=warning msg="Error processing operation DaprBuiltInActorNotFoundRetries. Retrying in 1s…" app_id=orchestrator-intent instance=kw-devvm-vm scope=dapr.runtime type=log ver=1.16.0-rc.5
...
```

As pubsub and all state stores are Redis, I tried to clean out before starting the workload with `redis-cli --scan | xargs redis-cli DEL` - but that did not change the behavior.

[...]

PR #201, opened 2025-09-16
fix: ensure wf runtime started on orchestrators too

sicoyle:
I went ahead and separated out the fix into a diff PR since I know my other one is already pretty big. This fixes:
https://github.com/dapr/dapr/issues/9038#issuecomment-3292483892
https://github.com/dapr/dapr-agents/issues/198

### DIFF
--- a/dapr_agents/workflow/orchestrators/base.py
+++ b/dapr_agents/workflow/orchestrators/base.py
@@ -44,7 +44,6 @@ class OrchestratorWorkflowBase(AgenticWorkflow, ABC):
         self.register_agentic_system()
 
         # Start the runtime if it's not already running
-        logger.info("Starting workflow runtime...")
         self.start_runtime()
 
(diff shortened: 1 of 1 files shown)


dapr_agents/workflow/orchestrators/base.py
```diff
@@ -43,6 +43,10 @@ def model_post_init(self, __context: Any) -> None:
         # Register agent metadata
         self.register_agentic_system()
 
+        # Start the runtime if it's not already running
+        logger.info("Starting workflow runtime...")
```
filintod: there is a log in start_runtime() so probably we can remove this one to avoid repeating it:

https://github.com/dapr/dapr-agents/blob/main/dapr_agents/workflow/base.py#L442
```suggestion
```

yaron2 (approved):

Issue #167, opened 2025-07-29
Refactor DurableAgent Workflows Implementation for Simplicity and Resilience


sicoyle:
### Context

The current implementation of `DurableAgent` creates a workflow with excessive granularity—defining an activity for nearly every operation, even those that do not involve external calls. This results in unnecessary overhead and complexity. To improve efficiency and resilience, the workflow should be refactored to only treat truly atomic, out-of-process operations (e.g., LLM calls, MCP calls, tool executions) as individual activities.

### Acceptance Criteria

* Refactor `DurableAgent` workflows to only create activities for out-of-process operations (LLM, MCP, tool calls).
* Simplify internal logic to reduce the number of workflow activities.
* Document updated workflow structure and reasoning.
* Review changes with the team and get alignment on the updated design.

# Example

The  diagram bellow 

<img src="https://uploads.linear.app/5ff56eb9-d5ff-4cd7-a2c9-fc0dbcb5cc2d/6ea25fa0-96d8-4850-833c-f5dce973216f/eeae3913-d4eb-4492-99a5-e56fe700cc55?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzVmZjU2ZWI5LWQ1ZmYtNGNkNy1hMmM5LWZjMGRiY2I1Y2MyZC82ZWEyNWZhMC05NmQ4LTQ4NTAtODMzYy1mNWRjZTk3MzIxNmYvZWVhZTM5MTMtZDRlYi00NDkyLTk5YTUtZTU2ZmU3MDBjYzU1IiwiaWF0IjoxNzU2Mzg3MDQ1LCJleHAiOjMzMzI2OTQ3MDQ1fQ.S1_6xKpYSwJ613ctkDqyrz-8Evf2EBNjYN0BlZlGk74 " alt="Screenshot 2025-08-27 at 18.43.02.png" width="1144" data-linear-height="699" />

should render as:

```

                Start
                  |
              LLM Call
                  |
+--Tool1--Tool2--Tool3--Tool4--Tool5--+
                  |
              LLM Call
                  |
                 End
```

sicoyle:
Sam to make
Resiliency applied to workflows
and orchestrators workflow as separate issues

PR #204, opened 2025-09-19
feat: simplify durable agent wf def

sicoyle:
issue
https://github.com/dapr/dapr-agents/issues/167

I updated the durable agent wf definition to:
- ensure stateful actions were durable by enabling state management aspect to be atomic with the operation. For ex: if a tool is executed, then the results are immediately saved within the same activity.
- make wf runtime task/wf registration atomic so things will only register once (this fixes a race condition where task/wf registrations occurred before runtime was fully up)
- use deterministic time formatting from the wf context
- updated log levels to improve dx and not log so much and also not log on replays of the wf
- create/use helpers in the activities so they're not as large with making state management more atomic

The workflow used to look like this:
image

And now its simplified to this :)
image

### DIFF
--- a/dapr_agents/observability/wrappers/workflow_task.py
+++ b/dapr_agents/observability/wrappers/workflow_task.py
@@ -677,13 +677,11 @@ class WorkflowTaskWrapper:
         """
         if task_name in ["record_initial_entry", "get_workflow_entry_info"]:
             return "initialization"
-        elif task_name in ["append_assistant_message", "append_tool_message"]:
-            return "state_management"
         elif task_name in ["finalize_workflow", "finish_workflow"]:
             return "finalization"
         elif task_name in ["broadcast_message_to_agents", "send_response_back"]:
             return "communication"
-        elif task_name == "generate_response":
+        elif task_name == "generate_llm_response":
             return "llm_generation"
         elif task_name == "run_tool":
             return "tool_execution"
--- a/dapr_agents/observability/wrappers/workflow_task.py
+++ b/dapr_agents/observability/wrappers/workflow_task.py
@@ -194,7 +194,7 @@ class WorkflowTaskWrapper:
             attributes["resource.workflow.instance_id"] = instance_id
             attributes[
                 "resource.workflow.name"
-            ] = "ToolCallingWorkflow"  # Could be dynamic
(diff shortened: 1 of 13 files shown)


sicoyle:
I'm looking for input on activity naming still. For the generate_llm_response calls, I think that could be clearer with a few options:
1. `plan_and_call_tools` and the second call being `llm_final_response`
2. `llm_plan` and the second call being `llm_generate_response`
3. More ReAct focused: `reason_and_act` and the second call being `observe_and_respond`

Open to ideas here and on other activity call names please.

pyproject.toml
```diff
@@ -35,6 +35,7 @@ dependencies = [
     "cloudevents>=1.11.0,<2.0.0",
     "numpy>=2.2.2,<3.0.0",
     "mcp>=1.7.1,<2.0.0",
+    "websockets>=15.0.0,<16.0.0",
```
sicoyle, line 38: this was added bc MCP's websocket client requires it so in the tests i saw MCP tool schema creation fail silently... so went ahead and added a core dep

dapr_agents/observability/wrappers/workflow_task.py
Cyb3rWard0g, line 1: does the task_name `generate_llm_response` exist yet? or this is a placeholder for when the task names are updated

Cyb3rWard0g:
Other than the question about the name of the task, LGTM. If the task named "generate_llm_response" does not exist yet, then it will not trigger right?

dapr_agents/observability/wrappers/workflow_task.py
sicoyle: yes, i renamed generate_response to generate_llm_response so now thats what it appears as in tracing
image

bibryam:
1. IMO ToolCallingWorkflow -> AgenticWorkflow
as it is not always tool calling, and in fact it is an agentic workflow as per definition: 
 
**An agentic workflow is an AI-driven process where autonomous, intelligent AI agents manage complex tasks and processes with minimal human intervention, making decisions, coordinating actions, and adapting to changing circumstances in real-time.**

2.  We cannot use plan, final_call, etc as we not every LLM interaction starts with a plan, or ends up in tool call. 
I think we need two names:
run_tool -> run_tool (this is ok)
generate_llm_response -> call_llm

PR #213, opened 2025-09-23
feat: fix llm orchestrator + tracing + sessions for long-term memory

sicoyle:
This PR adds tracing instrumentation to the 05-multi-agent-workflows quickstart so I can investigate orchestrator behavior. While implementing workflow improvements, I discovered that removing global chat history in a previous PR changed orchestrator semantics and broke some orchestrator flows.

I removed global chat history to let in-flight workflows resume after app restarts and to avoid incorrectly interleaving messages (which produced invalid chat history and LLM provider errors). I ran the quickstarts after that change, but for orchestrators I only checked for workflow completion logs and killed the run — that was not sufficient and masked context-related failures that we see now.

Previously, orchestrators used a two-step approach:

1. Broadcast a message to all agents (updating a global conversation history).

2. Send a TriggerAction to the specific agent that should respond.

With global chat history gone, we need to decide how orchestrators should handle context going forward from an event-driven perspective:

This is what I went with: a two step approach:

- First event: orchestrator broadcasts context to all agents (agents store/update their local view of the workflow/conversation). This does not trigger the entire workflow, so agents do not respond at this point.

- Second event: orchestrator sends a TriggerAction to the target agent to act.

This PR so far:

- Improves parts of the orchestrator workflow definition.

- Updates prompts where the orchestrator was skipping steps or producing unexpected behavior.

- Adds tracing setup/fixes for orchestrators (still WIP).

- Fixes orchestrators to behave as expected.

- creates an internal trigger action and supports the external triggerAction, so when we have orchestrators trigger agents they do not also trigger themselves, and users still can trigger agents via pubsub as expected

- added context session_id to orchestrators so if you start them and let them run and the plan is not complete and you rerun them with the same session then it picks up where it left off prior
image
See here where it starts with substep 1.1
image

THen I let it run and it completes 1.1, 1.2, 1.3 so step 2 is next. Now, when I restart my orchestrator I see it pick up where it left off
image

Note: This only fixes LLM Orchestrator. I need to look at the random and roundrobin ones still.

### DIFF
--- a/dapr_agents/workflow/orchestrators/llm/orchestrator.py
+++ b/dapr_agents/workflow/orchestrators/llm/orchestrator.py
@@ -130,9 +130,9 @@ class LLMOrchestrator(OrchestratorWorkflowBase):
     def _does_workflow_exist(self, instance_id: str) -> bool:
         """
         Check if a workflow instance exists and is accessible via the Dapr client.
-        
+
         This function attempts to retrieve the workflow metadata from Dapr. A successful
-        response indicates the workflow exists in Dapr's state store, while failures 
+        response indicates the workflow exists in Dapr's state store, while failures
         (e.g., not found errors) indicate the workflow is no longer accessible.
 
         Args:


quickstarts/05-multi-agent-workflows/requirements.txt
```diff
@@ -1,3 +1,7 @@
-dapr-agents
+# dapr-agents
+# dapr-agents[observability]
```
filintod: rollback section

Cyb3rWard0g (approved):
LGTM Thank you @sicoyle !

dapr_agents/agents/durableagent/agent.py
```diff
@@ -843,14 +886,52 @@ def _construct_messages_with_instance_history(
...
+        # This ensures agents have access to broadcast messages and persistent context
+        long_term_memory_data = self.state.get("chat_history", [])
+
+        # Convert long-term memory to dict format for LLM consumption
+        long_term_memory_messages = []
+        for msg in long_term_memory_data:
+            if isinstance(msg, dict):
+                long_term_memory_messages.append(msg)
+            elif hasattr(msg, "model_dump"):
+                long_term_memory_messages.append(msg.model_dump())
```
filintod, line 913: is it possible to be any other type? or maybe we should log it if we get not `dict` and not `pydantic`, again not sure if possible

sicoyle: given our current setup i think this is it. State layer maintains dictionary so most times it is that, but everything else is a pydantic model

dapr_agents/agents/durableagent/agent.py
```diff
@@ -843,14 +886,52 @@ def _construct_messages_with_instance_history(
...
+
+        # For broadcast-triggered workflows, also include additional context memory
+        source = instance_data.get("source") if instance_data else None
```
filintod, line 916: let's add a todo/issue to design getting context for agents/sub-agents, context engineering is important to keep on task, too much context and it can derail the agent

dapr_agents/llm/dapr/chat.py
```diff
@@ -306,6 +303,11 @@ def generate(
             structured_mode=structured_mode,
         )
 
+        logger.info(f"Processed parameters for Dapr: {params}")
```
filintod: do we need all these logs at `info` level, seem very verbose

dapr_agents/llm/dapr/chat.py
```diff
@@ -347,6 +349,8 @@ def generate(
                 raw, llm_component or self._llm_component
             )
             logger.info("Chat completion retrieved successfully.")
+            logger.info(f"Dapr Conversation API response: {raw}")
```
filintod: same here, not sure we want `info` level

dapr_agents/llm/dapr/chat.py
```diff
@@ -369,9 +373,7 @@ def _check_dapr_runtime_support(metadata: "GetMetadataResponse"):  # noqa: F821
     dapr_runtime_version = extended_metadata.get("daprRuntimeVersion", None)
     if dapr_runtime_version is not None:
         # Allow only versions >=1.16.0, edge, and <2.0.0 for Alpha2 Chat Client
-        if not is_version_supported(
-            str(dapr_runtime_version), ">=1.16.0, edge, <2.0.0"
-        ):
+        if not is_version_supported(str(dapr_runtime_version), ">=1.16.0, <2.0.0"):
             raise DaprRuntimeVersionNotSupportedError(
                 f"!!!!! Dapr Runtime Version {dapr_runtime_version} is not supported with Alpha2 Dapr Chat Client. Only Dapr runtime versions >=1.16.0, edge, and <2.0.0 are supported."
```
filintod: I guess we need to also remove `edge` from error message

dapr_agents/llm/utils/structure.py
```diff
@@ -575,18 +575,22 @@ def validate_against_signature(result: Any, expected_type: Any) -> Any:
...
-            try:
-                if isinstance(result, list):
-                    return [
-                        StructureHandler.validate_response(item, model_cls).model_dump()
-                        for item in result
-                    ]
-                else:
+        if models:
+            # Try each model, but raise the last ValidationError if all fail
+            last_error = None
```
filintod: why not accumulate the errors? or maybe at least log the error for each

dapr_agents/observability/instrumentor.py
```diff
@@ -262,12 +263,28 @@ async def context_wrapped_coro():
...
                 except RuntimeError:
                     # No running loop - create new one
-                    loop = asyncio.new_event_loop()
-                    asyncio.set_event_loop(loop)
                     try:
+                        loop = asyncio.new_event_loop()
+                        asyncio.set_event_loop(loop)
                         return loop.run_until_complete(context_wrapped_coro())
+                    except Exception as e:
+                        logger.warning(
```
filintod, line 273: this section, and maybe the whole context_wrap needs to be analyze in the context of where is this executed in workflows and/or activities. it is strange for new_event_loop to fail, it might indicate other subtle bug,

sicoyle: This part of the code is managing OpenTelemetry context across thread boundaries. I'll add a TODO comment on this, but I think when we get to the point of using trace context propagation through Dapr all of this will be resolved

dapr_agents/workflow/orchestrators/llm/orchestrator.py
```diff
@@ -45,21 +50,117 @@ class LLMOrchestrator(OrchestratorWorkflowBase):
...
+                result.append(obj.model_dump())
+            elif isinstance(obj, dict):
+                # Already a dictionary
+                result.append(obj)
+            else:
+                # Fallback: try to convert to dict
+                result.append(dict(obj) if hasattr(obj, "__dict__") else obj)
+        return result
+
+    def _is_workflow_running(self, instance_id: str) -> bool:
```
filintod: probably change the name and docstring to match intention

dapr_agents/workflow/orchestrators/llm/orchestrator.py
```diff
@@ -136,48 +241,58 @@ def main_workflow(self, ctx: DaprWorkflowContext, message: TriggerAction):
             step_id = next_step.get("step", None)
             substep_id = next_step.get("substep", None)
 
-            # Step 5: Validate Step Before Proceeding
+            if not ctx.is_replaying:
```
filintod: this section seems strange, or maybe was during debug? as plan_dicts is not used anymore in this method I think

sicoyle: yes debug. The dict is just used in a print out - i can clean up

dapr_agents/workflow/orchestrators/llm/orchestrator.py
```diff
@@ -702,61 +631,433 @@ async def finish_workflow(
...
+                                plan_schema=schemas.plan,
+                            ),
+                        }
+                    ],
+                    response_format=IterablePlanStep,
+                    structured_mode="json",
+                )
+
+                # Parse the response - now we get a Pydantic model directly
+                if hasattr(response, "choices") and response.choices:
```
filintod, line 690: do we need this `choices`? isn't this deprecated by now?

dapr_agents/workflow/orchestrators/llm/orchestrator.py
```diff
@@ -702,61 +631,433 @@ async def finish_workflow(
...
+                            progress_check_schema=schemas.progress_check,
+                        ),
+                    }
+                ],
+                response_format=ProgressCheckOutput,
+                structured_mode="json",
+            )
+
+            # Parse the response - now we get a Pydantic model directly
+            if hasattr(progress_response, "choices") and progress_response.choices:
```
filintod, line 840: also here "if" not using raw data anymore

dapr_agents/workflow/orchestrators/llm/orchestrator.py
sicoyle: yes we need choices as it is part of the conversation api response matching openai tool calling format

dapr_agents/workflow/orchestrators/llm/orchestrator.py
sicoyle: the if depends on which llm provider / api version. Most do support choices field with openai tool calling format, but not all do so that's why i use the if

filintod (approved):
lgtm to make it work, we should get together and redesign at some point some aspects of this, and for sure we need to add comprehensive tests to the coordination mechanism

PR #223, opened 2025-10-01
Fix: durable agent context

sicoyle:
This is a branch off the branch in this PR https://github.com/dapr/dapr-agents/pull/213

This PR fixes the durable agent to use session IDs for maintaining context across workflows. See here where i had one browser window open for this chat
image
I then killed the app and restarted things and asked about context from prior
image

### DIFF
--- a/dapr_agents/agents/durableagent/agent.py
+++ b/dapr_agents/agents/durableagent/agent.py
@@ -20,6 +20,7 @@ from dapr_agents.types import (
 from dapr_agents.types.workflow import DaprWorkflowStatus
 from dapr_agents.workflow.agentic import AgenticWorkflow
 from dapr_agents.workflow.decorators import message_router, task, workflow
+from dapr_agents.memory import ConversationDaprStateMemory
 
 from .schemas import (
     AgentTaskResponse,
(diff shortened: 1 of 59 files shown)


Cyb3rWard0g (approved):
LGTM 🔥

dapr_agents/agents/durableagent/agent.py
```diff
@@ -859,7 +874,11 @@ async def process_broadcast_message(self, message: BroadcastMessage):
 
                 # Start the agent's workflow
                 await self.run_and_monitor_workflow_async(
+<<<<<<< HEAD
```
filintod: bad merge

dapr_agents/agents/durableagent/agent.py
```diff
@@ -900,48 +919,50 @@ def _construct_messages_with_instance_history(
...
-        # Add long-term memory first (broadcast messages, persistent context)
-        chat_history.extend(long_term_memory_messages)
+        # First add persistent memory and long-term memory as user messages for context
+        # This ensures we have cross-workflow context but doesn't interfere with tool state order
+        for msg in persistent_memory_messages + long_term_memory_messages:
+            msg_dict = msg.model_dump() if hasattr(msg, "model_dump") else dict(msg)
+            if msg_dict in chat_history:
+                continue
+            # Convert tool-related messages to user messages to avoid conversation order issues
+            if msg_dict.get("role") in ["tool", "assistant"] and (
```
filintod, line 953: let's create an issue/todo to properly design session and storing session, as we should not need this if the session is correctly store, we should be able to just retrieve and use without sorting it out

PR #233, opened 2025-10-14
Enable Native Dapr Workflows with Message Router Decorators

Cyb3rWard0g:
# This PR Depends on https://github.com/dapr/dapr-agents/pull/232

## Overview

This PR introduces a new `@message_router` decorator that enables native Dapr workflow orchestration triggered directly from Pub/Sub messages without relying on legacy Dapr Agents abstractions. Developers now have full control over the Dapr Workflow runtime and Pub/Sub client, allowing workflows and LLM/Agent activities to be registered, composed, and executed using the official Dapr APIs.

This update modernizes the message routing layer to align with Dapr’s workflow primitives, improving transparency, flexibility, and maintainability.

## Key Changes

* Added `decorators/routers.py`
  * Implements a new `@message_router` decorator for schema-aware message routing
  * Fully compatible with native `WorkflowRuntime` and DaprClient lifecycle
  * Supports Union types, forward references, and robust model validation
  * Designed for explicit developer control rather than hidden orchestration
* Added supporting utilities
  * `utils/routers.py`: message extraction, validation, and CloudEvent parsing helpers
  * `utils/registration.py`: `register_message_handlers()` for runtime subscription via the active DaprClient
* Deprecated legacy message router
  * Added deprecation warnings to `decorators/messaging.py`
  * Directs users to migrate to the new Dapr-native decorator
* Added new `Message Router Workflow` Quickstart
  * Demonstrates event-driven orchestration using Pub/Sub → Workflow integration
  * Includes LLM-powered create_outline and write_post workflow activities
  * Adds `message_client.py` publisher and updated README with full configuration
* Updated quickstart index
  * Added `Message Router Workflow` alongside LLM- and Agent-based patterns

> Check the new quickstart `quickstarts/04-agent-based-workflows/README.md`

### DIFF
--- a/tests/workflow/test_message_router.py
+++ b/tests/workflow/test_message_router.py
@@ -1,0 +1,663 @@
+import pytest
+from typing import Union, Optional
+from dataclasses import dataclass
+from unittest.mock import MagicMock
+from pydantic import BaseModel, Field
+
+from dapr_agents.workflow.decorators.routers import message_router
+from dapr_agents.workflow.utils.routers import (
+    extract_message_models,
+    extract_cloudevent_data,
+    validate_message_model,
+    parse_cloudevent,
+)
+from dapr_agents.workflow.utils.registration import register_message_handlers
+
+
+# Test Models
+class OrderCreated(BaseModel):
+    """Test Pydantic model for order creation events."""
+
+    order_id: str = Field(..., description="Unique order identifier")
+    amount: float = Field(..., description="Order amount")
+    customer: str = Field(..., description="Customer name")
+
+
+class OrderCancelled(BaseModel):
+    """Test Pydantic model for order cancellation events."""
+
+    order_id: str = Field(..., description="Order ID to cancel")
+    reason: str = Field(..., description="Cancellation reason")
+
+
+@dataclass
+class ShipmentCreated:
+    """Test dataclass for shipment events."""
+
+    shipment_id: str
+    order_id: str
+    carrier: str
+
+
+# Tests for extract_message_models utility
+
+
+def test_extract_message_models_single_class():
+    """Test extracting a single model class."""
+    models = extract_message_models(OrderCreated)
+    assert models == [OrderCreated]
+
+
+def test_extract_message_models_union():
+    """Test extracting models from Union type hint."""
+    models = extract_message_models(Union[OrderCreated, OrderCancelled])
+    assert set(models) == {OrderCreated, OrderCancelled}
+
+
+def test_extract_message_models_optional():
+    """Test extracting models from Optional type hint (filters out None)."""
+    models = extract_message_models(Optional[OrderCreated])
+    assert models == [OrderCreated]
+
+
+def test_extract_message_models_pipe_union():
+    """Test extracting models from pipe union syntax (Python 3.10+)."""
+    # Note: This test requires Python 3.10+ for the | syntax
+    try:
+        hint = eval("OrderCreated | OrderCancelled")
+        models = extract_message_models(hint)
+        assert set(models) == {OrderCreated, OrderCancelled}
+    except SyntaxError:
+        pytest.skip("Python 3.10+ required for pipe union syntax")
+
+
+def test_extract_message_models_none_input():
+    """Test extracting models from None returns empty list."""
+    models = extract_message_models(None)
+    assert models == []
+
+
+def test_extract_message_models_non_class():
+    """Test extracting models from non-class type returns empty list."""
+    models = extract_message_models("not a class")
+    assert models == []
+
+
+# Tests for message_router decorator
+
+
+def test_message_router_requires_pubsub():
+    """Test that message_router raises ValueError when pubsub is missing."""
+    with pytest.raises(
+        ValueError,
+        match="`pubsub` and `topic` are required when using @message_router with arguments",
+    ):
+
+        @message_router(topic="orders")
+        def handler(message: OrderCreated):
+            pass
+
+
+def test_message_router_requires_topic():
+    """Test that message_router raises ValueError when topic is missing."""
+    with pytest.raises(
+        ValueError,
+        match="`pubsub` and `topic` are required when using @message_router with arguments",
+    ):
+
+        @message_router(pubsub="messagepubsub")
+        def handler(message: OrderCreated):
+            pass
+
+
+def test_message_router_requires_message_parameter():
+    """Test that message_router raises ValueError when 'message' parameter is missing."""
+    with pytest.raises(ValueError, match="must have a 'message' parameter"):
+
+        @message_router(pubsub="messagepubsub", topic="orders")
+        def handler(data: OrderCreated):  # Wrong parameter name
+            pass
+
+
+def test_message_router_requires_type_hint():
+    """Test that message_router raises TypeError when message parameter has no type hint."""
+    with pytest.raises(TypeError, match="must type-hint the 'message' parameter"):
+
+        @message_router(pubsub="messagepubsub", topic="orders")
+        def handler(message):  # No type hint
+            pass
+
+
+def test_message_router_unsupported_type():
+    """Test that message_router raises TypeError for unsupported message types."""
+    with pytest.raises(TypeError, match="Unsupported model type"):
+
+        @message_router(pubsub="messagepubsub", topic="orders")
+        def handler(message: str):  # str is not a supported model
+            pass
+
+
+def test_message_router_basic_decoration():
+    """Test basic message_router decoration with single model."""
+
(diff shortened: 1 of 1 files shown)


dapr_agents/workflow/decorators/messaging.py
```diff
@@ -1,11 +1,21 @@
...
 
 logger = logging.getLogger(__name__)
 
+_MESSAGE_ROUTER_DEPRECATION_MESSAGE = (
+    "@message_router (legacy version from dapr_agents.workflow.decorators.messaging) "
+    "is deprecated and will be removed in a future release. "
+    "Please migrate to the updated decorator in "
+    "`dapr_agents.workflow.decorators.routers`, which supports "
+    "Union types, forward references, and explicit Dapr workflow integration."
+)
```
sicoyle, line 17: i vote we just remove since we are not v1.0 and beyond, we can make breaking changes and remove the message_router as long as we let folks know in the next release announcement :)

quickstarts/04-message-router-workflow/README.md
```diff
@@ -0,0 +1,539 @@
...
+```text
+04-message-router-workflow/
+├─ components/                 # Dapr components (pubsub, conversation, workflow state)
+├─ app.py                      # Starts WorkflowRuntime + subscribes message routers
+├─ handlers.py                 # @message_router handler (validates + schedules workflow)
+├─ workflow.py                 # workflow & activities (decorated with @llm_activity)
+└─ message_client.py           # publishes a test message to the topic
+```
+
+## Files
```
sicoyle, line 122: i think it might be nice moving forward for us to not put the code in the readmes and just reference/link the files/components. Thoughts?

dapr_agents/workflow/decorators/routers.py
```diff
@@ -0,0 +1,113 @@
...
+
+        @message_router(pubsub="pubsub", topic="events")
+        def on_event(message: Union[Foo, Bar]): ...
+
+    Args:
+        func: (optional) bare-decorator form support.
+        pubsub: Name of the Dapr pub/sub component (required when used with args).
+        topic: Topic name to subscribe to (required when used with args).
+        dead_letter_topic: Optional dead-letter topic (defaults to f"{topic}_DEAD").
+        broadcast: Optional flag you can use downstream for fan-out semantics.
```
sicoyle, line 44: some of these say required in the docs string but the code shows optional. Are func, pubsub, topic the ones that we require?

Cyb3rWard0g:
> 1. IMO, In addition to declarative, programatic, and streaming subscriptions, the Message Router is another (4th) type of subscribing to a topic. It is probably the most pythonic version among all.
> 2. Is there anything specific to DaprAgents or can message router be even at python sdk?
> 3. I love this feature, just not sure where it fits best (DA or Python SDK) and what to call it? There is already the concept of subscription, and @app.route in python sdk

2) I believe it can be part of the python SDK. I just like to simplify how we could register those subscriptions, and since we do not have a class in the Python SDK that makes it easy to define `streaming` subscriptions (https://docs.dapr.io/developing-applications/building-blocks/pubsub/subscription-methods/#streaming-subscriptions) without relying on HTTP endpoints, I created those utils.
3) How long would it take to add something like that to the Python SDK. Once again, streaming pubsub subscriptions without HTTP endpoints. `app.route` relies on an HTTP endpoint right?

bibryam:
> 2. I believe it can be part of the python SDK. I just like to simplify how we could register those subscriptions, and since we do not have a class in the Python SDK that makes it easy to define `streaming` subscriptions (https://docs.dapr.io/developing-applications/building-blocks/pubsub/subscription-methods/#streaming-subscriptions) without relying on HTTP endpoints, I created those utils.


Isn't this the way to create a streaming consumer?
```
        subscription = client.subscribe(
            pubsub_name='pubsub', topic='orders', dead_letter_topic='orders_dead'
        )
```

> 3. How long would it take to add something like that to the Python SDK. Once again, streaming pubsub subscriptions without HTTP endpoints. `app.route` relies on an HTTP endpoint right?

There is no extra time requirement, but we have to make sure:
 - it is documented in th SDK docs how it works
 - when to use this instead of other ways
 - has example (that is in the sdk repo)
 - aligned with existing naming convention
Once these are in place (btw we would need these even if it is in the dapr agents repo), the review and merge could be pretty quick.

Cyb3rWard0g:
> > 2. I believe it can be part of the python SDK. I just like to simplify how we could register those subscriptions, and since we do not have a class in the Python SDK that makes it easy to define `streaming` subscriptions (https://docs.dapr.io/developing-applications/building-blocks/pubsub/subscription-methods/#streaming-subscriptions) without relying on HTTP endpoints, I created those utils.
> 
> Isn't this the way to create a streaming consumer?
> 
> ```
>         subscription = client.subscribe(
>             pubsub_name='pubsub', topic='orders', dead_letter_topic='orders_dead'
>         )
> ```
> 
> > 3. How long would it take to add something like that to the Python SDK. Once again, streaming pubsub subscriptions without HTTP endpoints. `app.route` relies on an HTTP endpoint right?
> 
> There is no extra time requirement, but we have to make sure:
> 
> * it is documented in th SDK docs how it works
> * when to use this instead of other ways
> * has example (that is in the sdk repo)
> * aligned with existing naming convention
>   Once these are in place (btw we would need these even if it is in the dapr agents repo), the review and merge could be pretty quick.

that is not enough

```python
subscription = client.subscribe(
    pubsub_name='pubsub', topic='orders', dead_letter_topic='orders_dead'
)
```

We need to map it to a handler which can be done with 

```python
close_fn = client.subscribe_with_handler(
      pubsub_name='pubsub', topic='orders', handler_fn=process_message,
      dead_letter_topic='orders_dead'
  )
```

But once again, that is not enough, the "handler_fun" in our project would point to a Dapr Workflow script which would need to be run using a dapr workflow client. Therefore, we need to create that abstraction so that we can use it on our DurableAgent or Orchestrator class which rely on a `message_router` decorator. For users not using DurableAgent or Orchestrator classes but writing their own Dapr Workflows, they can still use message utils that I have added in this PR to make it easy to subscribe a handler that is a dapr workflow function.

bibryam:
Thanks for the explanation @Cyb3rWard0g, I see why is this needed.
Considering message_router is a higher level abstraction than what is available in python-sdk, adding it to dapr-agents repo is also fine.
I'm OK either way, I'll let @yaron2 and @sicoyle to weigh in.

quickstarts/04-message-router-workflow/README.md
sicoyle: ignore this comment pls - ill make this change in a pr applying it to all quickstarts

dapr_agents/workflow/decorators/routers.py
Cyb3rWard0g: Ah yes they are optional. Sorry forgot to update the docstring. However, in the big PR, it is documented properly. I also added `message_model` in the latest PR to validate the schema of incoming messages.

```python
def message_router(
    func: Optional[Callable[..., Any]] = None,
    *,
    pubsub: Optional[str] = None,
    topic: Optional[str] = None,
    dead_letter_topic: Optional[str] = None,
    broadcast: bool = False,
    message_model: Optional[Any] = None,
) -> Callable[[Callable[..., Any]], Callable[..., Any]]:
    """
    Tag a callable as a **Pub/Sub → Workflow** entry with routing + schema metadata.

    Args:
        func (Optional[Callable[..., Any]]):
            The function to decorate (if used without parentheses).
        pubsub (Optional[str]):
            The name of the Dapr pub/sub component. Optional when wiring via `PubSubRouteSpec`.
        topic (Optional[str]):
            The pub/sub topic to subscribe to. Optional when wiring via `PubSubRouteSpec`.
        dead_letter_topic (Optional[str]):
            The dead-letter topic to publish failed messages to.
        broadcast (bool):
            Whether to treat this as a broadcast subscription.
        message_model (Optional[Any]):
            The message model class or Union[...] to use for validation.

    Returns:
        Callable[[Callable[..., Any]], Callable[..., Any]]:
            The decorated function.
    """
    ```

dapr_agents/workflow/decorators/messaging.py
Cyb3rWard0g: I removed this in the big PR ;) Thank you for the suggestion.

sicoyle:
> Thanks for the explanation @Cyb3rWard0g, I see why is this needed.
Considering message_router is a higher level abstraction than what is available in python-sdk, adding it to dapr-agents repo is also fine.
I'm OK either way, I'll let @yaron2 and @sicoyle to weigh in.

Yes, dapr agents is the place for this for now

Issue #234, opened 2025-10-15
Allow to configure grpc.max_send_message_length and grpc.max_receive_message_length for underlying durable tasks


fabistb:
# General
Dapr Agents / or the python-sdk in general doesn't provide a clear way to overwrite the default 4Mb `grpc.max_send_message_length` and `grpc.max_receive_message_length` for workflows. This can cause `RESOURCE_EXHAUSTED` exceptions. Especially for AI powered workfows this isn't ideal since the message length can't always be kept small, especially when images are taken into consideration.

# Error example:
```
2025-10-13 09:06:47.452 durabletask-worker WARNING: Application-level gRPC error (StatusCode.RESOURCE_EXHAUSTED): <_MultiThreadedRendezvous of RPC that terminated with:
  status = StatusCode.RESOURCE_EXHAUSTED
  details = "CLIENT: Received message larger than max (9108880 vs. 4194304)"
  debug_error_string = "UNKNOWN:Error received from peer  {grpc_message:"CLIENT: Received message larger than max (9108880 vs. 4194304)", grpc_status:8}"
``` 

# Description
It appears that there are ways to work around this by updating the `durabletask.internal.shared.get_grpc_channel` before the workflow is instantiated. However this isn't really a nice way of doing things.

# Related GitHub issues
- https://github.com/dapr/python-sdk/issues/371
- https://github.com/dapr/python-sdk/issues/451

PR #238, opened 2025-10-27
feat: enable grpc config on agent instantiation

sicoyle:
fix for https://github.com/dapr/dapr-agents/issues/234 + tests

to bubble up the grpc config fields to set GRPC max send and receive message lengths.

ex:
```
agent = DurableAgent(
        ...
        grpc_max_send_message_length=32 * 1024 * 1024,    # 32MB
        grpc_max_receive_message_length=32 * 1024 * 1024,  # 32MB
    )
```

or you can set on workflow app itself:
```
 app = WorkflowApp(
       ...
        grpc_max_send_message_length=4 * 1024 * 1024,     # 4MB (default)
        grpc_max_receive_message_length=64 * 1024 * 1024,  # 64MB (large responses)
    )
```

### DIFF
--- a/tests/workflow/test_grpc_config.py
+++ b/tests/workflow/test_grpc_config.py
@@ -1,6 +1,7 @@
 """Tests for gRPC configuration in WorkflowApp."""
 import pytest
 from unittest.mock import MagicMock, patch, call
+import types
 from dapr_agents.workflow.base import WorkflowApp
 
 
@@ -125,10 +126,18 @@ def test_grpc_channel_patching():
 
     mock_shared.get_grpc_channel = original_get_grpc_channel
 
+    # Create dummy package/module structure so 'from durabletask.internal import shared' works
+    durabletask_module = types.ModuleType("durabletask")
+    internal_module = types.ModuleType("durabletask.internal")
+    setattr(durabletask_module, "internal", internal_module)
+    setattr(internal_module, "shared", mock_shared)
+
     with patch.dict(
         "sys.modules",
         {
             "grpc": mock_grpc,
+            "durabletask": durabletask_module,
+            "durabletask.internal": internal_module,
             "durabletask.internal.shared": mock_shared,
         },
     ), patch("dapr_agents.workflow.base.WorkflowRuntime"), patch(


dapr_agents/workflow/base.py
```diff
@@ -92,6 +103,59 @@ def model_post_init(self, __context: Any) -> None:
...
+                    )
+                )
+                logger.debug(
+                    f"Configured gRPC max_receive_message_length: {self.grpc_max_receive_message_length} bytes ({self.grpc_max_receive_message_length / (1024 * 1024):.2f} MB)"
+                )
+
+            # Patch the function to include our custom options
+            def get_grpc_channel_with_options(address: str):
+                """Custom gRPC channel factory with configured message size limits."""
+                return grpc.insecure_channel(address, options=options)
```
filintod: we don't provide an option here for a secured channel that will be required when the sidecar is not local

filintod: also the signature on durable task is:

```
def get_grpc_channel(
        host_address: Optional[str],
        secure_channel: bool = False,
        interceptors: Optional[Sequence[ClientInterceptor]] = None) -> grpc.Channel:
```

we are changing the signature (dropping two kwargs), maybe it does not matter but maybe it would fail if executed and those two arguments are passed https://github.com/dapr/durabletask-python/blob/9bba47962a86772f00ff544b026c345dc901ad24/durabletask/client.py#L114.  have we tested with the workflow running? not mocked

dapr_agents/workflow/base.py
```diff
@@ -92,6 +103,59 @@ def model_post_init(self, __context: Any) -> None:
...
+        This is particularly useful for AI-powered workflows that may need to handle large payloads
+        such as images, which can exceed the default 4MB gRPC message size limit.
+        """
+        try:
+            import grpc
+            from durabletask.internal import shared
+
+            # Create custom options list
+            options = []
+            if self.grpc_max_send_message_length:
```
filintod, line 137: let's validate input to avoid devs shooting themselves in the foot.

dapr_agents/workflow/base.py
sicoyle: updated and yes i checked :)
```
...
 initialized with session ID: my-unique-id
== APP == INFO:dapr_agents.workflow.agentic:State store 'workflowstatestore' initialized.
== APP == WARNING:dapr_agents.workflow.mixins.state:Skipping state save: Empty state.
== APP == INFO:dapr_agents.workflow.base:Configured gRPC max_send_message_length: 33554432 bytes (32.00 MB)
== APP == INFO:dapr_agents.workflow.base:Configured gRPC max_receive_message_length: 33554432 bytes (32.00 MB)
== APP == INFO:dapr_agents.workflow.base:WorkflowApp initialized; discovering tasks and workflows.
== APP == INFO:dapr_agents.workflow.base:Starting workflow runtime.
== APP == 2025-10-29 14:45:53.725 durabletask-worker INFO: Starting gRPC worker that connects to dns:127.0.0.1:51517
== APP == INFO:dapr_agents.workflow.base:Sleeping for 5 seconds to ensure runtime is started.
== APP == 2025-10-29 14:45:53.749 durabletask-worker INFO: Created fresh connection to dns:127.0.0.1:51517
```

filintod:
lgtm, this should actually be in python-sdk/durabletask-python and we have some already pending there. Once that is done we can refactor/simplify this to use that

Cyb3rWard0g:
LGTM. We just need to import that back when we merge the big PR. It should be easy to move those scripts to the right place.

Cyb3rWard0g (approved):

Issue #65, opened 2025-03-28
Externalize Agent Configuration


bibryam:
**Context**
The current setup for creating an agent includes both core agent logic and infrastructure configuration in one place:

```
travel_planner = AssistantAgent(
    name="TravelBuddy",
    role="Travel Planner",
    goal="Help users find flights and remember preferences",
    instructions=[
        "Find flights to destinations",
        "Remember user preferences",
        "Provide clear flight info"
    ],
    tools=[search_flights],
    message_bus_name="messagepubsub",
    state_store_name="workflowstatestore",
    state_key="workflow_state",
    agents_registry_store_name="workflowstatestore",
    agents_registry_key="agents_registry",
    service_port=8001,
    daprGrpcPort=50001
)

travel_planner.as_service(port=8001)
```

**Expected behaviour**

This mixes business logic (agent’s purpose, tools, instructions) with low-level plumbing details (state store keys, ports, Dapr config).
Moving these to a config file with sensible defaults would improve separation of concerns, readability, and reusability.
I should be able to specify most of the following as env variables or have good defaults based on naming convention

```
    message_bus_name="messagepubsub",
    state_store_name="workflowstatestore",
    state_key="workflow_state",
    agents_registry_store_name="workflowstatestore",
    agents_registry_key="agents_registry",
    service_port=8001,
    daprGrpcPort=50001
```

yaron2:
Agree, I think to simplify configuration further we can assume a single state store is provided and merge state store and agent registry store

bibryam:
@msfussell and I discussed this issue, and we might need to investigate a bit more on this, whether language specific environment files or a dapr-specific conig file is the preferred direction forward.

sicoyle:
# Current Thinking

Short-term (v1): Leverage subclasses more extensively to separate fields into logical subgroups, improving structure and maintainability.

Long-term (v2, optional): Introduce support for a YAML-based configuration file for greater flexibility and alignment with ecosystem practices.

# References

1. CrewAI: YAML definitions with relatively flat field structures for agents

- Docs: [CrewAI Agents](https://docs.crewai.com/en/concepts/agents)
- Example: [Base Agent implementation](https://github.com/crewAIInc/crewAI/blob/main/src/crewai/agents/agent_builder/base_agent.py)

2. OpenAI: Provides a Python SDK, but the documentation lacks a comprehensive overview of all optional fields, making it difficult to assess full configuration support

[...]

PR #241, opened 2025-10-30
feat: create new prompt class

sicoyle:
issue: https://github.com/dapr/dapr-agents/issues/65

This PR continues to make agent initialization simpler by creating a new separate prompt class. I am not really sure that we want to expose this class as most folks probably don't use it, but open to thoughts here as this is for the prompt system to handle how:
1. agent prompts (system messages and full chat context) are built, formatted, and sent to the LLM. Right now it takes name, role, goal, instructions (agent identiy/behavior) and turns these into a templated prompt string or chat sequence.
2. prompt template logic to support default and user-provided templates
3. chat history integration packaging chat history into the message list or formatted prompt in a suitable way for the LLM

This is likely not used by 99% of folks, but maybe more useful to:
- advanced agent users
- platform admins who want to override the entire system prompt logic to inject a prompt or prompt template

This PR mostly moves the prompt stuff out of the agent logic into it's own class, renames things, and has minor tweaks along the way.

Behavior change: before, the agent's prompt template was pushed/synced with the llm prompt template. However, now, this PR makes the agent own its prompt and does not mutate the llm internal prompt template. If the LLM has this set, then the agent ignores it and still sets its own prmopt template. In other words, the llm client shouldn't own the prompt anymore as the agent.prompt builds the system prompt, injects context/variables, includes chat history, and properly formats messages. At call time, the agent passes these formatted messages to the llm, and the llm just executes it. However, I do have to leave prompt template field on the LLM as if someone uses just our llm clients, they might want that functionality, but with an agent, then we just use the agent prompt template.

### DIFF
--- a/tests/agents/durableagent/test_durable_agent.py
+++ b/tests/agents/durableagent/test_durable_agent.py
@@ -719,8 +719,8 @@ class TestDurableAgent:
         }
 
         # Mock prompt template
-        basic_durable_agent.prompt_template = Mock()
-        basic_durable_agent.prompt_template.format.return_value = [
+        basic_durable_agent.prompt.template = Mock()
+        basic_durable_agent.prompt.template.format.return_value = [
             {"role": "system", "content": "System prompt"}
         ]
 


dapr_agents/agents/base.py
```diff
@@ -101,28 +85,7 @@ class AgentBase(BaseModel, ABC):
...
-## Role
-Your role is {role}.
-
-## Goal
-{goal}.
-
-## Instructions
-{instructions}.
-""".strip()
+    prompt: Optional[Prompt] = Field(default_factory=Prompt, description="TODO SAM")
```
filintod: remember to change the TODO

dapr_agents/agents/base.py
```diff
@@ -214,84 +182,22 @@ def model_post_init(self, __context: Any) -> None:
...
-        self.prompt_template = self._initialize_prompt_template()
-        # Ensure LLM client and agent both reference the same template
-        if self.llm is not None:
-            self.llm.prompt_template = self.prompt_template
-
-        self._validate_prompt_template()
-        self.prefill_agent_attributes()
+        if self.prompt is None:
+            self.prompt = Prompt()
+        if getattr(self.prompt, "context", None) is None:
```
filintod, line 190: when you create a prompt you also create a  context in post_init but you don't fill out the information so this context information (name, role, goal and instruction) should be applied always, no?

sicoyle: updated to yes always use the agent info from end users

dapr_agents/prompt/agent_prompt.py
```diff
@@ -0,0 +1,225 @@
...
+        default=None,
+        description="A custom system prompt, overriding name, role, goal, and instructions.",
+    )
+    template: Optional[PromptTemplateBase] = Field(
+        default=None, description="The prompt template for the agent."
+    )
+    # TODO: we should have a system_template, prompt_template, and response_template, or better separation here.
+    # If we have something like a customer service agent, we want diff templates for different types of interactions.
+    # In future, we could also have a way to dynamically change the template based on the context of the interaction.
+    template_format: Literal["f-string", "jinja2"] = Field(
```
filintod, line 29: you can create constants for these, or enum class, that way we reduce typos and hardcoded strings.

Constants + literal type example:

```
# Template Format constants
TPL_FORMAT_F_STRING = "f-string"
TPL_FORMAT_JINJA = "jinja2"
# TPL_FORMAT_T_STRING = "t-string"  # for future template string

# Type alias for tool types
TemplateFormatType = Literal["f-string", "jinja2"]
```

in your class:

```
template_format: TemplateFormatType = Field(default=TPL_FORMAT_JINJA, description=...)
```

dapr_agents/prompt/agent_prompt.py
```diff
@@ -0,0 +1,225 @@
...
+        Build the system prompt for the agent using a single template string.
+        - Fills in the current date.
+        - Leaves placeholders for name, role, goal, and instructions as variables (instructions only if set).
+        - Converts placeholders to Jinja2 syntax if requested.
+
+        Returns:
+            str: The formatted system prompt string.
+        """
+        # Default f-string template; placeholders will be swapped to Jinja if needed.
+        default_sys_prompt = """
```
filintod, line 149: probably these template strings can be on their own file? or as variables without indentation somewhere, there are extra prefix spaces

filintod (commented):
lgtm, just some nits

dapr_agents/prompt/agent_prompt.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def _collect_template_attrs(self) -> tuple[Dict[str, str], List[str]]:
+        """
+        Collect agent attributes for prompt template pre-filling and warn about unused ones.
+        - valid: attributes set on self and declared in prompt_template.input_variables.
+        - unused: attributes set on self but not present in the template.
+        Returns:
+            (valid, unused): Tuple of dict of valid attrs and list of unused attr names.
+        """
+        attrs = ["name", "role", "goal", "instructions"]
```
filintod, line 117: I don't get where this is coming from or if it should be more prominent, like template attributes somewhere.

dapr_agents/prompt/agent_prompt.py
```diff
@@ -0,0 +1,225 @@
...
+                else:
+                    valid[attr] = str(val)
+            else:
+                unused.append(attr)
+        return valid, unused
+
+    def construct_system_prompt(self) -> str:
+        """
+        Build the system prompt for the agent using a single template string.
+        - Fills in the current date.
```
filintod, line 141: was wondering what is the usage of date? seems strange to need it

Cyb3rWard0g: The date provides essential temporal context that allows the LLM to understand and reason about time-relative concepts like 'tomorrow,' calculate durations, handle scheduling tasks, and maintain consistency when discussing events or deadlines throughout the conversation.

filintod: understood, but temporal-related actions are a specific use case(s) and we embed this all the time even if not used, where that could be in instructions that is dynamic and people can customize it. We could add an option like add date, but again users can do it.

sicoyle:
closing as this is outdated and low priority but can circle back here eventually :)

PR #249, opened 2025-11-02
Refactor: Auto-inject Agent State Schemas and simplify agent configuration API

Cyb3rWard0g:
## **Overview:**
This PR eliminates unsafe user-facing schema customization by auto-injecting the correct state model bundle based on agent/orchestrator type. Users now provide only storage configuration, and the framework automatically selects the appropriate schemas (`AgentWorkflowState` for agents, `LLMWorkflowState` for orchestrators), preventing runtime `AttributeError` crashes from incompatible custom schemas.

## **Key Changes:**
- **Introduced `StateModelBundle` abstraction** to encapsulate state schemas and hook functions with validation guards
- **Auto-injection pattern** where agent/orchestrator classes inject their required schema bundle during initialization without user intervention
- **Simplified `AgentStateConfig` API** by removing exposed `state_model_cls` and `message_model_cls` fields that allowed unsafe customization
- **Three-tier fallback resolution** for state hooks: config overrides → default bundle → hardcoded defaults
- **Added type hints** for `default_bundle` parameter using `TYPE_CHECKING` to prevent circular imports
- **Fixed incorrect import paths** in LLM orchestrator compensation logic (`dapr_agents.workflow.orchestrators` → `dapr_agents.agents.orchestrators`)
- **Updated all quickstarts** to use simplified configuration syntax (removed `pubsub_config=`, `state_config=` prefixes)
- **Added test coverage** for bundle injection and conflict detection

### DIFF
--- a/dapr_agents/agents/components.py
+++ b/dapr_agents/agents/components.py
@@ -101,6 +101,11 @@ class AgentComponents:
             )
             bundle = DEFAULT_AGENT_WORKFLOW_BUNDLE
 
+        # I considered splitting into separate classes, but that would duplicate several lines
+        # of infrastructure code (pub/sub, state operations, registry mutations). The current design
+        # uses the Strategy Pattern to share infrastructure while maintaining type-safe schemas per
+        # agent/orchestrator type. The "complexity" is just 5 lines of bundle extraction vs maintaining
+        # duplicate codebases.
         self._state_model_cls = bundle.state_model_cls
         self._message_model_cls = bundle.message_model_cls
         self._entry_factory = bundle.entry_factory


dapr_agents/agents/components.py
```diff
@@ -13,15 +13,15 @@
...
-    AgentWorkflowEntry,
-    AgentWorkflowMessage,
-    AgentWorkflowState,
-)
+from dapr_agents.agents.schemas import AgentWorkflowEntry
 from dapr_agents.storage.daprstores.stateservice import StateStoreError
 from dapr_agents.types.workflow import DaprWorkflowStatus
 
+if TYPE_CHECKING:
+    from dapr_agents.agents.configs import StateModelBundle
```
sicoyle: can we just import the pkg instead of the if check here? is there a benefit to the if?

dapr_agents/agents/components.py
```diff
@@ -41,70 +41,75 @@ def __init__(
...
-            Callable[[BaseModel], Optional[dict]]
-        ] = (
-            getattr(state_config, "entry_container_getter", None)
-            if state_config
-            else None
-        )
+        bundle = None
+        if state is not None:
+            if default_bundle is not None:
+                state.ensure_bundle(default_bundle)
```
sicoyle, line 89: can you add a comment somewhere in here that this is necessary to support using the common state class for both orchestrators and agents to then override their state model setup pls?

dapr_agents/agents/components.py
```diff
@@ -41,70 +41,75 @@ def __init__(
...
+                "No state bundle for %s; using default AgentWorkflowState schema",
+                self.name,
+            )
+            bundle = DEFAULT_AGENT_WORKFLOW_BUNDLE
+
+        self._state_model_cls = bundle.state_model_cls
+        self._message_model_cls = bundle.message_model_cls
+        self._entry_factory = bundle.entry_factory
+        self._message_coercer = bundle.message_coercer
+        self._entry_container_getter = bundle.entry_container_getter
```
sicoyle, line 113: i wonder if it would be easier to have two diff classes for agent state vs orchestrator state that can then use the same base class instead of these factory/coercers. what are your thoughts on that as an option? 🤔

dapr_agents/agents/durable.py
```diff
@@ -410,12 +411,16 @@ def call_llm(
         if user_copy is not None:
             self.text_formatter.print_message({str(k): v for k, v in user_copy.items()})
 
+        tools = self.get_llm_tools()
+        generate_kwargs = {
+            "messages": messages,
+            "tools": tools,
+        }
+        if tools and self.execution.tool_choice is not None:
+            generate_kwargs["tool_choice"] = self.execution.tool_choice
```
sicoyle, line 420: does this fix that openai issue you created recently?

dapr_agents/agents/orchestrators/base.py
```diff
@@ -3,18 +3,22 @@
...
 from dapr_agents.agents.configs import (
+    AgentExecutionConfig,
     AgentPubSubConfig,
     AgentRegistryConfig,
     AgentStateConfig,
 )
 from dapr_agents.agents.utils.text_printer import ColorTextFormatter
 
+if TYPE_CHECKING:
+    from dapr_agents.agents.configs import StateModelBundle
```
sicoyle: if we need this import 99% of the time should we just go ahead and import it instead of wrapping the conditional around it?

sicoyle (commented):
thank you for your efforts on this! I love the idea of having the state implementation unified for agents and orchestrators  and make things easier on us to maintain 🙌 🌟 

few comments so far, but yeah I'm curious for your thoughts on. What are the benefits of the factory/coercer setup vs just having a static state class for agents and another for orchestrators that are slightly diff but use the same underlying storage base class? Would that be easier than this approach?

dapr_agents/agents/components.py
Cyb3rWard0g: done ✅

dapr_agents/agents/durable.py
Cyb3rWard0g: correct.

dapr_agents/agents/orchestrators/base.py
Cyb3rWard0g: I removed type checking. thank you

dapr_agents/agents/components.py
Cyb3rWard0g: I considered splitting into separate classes, but that would duplicate ~700 lines of infrastructure code (pub/sub, state operations, registry mutations). The current design uses the Strategy Pattern to share infrastructure while maintaining type-safe schemas per agent/orchestrator type. The "complexity" is just 5 lines of bundle extraction vs maintaining duplicate codebases. Happy to add inline documentation if that would help clarify the design!

sicoyle: thank you for explaining, that makes sense. if possible it would be great to have comments in code on this pls 🙏

Cyb3rWard0g: done ✅

PR #250, opened 2025-11-02
Support configurable workflow gRPC payload limits for agents and orchestrators

Cyb3rWard0g:
Reintroduced @sicoyle idea from here https://github.com/dapr/dapr-agents/pull/238. I added a shared gRPC runtime options model and helper that raises Durable Task channel limits when needed, wired it through durable agents and orchestrators, and added unit tests covering the patching behavior.

## Key Changes

- Added `WorkflowGrpcOptions` and a central `apply_grpc_options` helper for Durable Task channel overrides.
- Threaded gRPC options through `AgentComponents`, `DurableAgent`, and all `workflow orchestrators`.
- Added a quickstart to pass execution configs and demo larger gRPC payload limits.
- Added targeted tests verifying the gRPC helper’s patching logic.


dapr_agents/agents/base.py
```diff
@@ -53,82 +55,86 @@ def __init__(
...
-        memory_config: Optional[AgentMemoryConfig] = None,
+        memory: Optional[AgentMemoryConfig] = None,
         llm: Optional[ChatClientBase] = None,
         tools: Optional[Iterable[Any]] = None,
         # Metadata
         agent_metadata: Optional[Dict[str, Any]] = None,
+        workflow_grpc: Optional[WorkflowGrpcOptions] = None,
         # Execution
-        execution_config: Optional[AgentExecutionConfig] = None,
+        execution: Optional[AgentExecutionConfig] = None,
```
sicoyle, line 71: thoughts on nesting this under AgentExecutionConfig class?

dapr_agents/workflow/utils/grpc.py
```diff
@@ -0,0 +1,83 @@
...
+    Patch Durable Task's gRPC channel factory with custom message size limits.
+
+    Durable Task (and therefore Dapr Workflows) creates its gRPC channels via
+    ``durabletask.internal.shared.get_grpc_channel``.  This helper monkey patches
+    that factory so that subsequent runtime/client instances honour the provided
+    ``grpc.max_send_message_length`` / ``grpc.max_receive_message_length`` values.
+    """
+    if not options:
+        return
+    if not options.max_send_message_length and not options.max_receive_message_length:
```
sicoyle: can you update so folks can update one and not the other field and the options still apply pls?

dapr_agents/workflow/utils/grpc.py
```diff
@@ -0,0 +1,83 @@
...
+
+import logging
+from typing import Optional, Sequence
+
+from dapr_agents.agents.configs import WorkflowGrpcOptions
+
+logger = logging.getLogger(__name__)
+
+
+def apply_grpc_options(options: Optional[WorkflowGrpcOptions]) -> None:
```
sicoyle, line 14: pls add a comment somewhere similar to how i had for an easy ref in durabletask where this is:
```
# This is a copy of the original get_grpc_channel function in durabletask.internal.shared at
                # https://github.com/dapr/durabletask-python/blob/7070cb07d07978d079f8c099743ee4a66ae70e05/durabletask/internal/shared.py#L30C1-L61C19
                # but with my option overrides applied above.
```

quickstarts/12-test-durable-agent-grpc/README.md
```diff
@@ -0,0 +1,41 @@
+# DurableAgent with custom gRPC limits
```
sicoyle: can we add these optional fields to one of the existing quickstarts and ensure the user facing docs highlight what these fields are to be used for? we can also add this to the upstream docs.

sicoyle (commented):
thank you for bringing these back 🙌

dapr_agents/agents/base.py
Cyb3rWard0g: I was not sure if the concept of "Workflow Options" would go under a config that is trying to set how many iterations the agent can take to solve a goal or tool options. I guess it makes sense if we say that the `WorkflowGrpcOptions` add context to how tasks / activities are executed in the workflow? If so, yes, it would not be a problem.

Cyb3rWard0g: Actually, while integrating both and reviewing all the scripts again, I changed my mind. Keeping them separate is better because `AgentExecutionConfig` controls agent behavior (iterations, tool_choice), while `WorkflowGrpcOptions` controls infrastructure (network transport limits). These are semantically different layers - execution logic vs transport configuration. Also, gRPC limits are used by DurableAgent AND OrchestratorBase  (which doesn't have execution config), so nesting it under `AgentExecutionConfig` would create an awkward dependency for orchestrators that only need gRPC settings. Alsom standalone agents use the execution config and giving the user the alternative to set grpc settings, when the standalone agent does not rely on it might be uncessary. Keeping it separate we can keep workflow grpc settings to only classes that would use it.

dapr_agents/workflow/utils/grpc.py
Cyb3rWard0g: Good catch! The original logic already supported setting one or both options, but the double-negative check (if not X and not Y) was confusing. I've updated it to explicitly check is None for clarity and added a note in the docstring that users can set either or both options. Now it's clear that you can configure just `max_send_message_length`, just `max_receive_message_length`, or both independently.

dapr_agents/workflow/utils/grpc.py
Cyb3rWard0g: done ✅

quickstarts/12-test-durable-agent-grpc/README.md
Cyb3rWard0g: Yes I fixed that in the latest PR where I added Agent Handoffs

dapr_agents/agents/base.py
sicoyle: very good point and i agree now so thank you

PR #251, opened 2025-11-04
Enable Agent Handoffs via PubSub

Cyb3rWard0g:
This PR introduces `HandoffSpec` for declarative agent-to-agent delegation, enabling agents to hand off tasks to specialized agents in both triage and cyclic workflow patterns. It also adds support for structured output responses and improves workflow runner reliability.

## Key Changes

- **HandoffSpec**: New declarative API for defining agent-to-agent handoffs with automatic tool generation
- **Structured Output Support**: Added `output_type` parameter to agents for enforcing Pydantic model responses
- **Optional PubSub Config**: Made `AgentPubSubConfig` optional to support agents triggered only through `AgentRunner`
- **Runner Cleanup**: Fixed shutdown sequence before workflow runtime shutdown
- **Signal Handler Management**: Improved signal handler lifecycle to prevent duplicate registrations
- **Quickstart Examples**: Added comprehensive examples for translator triage pattern and research-writer cyclic workflow

### DIFF
--- a/tests/agents/agent/test_agent.py
+++ b/tests/agents/agent/test_agent.py
@@ -279,10 +279,13 @@ class TestAgent:
         # Call the actual internal method
         result = await agent_with_tools._run_tool_call("test-instance", tool_call)
 
-        # Verify the result is a tool message dict
-        assert result["role"] == "tool"
-        assert result["name"] == echo_tool.name
-        assert result["content"] == "value1"
+        # Verify the result structure (now returns dict with 'tool_result' and 'message')
+        assert "message" in result
+        assert "tool_result" in result
+        message = result["message"]
+        assert message["role"] == "tool"
+        assert message["name"] == echo_tool.name
+        assert message["content"] == "value1"
 
     async def test_run_tool_failure(self, agent_with_tools):
         """Test tool execution failure via _run_tool_call method."""
--- a/tests/agents/durableagent/test_durable_agent.py
+++ b/tests/agents/durableagent/test_durable_agent.py
@@ -445,8 +445,7 @@ class TestDurableAgent:
         assert entry.output == final_output
         assert entry.end_time is not None
 
-    @pytest.mark.asyncio
-    async def test_run_tool(self, basic_durable_agent, mock_tool):
+    def test_run_tool(self, basic_durable_agent, mock_tool):
         """Test that run_tool atomically executes and persists tool results."""
         from datetime import datetime, timezone
 
@@ -489,7 +488,7 @@ class TestDurableAgent:
             mock_ctx = Mock()
 
             with patch.object(basic_durable_agent, "save_state"):
-                result = await basic_durable_agent.run_tool(
+                result = basic_durable_agent.run_tool(
                     mock_ctx,
                     {
                         "tool_call": tool_call,
@@ -500,20 +499,9 @@ class TestDurableAgent:
                 )
 
             # Verify tool was executed and result was returned
-            assert result["tool_call_id"] == "call_123"
-            assert result["tool_name"] == "test_tool"
-            assert result["execution_result"] == "tool_result"
-
-            # Verify state was updated atomically
-            entry = basic_durable_agent._state_model.instances[instance_id]
-            assert len(entry.messages) == 1  # Tool message added
-            assert len(entry.tool_history) == 1  # Tool execution record added
-
-            # Verify tool execution record in tool_history
-            tool_history_entry = entry.tool_history[0]
-            assert tool_history_entry.tool_call_id == "call_123"
-            assert tool_history_entry.tool_name == "test_tool"
-            assert tool_history_entry.execution_result == "tool_result"
+            assert result["tool_result"]["tool_call_id"] == "call_123"
+            assert result["tool_result"]["tool_name"] == "test_tool"
+            assert result["tool_result"]["execution_result"] == "tool_result"
 
     def test_record_initial_entry(self, basic_durable_agent):
         """Test record_initial_entry helper method."""
@@ -707,8 +695,7 @@ class TestDurableAgent:
         result = basic_durable_agent._state_model.instances.get("non-existent")
         assert result is None
 
-    @pytest.mark.asyncio
-    async def test_create_tool_message_objects(self, basic_durable_agent):
+    def test_create_tool_message_objects(self, basic_durable_agent):
         """Test that tool message objects are created correctly (via run_tool activity)."""
         from datetime import datetime, timezone
 
@@ -744,7 +731,7 @@ class TestDurableAgent:
             mock_ctx = Mock()
 
             with patch.object(basic_durable_agent, "save_state"):
-                result = await basic_durable_agent.run_tool(
+                result = basic_durable_agent.run_tool(
                     mock_ctx,
                     {
                         "tool_call": tool_call,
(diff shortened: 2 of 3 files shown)


dapr_agents/agents/durable.py
```diff
@@ -2,17 +2,25 @@
...
     AgentPubSubConfig,
     AgentRegistryConfig,
     AgentStateConfig,
-    AgentExecutionConfig,
+    WorkflowGrpcOptions,
+)
+from dapr_agents.agents.handoffs import (
+    HandoffSpec,
+    create_handoff_tool,
+    generate_handoff_aliases,
```
sicoyle, line 23: note to self to look at what this does

dapr_agents/agents/durable.py
```diff
@@ -24,13 +32,16 @@
 from dapr_agents.prompt.base import PromptTemplateBase
 from dapr_agents.types import (
     AgentError,
+    AgentToolExecutorError,
```
sicoyle, line 35: not a requirement here but would be nice if it was just ToolError

sicoyle:
still looking through the code, but does this work for orchestrator handoffs?

dapr_agents/agents/durable.py
```diff
@@ -59,71 +71,162 @@ def __init__(
...
+                        "handoffs iterable must contain HandoffSpec instances."
+                    )
+                normalized_specs.append(spec)
+
+        if normalized_specs:
+
+            def metadata_resolver() -> Dict[str, Any]:
+                return self.get_agents_metadata(
+                    exclude_self=True,
+                    exclude_orchestrator=True,
```
sicoyle, line 172: do we always want this set to true?

dapr_agents/agents/durable.py
```diff
@@ -59,71 +71,162 @@ def __init__(
...
+        self,
+        spec: HandoffSpec,
+        metadata_resolver: Callable[[], Mapping[str, Any]],
+    ) -> None:
+        """
+        Materialize a handoff tool, register it with the executor, and track aliases.
+        """
+        tool = create_handoff_tool(
+            target_agent_name=spec.agent_name,
+            source_agent_name=self.name,
```
sicoyle, line 192: still reviewing, but maybe we can renameif the target/source is an orchestrator (not technically an agent)? do we set agent_name on orchestrators?

dapr_agents/agents/durable.py
```diff
@@ -59,71 +71,162 @@ def __init__(
...
+            self.tool_executor.register_tool(tool)
+        except AgentToolExecutorError:
+            logger.warning("Duplicate handoff tool skipped: %s", tool.name)
+            return
+
+        self.tools.append(tool)
+
+        alias_keys = {
+            alias.lower()
+            for alias in generate_handoff_aliases(tool.name, spec.agent_name)
```
sicoyle, line 209: why do we need a handoff alias? if we register an agent/tool name, then don't we have all the info we need to not need an alias? am i missing something here?

dapr_agents/agents/durable.py
```diff
@@ -236,35 +346,122 @@ def agent_workflow(self, ctx: wf.DaprWorkflowContext, message: dict):
...
+                        yield ctx.call_activity(
+                            self.persist_tool_results,
+                            input={
+                                "instance_id": ctx.instance_id,
+                                "tool_results": persisted_results,
+                                "time": ctx.current_utc_datetime.isoformat(),
+                            },
+                        )
+
+                    if handoff_results:
```
sicoyle, line 372: how is handoff different than supporting agents as tools? or is this basically just that?

dapr_agents/agents/durable.py
```diff
@@ -236,35 +346,122 @@ def agent_workflow(self, ctx: wf.DaprWorkflowContext, message: dict):
...
+                            self.name,
+                            turn,
+                            ctx.instance_id,
+                        )
+                    break
+
                 if not ctx.is_replaying:
                     logger.debug(
-                        "Agent %s produced final response on turn %d (instance=%s)",
+                        "Agent %s response did not satisfy output schema on turn %d; retrying.",
```
sicoyle, line 442: is this log expected?

dapr_agents/agents/durable.py
```diff
@@ -236,35 +346,122 @@ def agent_workflow(self, ctx: wf.DaprWorkflowContext, message: dict):
...
+        if self.broadcast_topic_name and not is_handoff_exit:
             yield ctx.call_activity(
                 self.broadcast_message_to_agents,
                 input={"message": final_message},
             )
 
         # Optionally send a direct response back to the trigger origin.
-        if source and trigger_instance_id:
+        # Skip if expect_response=False (e.g., handoff chains where source doesn't need response).
+        if source and trigger_instance_id and expect_response:
```
sicoyle, line 464: how do we from a DX configure expect_response, ie ack here?

dapr_agents/agents/durable.py
```diff
@@ -407,40 +614,100 @@ def call_llm(
...
+        if self.output_type:
+            generate_kwargs["response_format"] = self.output_type
+
         try:
-            response: LLMChatResponse = self.llm.generate(
-                messages=messages,
-                tools=self.get_llm_tools(),
-                tool_choice=self.execution_config.tool_choice,
-            )
+            llm_result = self.llm.generate(**generate_kwargs)
```
sicoyle, line 634: is there a way we can avoid updating the func from defined/concrete params to accepting kwargs? That can be difficult to know what exactly to pass in without looking concretely at generate() implementation. I know its likely driven by tool_choice param and not wanting to set it if None, but can this be done within generate() itself instead of here pls?

dapr_agents/agents/durable.py
```diff
@@ -407,40 +614,100 @@ def call_llm(
...
+            return False
+
+        if not self.output_type or self.output_type is str:
+            return True
+
+        try:
+            parsed = json.loads(content) if isinstance(content, str) else content
+        except json.JSONDecodeError:
+            logger.debug(
+                "Agent %s structured output is not valid JSON yet; continuing loop.",
```
sicoyle, line 684: does this also acct for if max_iterations is reached?

dapr_agents/agents/handoffs.py
```diff
@@ -0,0 +1,176 @@
...
+from dapr_agents.agents.schemas import TriggerAction
+from dapr_agents.tool.base import AgentTool
+from dapr_agents.types import AgentError
+from dapr_agents.workflow.utils.pubsub import send_message_to_agent
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(slots=True)
+class HandoffSpec:
```
sicoyle, line 18: what does the dataclass(slots=True) do? Also, is there any reference you used in creating the fields for this class? like other ai framework or something along those lines?

quickstarts/README.md
```diff
@@ -146,6 +146,19 @@ This quickstart demonstrates a Lord of the Rings themed multi-agent system where
...
+Spin up a small team of durable agents that hand off work to one another without an
+orchestrator. All agents share the same conversation memory stored in Dapr state
+and publish handoff requests via pub/sub.
+
+- **Durable Agents**: Each agent keeps workflow state and tool history
+- **Shared Memory**: Agents read/write a common `ConversationDaprStateMemory`
+- **Handoff Tools**: Use tool-calling to delegate work between agents
+- **Event-driven Collaboration**: Handoffs publish `TriggerAction` events via Dapr pub/sub
+
+[Go to Durable Swarm with Handoffs](./13-swarm-handoffs)
```
sicoyle, line 160: how is this different than defining an agent as a tool? I'm still trying to digest this PR, so sorry if I am missing that, but yeah trying to delineate handoff vs agent used as a tool to another agent/orchestrator?

dapr_agents/agents/durable.py
Cyb3rWard0g: Yes for `exclude_self` and not for `exclude_orchestrator`. I will have another PR later for "Agents as Tools" which is work on the top of the agent handoff idea where we add a way to say "respond to agent that handed off a task to you". We will then be able to say it is fine to handoff a task back to another agent / orchestrator .

dapr_agents/agents/durable.py
Cyb3rWard0g: Interesting, I guess if the user sets the `HandOffSpe` name to the orchestrator name, we should be able to task the orchestrator. However, today, orchestrators have their own logic that might not follow your typical Agent <-> Agent communication. An orchestrator might want to run their "orchestrator" logic rather than taking the input and reason over what to do next based on the shared chat history.

sicoyle:
i am holding off on reviewing until all quickstarts are g2g in main pls 🙏

Cyb3rWard0g:
Now that I was able to apply all the fixes needed to run all quickstarts: https://github.com/dapr/dapr-agents/pull/267, I will close this one and open a new one to only add "Agent handoffs" features in a fresh PR aligned with latest `main`.

Issue #129, opened 2025-06-11
Add Acceptance Test for Dapr Agents


bibryam:
**Context**
To ensure Dapr Agents work end-to-end and remain reliable over time, we need a reliable acceptance testing strategy. This involves deciding whether to formalize the Quick Start as a mechanical markdown based test or automate the cookbooks The goal is to establish a reliable, repeatable acceptance test that is maintainable and runs automatically in CI

**Acceptance Criteria**

- Decide on the acceptance test source
- Implement the chosen approach as an acceptance test
- Integrate test into CI to run automatically on every commit

PR #253, opened 2025-11-04
test: automate testing our quickstarts

sicoyle:
This PR introduces automation to validate our quickstarts as a new build step. Because these quickstarts consume API credits from NVIDIA, OpenAI, and ElevenLabs, they should only run when a PR is ready for merge consideration—not on every commit. We should never merge PRs that break our quickstarts, and this change enforces that by integrating them as part of an official integration test suite.

I’m also including a commit to remove the previous Mechanical Markdown setup, which was incomplete and not easily extensible to support additional services like Docker containers or dapr init, nor simple to integrate into CI in a Python-native test flow. Tests should be a core part of our developer workflow, and since pytest is the standard for Python-based testing, it’s the natural choice for verifying our quickstarts, and will align as we continue adding more integration tests habitually. This approach removes the Mechanical Markdown dependency, simplifies our tooling, and creates a more scalable foundation for testing beyond just return codes and response strings.

For now, this PR focuses on establishing that initial setup. 🙂

Follow-up question for Roberto: Is it okay to use NVIDIA and ElevenLabs for our automated tests?

To use, you must put an "/ok-to-test" comment on PR, and only maintainers should have perms

fixes https://github.com/dapr/dapr-agents/issues/129

### DIFF
--- a/dapr_agents/agents/base.py
+++ b/dapr_agents/agents/base.py
@@ -12,6 +12,7 @@ from dapr_agents.agents.configs import (
     AgentRegistryConfig,
     AgentStateConfig,
     AgentExecutionConfig,
+    DEFAULT_AGENT_WORKFLOW_BUNDLE,
 )
 from dapr_agents.agents.prompting import AgentProfileConfig, PromptingAgentBase
 from dapr_agents.agents.utils.text_printer import ColorTextFormatter
@@ -44,7 +45,7 @@ class AgentBase(AgentComponents):
         self,
         *,
         # Profile / prompt
-        profile_config: Optional[AgentProfileConfig] = None,
+        profile: Optional[AgentProfileConfig] = None,
         name: Optional[str] = None,
         role: Optional[str] = None,
         goal: Optional[str] = None,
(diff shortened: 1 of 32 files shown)


.github/workflows/build.yaml
```diff
@@ -71,4 +71,4 @@ jobs:
         tox -e type
     - name: Run Tests
       run: |
-        tox -e pytest
\ No newline at end of file
+        tox -e pytest -- -m "not integration"
```
filintod: probably not needed if you already have it like that in tox.ini

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,221 @@
...
+      uses: astral-sh/setup-uv@v3
+      with:
+        version: "latest"
+    
+    - name: Install dependencies
+      run: |
+        uv venv
+        source .venv/bin/activate
+        uv pip install --upgrade pip setuptools wheel
+        uv pip install -e .
```
filintod, line 91: probably not needed? the next line "should" do it, but check first

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,221 @@
...
+        OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+        ELEVENLABS_API_KEY: ${{ secrets.ELEVENLABS_API_KEY }}
+        # Note: hugging face api not available in repo, so we can add later
+        # HUGGINGFACE_API_KEY: ${{ secrets.HUGGINGFACE_API_KEY }}
+        NVIDIA_API_KEY: ${{ secrets.NVIDIA_API_KEY }}
+      run: |
+        source .venv/bin/activate
+        pytest -m integration -v tests/integration/quickstarts/test_01_hello_world.py::TestHelloWorldQuickstart::test_basic_llm
+    
+    - name: Cleanup Dapr
```
filintod: do we need this clean up in the vm? same for cleanup docker

docs/development/README.md
```diff
@@ -97,6 +97,33 @@ tox -e pytest tests/test_random_orchestrator.py
...
+### Integration Tests
+
+> Note: we do not use `pytest-docker-compose` intentionally here because it is not compatible with Python2, 
+and requires an old version of pyyaml < version6, but the rest of our project requires >6 for this pkg.
+
+Requires Dapr CLI to be installed.
+
+```
+# Install test dependencies
+pip install -e .[test]
```
filintod, line 109: put quotes around .[test], that always gets me in my shell. `pip install -e ".[test]"`

filintod:
the one problem with this way of testing is that we are not testing the same that the user is if they are copy pasting from the README. like setting up the environment and all that, it's like we need to sync the README markdown to whatever we use for testing, where mechanical markdown was doing it, not the MM is the nicest, just pointing out what we would be missing

we need an automated way to sync all steps the user will be taking and our tests should have the same environment, etc.

sicoyle:
> the one problem with this way of testing is that we are not testing the same that the user is if they are copy pasting from the README. like setting up the environment and all that, it's like we need to sync the README markdown to whatever we use for testing, where mechanical markdown was doing it, not the MM is the nicest, just pointing out what we would be missing

> we need an automated way to sync all steps the user will be taking and our tests should have the same environment, etc.

I addressed this by spinning up a venv per quickstart and the tests install the requirements.txt file within each quickstart in a fresh venv, but i made it configurable so we can use our existing venvs locally if we want :) thanks for the great feedback!

PR #270, opened 2025-11-11
fix: correct state race condition

sicoyle:
still rerunning to see if i get diff results, but fix for
```... 
Invalid parameter: Duplicate value for 'tool_call_id' of 'call_XXX', 
...
```

Turns out there was a race condition due to tools all running and getting and then saving state all at the same time, so I was seeing a few diff openai issues. This fixes the race condition with making add_messages atomic and leverages memory as the source of truth upon app restart.

### DIFF
--- a/dapr_agents/agents/base.py
+++ b/dapr_agents/agents/base.py
@@ -587,15 +587,15 @@ class AgentBase(AgentComponents):
         container = self._get_entry_container()
         entry = container.get(instance_id) if container else None
         if entry is not None and hasattr(entry, "messages"):
-        # Use configured coercer / message model
-        message_model = (
-            self._message_coercer(user_message_copy)  # type: ignore[attr-defined]
-            if getattr(self, "_message_coercer", None)
-            else self._message_dict_to_message_model(user_message_copy)
-        )
-        entry.messages.append(message_model)  # type: ignore[attr-defined]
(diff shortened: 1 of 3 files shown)


dapr_agents/agents/durable.py
```diff
@@ -548,8 +548,27 @@ async def _execute_tool() -> Any:
...
                     entry.last_message = tool_message_model
 
-        # Always persist to memory + in-process tool history
-        self.memory.add_message(tool_message)
+        # Check if tool message already exists in persistent memory before adding
+        tool_call_id = agent_message["tool_call_id"]
+        existing_memory_messages = self.memory.get_messages()
+        tool_exists_in_memory = False
+        for mem_msg in existing_memory_messages:
+            msg_dict = (
```
Cyb3rWard0g, line 569: Is this necessary?

```python
            msg_dict = (
                mem_msg.model_dump()
                if hasattr(mem_msg, "model_dump")
                else (mem_msg if isinstance(mem_msg, dict) else {})
            )
   ```
   
   The get_message methods always return a list of dictionaries right?

sicoyle:
I tested an agent with memory configured, but no state config setup and it worked consistently (3 times), and i also checked an agent with memory and state configured and that was consistently fine as well (3 times)...

Issue #278, opened 2025-11-25
Workflow completion callback capabilities on Orchestrators


yaron2:
As a user, I want to be able to get the final output of the workflow started by the orchestrator when it runs to completion. Related to https://github.com/dapr/dapr-agents/issues/209

Issue #209, opened 2025-09-22
workflow completion callback capabilities on DurableAgent


sicoyle:
When running agent.run, if an in-flight workflow contains an activity that calls something like os.Exit(), the entire app is terminated once that activity executes.

After fixing the activity logic (e.g., removing the os.Exit() call), the workflow resumes and eventually completes successfully. At that point, I can see the logs (pretty-printed workflow conversation).

However, as an end user, I want a way to programmatically retrieve the workflow result once it completes, rather than just relying on logs.

We should introduce a mechanism such as an onWorkflowCompletion callback field on the agent. This would allow developers to define a handler that automatically receives the result of the .run call once the workflow finishes.

This would make it easier to:
- React to workflow completions.
- Capture and process results directly in application logic.
- Avoid relying solely on logs for completion details.

sicoyle:
Note to self: I'm going to think more on this too in terms of categorizing some of the fields bc this lends towards execution expectations upon completion, and I know I need to bubble up the workflow timeout from durabletask, and there's probably other fields related that relate to workflow execution specifically...

PR #279, opened 2025-11-25
Add callback method to orchestrator to get final output

yaron2:
Closes https://github.com/dapr/dapr-agents/issues/278


CasperGN:
@yaron2 does this also close #209?

yaron2:
> @yaron2 does this also close #209?

I believe so since a Durable Agent returns a result after our latest agent runner changes

PR #282, opened 2025-11-26
Workflow summary fix

yaron2:
This PR properly exposes the callback field to the user. Also changes max iterations from 8 to 1 for quickstarts and adds an example for how to use the summary callback function.


Cyb3rWard0g (approved):
LGTM

PR #283, opened 2025-11-27
Fix/277 runner lifecycling

### LINKED ISSUE #277: AgentRunner should manage full agent lifecycle and documentation needs updates
**Description**:
The current API requires calling `agent.start()` before using `runner.serve()` and calling `agent.stop()` after `runner.shutdown()`. It would be better to have a single component that manages both start and stop in a symmetric way. `runner.serve()` or `runner.run()` should start the agent if it is not already started, and shutting down the runner should shut down the entire agent without additional calls.

The documentation at [https://docs.dapr.io/developing-applications/dapr-agents/dapr-agents-core-concepts/](https://docs.dapr.io/developing-applications/dapr-agents/dapr-agents-core-concepts/) does not explain how the runner is intended to be used. The examples show agents being started directly, but they do not include the runner. The documentation should be updated to show the runner, describe how it is meant to be used, and clarify whether it can run a single agent or multiple agents.

**Relevant code snippet**:

```python
runner = AgentRunner()
try:
    runner.serve(travel_assistant, port=5001)
finally:
    runner.shutdown()
    travel_assistant.stop()
```

CasperGN:
# Description

This also carries a fix to the quickstarter tests.

Moving lifecycling of `DurableAgent` into `AgentRunner` for a more streamlined instantiation.
The `AgentRunner` will now handle running `agent.start()` in
- `run`
- `serve`
- `subscribe`
- `register_routes`

It will catch the `RuntimeError` thrown by the agent if already running and `pass`.

`AgentRunner().shutdown()` now overrides and takes the `agent: Any` parameter to call `agent.stop()` moving the lifecycle of the agent into the runner:

Before:
```python
agent = DurableAgent()
agent.start()
runner = AgentRunner()
try:
    runner.serve(agent, port=8001)
finally:
    runner.shutdown()
    agent.stop()
```

Now:
```python
agent = DurableAgent()
runner = AgentRunner()
try:
    runner.serve(agent, port=8001)
finally:
    runner.shutdown(agent)
```

## Issue reference

Please reference the issue this PR will close: Closes #277 

## Checklist

* [X] Created/updated tests
* [X] Tested this change against all the quickstarts
* [X] Extended the documentation
    * [X] Created the [dapr/docs](https://github.com/dapr/docs) PR: https://github.com/dapr/docs/pull/4955

No tests needed updates as the agents are never passed to an `AgentRunner`.
`tests/integration/quickstarts/test_03_durable_agent_tool_call.py` has a dependency on `arize-phoenix-otel` - @Cyb3rWard0g is this intentional or a leftover? [See here](https://github.com/dapr/dapr-agents/blob/main/quickstarts/03-durable-agent-tool-call/durable_weather_agent_tracing.py#L16)

### DIFF
--- a/dapr_agents/tool/executor.py
+++ b/dapr_agents/tool/executor.py
@@ -5,7 +5,7 @@ from rich.table import Table
 from rich.console import Console
 
 from dapr_agents.tool.base import AgentTool
-from dapr_agents.types import AgentToolExecutorError, ToolError
+from dapr_agents.types import AgentToolExecutorError, ToolError, ToolResult, TextContent
 
 logger = logging.getLogger(__name__)
 
@@ -132,7 +132,12 @@ class AgentToolExecutor(BaseModel):
             return tool(*args, **kwargs)
         except ToolError as e:
             logger.error(f"Tool execution error in '{tool_name}': {e}")
-            raise AgentToolExecutorError(str(e)) from e
+            return ToolResult(
+                content=[
+                    TextContent(text=f"Error executing tool '{tool_name}': {str(e)}")
+                ],
+                isError=True,
+            )
         except Exception as e:
             logger.error(f"Unexpected error in '{tool_name}': {e}")
             raise AgentToolExecutorError(
--- a/dapr_agents/types/__init__.py
+++ b/dapr_agents/types/__init__.py
@@ -29,6 +29,8 @@ from .tools import (
     OAIFunctionDefinition,
     OAIToolDefinition,
     ToolExecutionRecord,
+    TextContent,
+    ToolResult,
 )
 
 __all__ = [
@@ -63,4 +65,6 @@ __all__ = [
     "OAIFunctionDefinition",
     "OAIToolDefinition",
     "ToolExecutionRecord",
+    "TextContent",
+    "ToolResult",
 ]
--- a/dapr_agents/types/tools.py
+++ b/dapr_agents/types/tools.py
@@ -158,3 +158,37 @@ class ToolExecutionRecord(BaseModel):
         None,
         description="Result of the tool execution, if available.",
     )
+
+
+class TextContent(BaseModel):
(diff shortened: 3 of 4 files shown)


bibryam:
@CasperGN this simplifies runner usage significantly.

IMO shutdown could be simplified tiny bit more. 
Have you considered these 2 scenarios around shutdown:


```
# Create multiple agents
agent1 = DurableAgent(name="Agent1", ...)
agent2 = DurableAgent(name="Agent2", ...)
agent3 = DurableAgent(name="Agent3", ...)

# Create a runner and serve/subscribe multiple agents
runner = AgentRunner()

# Serve multiple agents
runner.serve(agent1, port=8001)
runner.subscribe(agent2)
runner.serve(agent3, port=8002)  # agent3 uses serve instead of subscribe

# Scenario 1: Shutdown a specific agent
# This should only stop agent2, leaving agent1 and agent3 running
runner.shutdown(agent2)

# Scenario 2: Shutdown all agents
# This should stop all remaining agents (agent1 and agent3)
runner.shutdown()

# Expected behavior:
# - runner.shutdown(agent) -> stops only that specific agent
# - runner.shutdown() -> stops all agents managed by the runner
```

CasperGN:
@bibryam actually now that I think more about this, the `runner.shutdown()` is problematic since it only takes a single agent entry and you're absolutely right, we could have multiple agents.

What about something like this:
```python

# then AgentRunner

class AgentRunner()...

  def __init__():
    self.managed_agents = []
    ..

  def run(): # and the other func's that calls agent.start()
    agent.start()
    self.managed_agents.append(agent)

  def shutdown(agent: Optional[Any]), ..):
    if agent:
      if agent in self.managed_agents:
        agent.stop()
        self.managed_agents.remove(agent)
        return
    # continue with full shutdown
```

Thoughts?

PR #288, opened 2025-12-01
Fix/287 extract structuredcontent mcp tool

### LINKED ISSUE #286: ToolError raised for for DurableAgent propagates to the Durabletask and fails the workflow
Specifically, [this section](https://github.com/dapr/dapr-agents/blob/main/dapr_agents/tool/mcp/client.py#L371-L408):

```python
async def executor(**kwargs: Any) -> Any:
    logger.debug(f"Executing tool '{tool_name}' with args: {kwargs}")
    try:
        logger.info(tool_name, kwargs)
        if (
            client.persistent_connections
            and server_name in client._sessions
        ):
            session = client._sessions[server_name]
            result = await session.call_tool(tool_name, kwargs)
            logger.debug(f"Used persistent session for tool '{tool_name}'")
        else:
            async with client.create_ephemeral_session(
                server_name
            ) as session:
                result = await session.call_tool(tool_name, kwargs)
                logger.debug(
                    f"Used ephemeral session for tool '{tool_name}'"
                )
        return client._process_tool_result(result) # This will raise a ToolError if validation fails
    except Exception as e:
        logger.exception(f"Execution failed for '{tool_name}'")
        raise ToolError(
            f"Error executing tool '{tool_name}': {str(e)}"
        ) from e
    return executor
```

We need to catch the `ToolError` and return it to the generator loop. Pseudo code:

```python
    return client._process_tool_result(result)
except ToolError as tool_err:
    return {"error": tool_err, "input": kwargs}
except Exception as e:
    logger.exception(f"Execution failed for '{tool_name}'")
    raise ToolError(
        f"Error executing tool '{tool_name}': {str(e)}"
    ) from e
```

This should allow the loop to continue the `tool_call`'ing flow.

### LINKED ISSUE #287: MCP Tool result does not return StructuredContent
The [MCPClient's _process_tool_result](https://github.com/dapr/dapr-agents/blob/main/dapr_agents/tool/mcp/client.py#L437-L473) does not return the `StructuredContent` that is in the [spec](https://mcp.mintlify.app/specification/2025-11-25/server/tools#structured-content):

```python
        # Handle error result
        if hasattr(result, "isError") and result.isError:
            error_message = "Unknown error"
            if hasattr(result, "content") and result.content:
                for content in result.content:
                    if hasattr(content, "text"):
                        error_message = content.text
                        break
            raise ToolError(f"MCP tool error: {error_message}")

        # Extract text content from result
        if hasattr(result, "content") and result.content:
            text_contents = []
            for content in result.content:
                if hasattr(content, "text"):
                    text_contents.append(content.text)

            # Return single string if only one content item
            if len(text_contents) == 1:
                return text_contents[0]
            elif text_contents:
                return text_contents
        # Fallback for unexpected formats
        return str(result)
```

Also see the [Python MCP bindings](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/types.py#L1130).

CasperGN:
# Description

Updating MCP Tool return to be `CallToolResult` which includes `structuredContent`. This also catches `ValidationError`, `ToolError` and `Exception` and properly return the `CallToolResult` to the model.

## Issue reference

Please reference the issue this PR will close: 
Fixes #287 
Fixes #286 

## Checklist

* [x] Created/updated tests
    * [X] Created the [dapr/docs](https://github.com/dapr/docs) PR: N/A

### DIFF
--- a/tests/agents/durableagent/test_mcp_streamable_http.py
+++ b/tests/agents/durableagent/test_mcp_streamable_http.py
@@ -1,4 +1,3 @@
-import json
 import pytest
 from unittest.mock import AsyncMock, Mock, patch
 from mcp.types import CallToolResult


dapr_agents/tool/base.py
```diff
@@ -69,140 +65,6 @@ def from_func(cls, func: Callable) -> "AgentTool":
         ToolHelper.check_docstring(func)
         return cls(func=func)
 
-    @classmethod
-    def from_mcp(
```
Cyb3rWard0g, line 73: We do not need these anymore? `from_mcp` and `from_mcp_many`?

CasperGN: They're unused throughout the codebase (except that one test case) and since the mcp client has `wrap_mcp_tool` to basically do the same I opted to remove them.

I ran both the `tox` suite and the full e2e suite with no issues (and tested against a local MCP server as well)

Cyb3rWard0g: mmm the idea was to have `from_mcp_session` to retrieve tools from an active MCP client session. That depends on the other methods also removed. I would not remove it just because there is not an example yet. It is a util that could be useful when a session has been already initialized. Can we review that in another PR and not use this PR to remove them? Do they interfere with the main changes in this PR? 

```
@classmethod
    async def from_mcp_session(
        cls, session: "ClientSession", process_result_fn=None
    ) -> list:
        """
        Fetch all tools and wrap them as AgentTool objects.

        Args:
            session: An active MCP ClientSession.
            process_result_fn: Optional function to process the tool result.

        Returns:
            List[AgentTool]: List of ready-to-use AgentTool objects.
        """
        mcp_tools_response = await session.list_tools()
        return cls.from_mcp_many(
            mcp_tools_response.tools,
            session=session,
            process_result_fn=process_result_fn,
        )
  ```

Comment (CasperGN) [reply dapr_agents/tool/base.py:73]: Absolutely, I'll just need to update the tool call return to mimic how this PR does it since it more or less had duplication of that piece. I'll push something up and @ you!

Comment (CasperGN) [reply dapr_agents/tool/base.py:73]: @Cyb3rWard0g I've been looking at this and I'm not really sure how we can do this neatly and type-safely.
Since there's no schema validation we effectively allow passing in an arbitrary tool with an arbitrary output, like we do in the test-case:
```python
@pytest.fixture
def mock_mcp_tool():
    mcp_tool = Mock()
    mcp_tool.name = "add"
    mcp_tool.description = "Add two numbers"
    # Provide an input schema so the tool expects 'a' and 'b' as direct arguments
    mcp_tool.inputSchema = {
        "type": "object",
        "properties": {"a": {"type": "integer"}, "b": {"type": "integer"}},
        "required": ["a", "b"],
    }
    return mcp_tool

..
agent_tool = AgentTool.from_mcp(mock_mcp_tool, session=mock_mcp_session)
```

If we print the result in the `from_mcp` `executor` the tool result is `4`. Which is nowhere near compatible with `CallToolResult` [from the sdk](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/types.py#L1359-L1365). 
I guess we could in theory do something like this:

```python
async def executor(**kwargs: Any) -> Any:
    try:
        logger.debug(f"Calling MCP tool '{tool_name}' with args: {kwargs}")
        if session is not None:
            result: CallToolResult = await session.call_tool(tool_name, kwargs)
        else:
            logger.debug(f"Starting transport session for tool '{tool_name}'")
            from dapr_agents.tool.mcp.transport import start_transport_session

            async with start_transport_session(connection) as tool_session:
                await tool_session.initialize()
                result: CallToolResult = await tool_session.call_tool(tool_name, kwargs)

        return CallToolResult(
            isError=False,
            content=[TextContent(
                type="text",
                text=str(result)
            )],
            structuredContent={},
        )
    except (ValidationError, ToolError) as e:
        err_type = type(e).__name__
        logger.error(f"{err_type} running tool: {str(e)}")
        return CallToolResult(
            isError=True,
            content=[
                TextContent(
                    type="text",
                    text=f"{err_type} during Tool Call. Arguments sent to Tool: {str(kwargs)}.\nError: {str(e)}",
                )
            ],
        )
    except Exception as e:
        raise ToolError(f"Error executing tool '{tool_name}': {str(e)}") from e
```

However it feels more correct to either require passing in a schema for the model that will map back to a `CallToolResult` or something along those lines.

Thoughts?

CasperGN: Executing the test case on the above yields a result as:
```python
{'tool_call_id': 'call_123', 'tool_name': 'Add', 'tool_args': {'a': 2, 'b': 2}, 'execution_result': '{"meta": null, "content": [{"type": "text", "text": "4", "annotations": null, "meta": null}], "structuredContent": {}, "isError": false}'}
```

Where the `execution_result` holds the equivalent to a `CallToolResult`. It still feels iffy to crudely push it into the format like this but I guess it can serve as a temporary in-between solution until we can come up with a more type-safe setup?

CasperGN:
@Cyb3rWard0g I've returned the `from_mcp*` functions however I've altered the return object to use `CallToolResult` from MCP to align better with the spec. I've also removed the ability to pass a process result func as we should **always** return `CallToolResult` when the result comes from an MCP interaction.

Issue #298, opened 2025-12-10
Further expansion of agent metadata


CasperGN:
To further enhance the ability to catalogue agents we should expose further fields in the metadata:

- `Registered At`: The time for which the agent registered the details (this can be with `time.now()` on startup
- `Agent Type`: This should hold which type the agent is `DurableAgent`, `Agent`, etc.
- `Tool Type`: This should hold whether the tool is from MCP or other (in code for example)

PR #299, opened 2025-12-10
fix: add agent type & registration at fields. Ensure metadata becomes…

CasperGN:
… properly nested for approproate json representation

# Description

The previous metadata object used flat structure with `{"agent.name": ..` rather than proper nesting. This PR also introduce the Agent type details (`Agent`, `DurableAgent`, ..) as well as the registration timestamp. 

It does not include the tool type information as that requires further analysis on what the best approach is - and whether we'd want to differentiate in the model.

## Issue reference

Please reference the issue this PR will closes: #298 

## Checklist

* [X] Created/updated tests
* [X] Tested this change against all the quickstarts
* [X] Extended the documentation


sicoyle (approved):
LGTM - thank you! 🙌

Issue #221, opened 2025-10-01
Durable Agent lacks default resiliency policy for activities


bibryam:
**Description:**
Durable Agent currently does not apply any resiliency policy to activities by default. As a result, a single failure in any activity causes the entire agent workflow to fail.

This behavior reduces robustness for long-running workflows where transient errors are expected.

**Expected Behavior:**

Every activity should, by default, have a resiliency policy applied with infinite retries.
Durable Agent should expose configuration to adjust this default, allowing users to:
- Disable retries (0).
- Set a fixed number of retries.

PR #314, opened 2025-12-18
Fix/221 retrypolicy

CasperGN:
# Description

Implement default retry policy and allow overwriting during agent instantiation.

## Issue reference

Please reference the issue this PR closes: #221 

## Checklist

* [x] Created/updated tests
    * [x] Created the [dapr/docs](https://github.com/dapr/docs) PR: https://github.com/dapr/docs/pull/4982

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,72 @@
+version: 2
+updates:
+  # Git
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    cooldown:
+      default-days: 7
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-patch"]
+  # Python
+  - package-ecosystem: "uv"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    cooldown:
+      default-days: 7
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-patch"]
+  # Docker - Frodo service
+  - package-ecosystem: "docker"
+    directory: "/quickstarts/05-multi-agent-workflow-k8s/services/frodo"
+    schedule:
+      interval: "daily"
+    cooldown:
+      default-days: 7
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-patch"]
+  # Docker - Sam service
+  - package-ecosystem: "docker"
+    directory: "/quickstarts/05-multi-agent-workflow-k8s/services/sam"
+    schedule:
+      interval: "daily"
+    cooldown:
+      default-days: 7
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-patch"]
+  # Docker - Gandalf service
+  - package-ecosystem: "docker"
+    directory: "/quickstarts/05-multi-agent-workflow-k8s/services/gandalf"
+    schedule:
+      interval: "daily"
+    cooldown:
+      default-days: 7
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-patch"]
+  # Docker - Legolas service
+  - package-ecosystem: "docker"
+    directory: "/quickstarts/05-multi-agent-workflow-k8s/services/legolas"
+    schedule:
+      interval: "daily"
+    cooldown:
+      default-days: 7
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-patch"]
+  # Docker - Workflow LLM service
+  - package-ecosystem: "docker"
+    directory: "/quickstarts/05-multi-agent-workflow-k8s/services/workflow-llm"
+    schedule:
+      interval: "daily"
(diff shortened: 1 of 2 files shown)


CasperGN:
@bibryam have a look at the updates

dapr_agents/agents/configs.py
```diff
@@ -265,3 +265,21 @@ class AgentExecutionConfig:
     # TODO: add stop_at_tokens
     max_iterations: int = 10
     tool_choice: Optional[str] = "auto"
+
+
+@dataclass
+class DurableRetryConfig:
```
sicoyle: I think this should be more workflow-specific. Something like WorkflowRetryConfig might be clearer, since this behavior only applies to the durable agent anyway, which makes “durable” feel a bit redundant here.

On retries more broadly, should we consider introducing a resiliency policy in our quickstarts—or at least documenting it—so users know how to apply retry policies for state operations and pub/sub? For example, we could reference the existing Dapr resiliency quickstart or general docs on this:
https://docs.dapr.io/getting-started/quickstarts/resiliency/resiliency-state-quickstart/

dapr_agents/agents/configs.py
```diff
@@ -265,3 +265,21 @@ class AgentExecutionConfig:
...
+        max_attempts: Maximum number of retry attempts.
+        initial_backoff_seconds: Initial backoff interval in seconds.
+        max_backoff_seconds: Maximum backoff interval in seconds.
+        backoff_multiplier: Multiplier for exponential backoff.
+    """
+
+    max_attempts: Optional[int] = 1
+    initial_backoff_seconds: Optional[int] = 5
+    max_backoff_seconds: Optional[int] = 30
+    backoff_multiplier: Optional[float] = 1.5
```
sicoyle, line 296: any reason there are four fields here yet docs show 5 options we can set?
https://docs.dapr.io/developing-applications/building-blocks/workflow/workflow-features-concepts/#retry-policies

dapr_agents/agents/durable.py
```diff
@@ -132,6 +136,25 @@ def __init__(
...
 
+        try:
+            retries = int(getenv("DAPR_API_MAX_RETRIES", ""))
+        except ValueError:
+            retries = retry_policy.max_attempts
+
+        if retries < 1:
+            raise (
+                ValueError("max_attempts or DAPR_API_MAX_RETRIES must be at least 1.")
+            )
```
sicoyle, line 147: we should consider if we want to deviate here. The default that i see in docs is actually a 0 value for this... which I think for durable agent durability we want retries enabled, just want confirmation on that :) 
https://docs.dapr.io/developing-applications/sdks/python/python-client/#initialising-the-client

dapr_agents/agents/configs.py
CasperGN: I'll relabel this `WorkflowRetryPolicy` to be explicit on it

dapr_agents/agents/configs.py
```diff
@@ -265,3 +265,21 @@ class AgentExecutionConfig:
...
+    Configuration for durable retry policies in workflows.
+
+    Attributes:
+        max_attempts: Maximum number of retry attempts.
+        initial_backoff_seconds: Initial backoff interval in seconds.
+        max_backoff_seconds: Maximum backoff interval in seconds.
+        backoff_multiplier: Multiplier for exponential backoff.
+    """
+
+    max_attempts: Optional[int] = 1
```
cicoyle, line 293: The issue is asking for infinite retries by default. In dapr we have `-1` in places to signal infinity. Would it make sense to allow for infinite retries here in that same manner as well?

dapr_agents/agents/configs.py
CasperGN: It's an oversight on my part. I've added the 5th as well!

dapr_agents/agents/durable.py
CasperGN: I have to check on 1, else it might fail [upstream](https://github.com/dapr/python-sdk/blob/main/ext/dapr-ext-workflow/dapr/ext/workflow/retry_policy.py#L53-L54)

dapr_agents/agents/configs.py
CasperGN: Unfortunately that's not how it is in the [upstream](https://github.com/dapr/python-sdk/blob/main/ext/dapr-ext-workflow/dapr/ext/workflow/retry_policy.py#L53-L54) sdk

cicoyle: so as-is, there is no way to specify retries indefinitely? maybe we should tweak the upstream sdk  eventually to support this :)

this PR is a good start!

CasperGN: @cicoyle yes I concur. This should be fixed upstream in the python-sdk then carried down here. I'll put it on our todo list!

CasperGN:
@sicoyle, @cicoyle I've added https://github.com/dapr/python-sdk/issues/870 for upstream fixing

Issue #310, opened 2025-12-16
[Observability] Enable default secret store for agent configuration


CasperGN:
To externalize some of the agent configuration, similarly to #309 we should implement an `agent-runtime` which maps to statestore for runtime configuration and `agent-secretstore` which maps to a secret store. 
This store could potentially hold API keys etc however the initial use-case sought is to allow easy OTel setup following the below structure:

<img width="561" height="399" alt="Image" src="https://github.com/user-attachments/assets/2dff4cb8-6c4b-43fc-9164-0f7c0ef00f07" />

The goal is to limit Agent configuration (if the standard configuration is found to:

```python
from dapr_agents.observability import DaprAgentsInstrumentor

agent = DurableAgent(
    name="Agent",
    role="Agent",
    goal="Some awesome goal"
    instructions=[
        ...
    ],
    tools=[..] + agent_tools,
    llm=OpenAIChatClient(model="gpt-5-2025-08-07", api_key=os.getenv("OPENAI_API_KEY")),
)  

instrumentor = DaprAgentsInstrumentor()
instrumentor.instrument()
```

Today tracing + logging setup requires a lot of boilerplate:

```python
from opentelemetry import trace
from opentelemetry import _logs
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk.trace.export import BatchSpanProcessor
from opentelemetry.exporter.zipkin.json import ZipkinExporter
from opentelemetry.sdk.resources import Resource
from opentelemetry.sdk._logs import LoggerProvider
from opentelemetry.sdk._logs.export import BatchLogRecordProcessor, ConsoleLogRecordExporter
from dapr_agents.observability import DaprAgentsInstrumentor

agent = DurableAgent(
    name="Agent",
    role="Agent",
    goal="Some awesome goal"
    instructions=[
        ...
    ],
    tools=[..] + agent_tools,
    llm=OpenAIChatClient(model="gpt-5-2025-08-07", api_key=os.getenv("OPENAI_API_KEY")),
)  

logger_provider = LoggerProvider()
log_processor = BatchLogRecordProcessor(ConsoleLogRecordExporter())
logger_provider.add_log_record_processor(log_processor)
_logs.set_logger_provider(logger_provider)
resource = Resource(attributes={"service.name": "agent"})
tracer_provider = TracerProvider(resource=resource)
zipkin_exporter = ZipkinExporter(endpoint="http://localhost:9411/api/v2/spans")
[...]

Issue #312, opened 2025-12-17
Revamp memory management of Agents


CasperGN:
Currently the session and agent memory is mixed up rendering the DSL for the Agent easily misunderstood. This issue attempts to encompass the proposed setup for the new memory management setup for Dapr Agents.

As we've progressed the default stores alongside the metadata for the agents, this carries over some of that work and builds on top of it.

Currently Dapr Agents require a `agent-wfstatestore`, `agent-statestore` and `agent-registry`. We are proposing streamlining the naming as well as correlating it with industry standard terms:

<img width="1115" height="343" alt="Image" src="https://github.com/user-attachments/assets/766ffabb-0007-4018-89b2-1b41e30f454c" />

The proposal requires renaming, reusing and creating statestores:

- `agent-wfstatestore` -> Renamed to `agent-workflow`. Today this holds both workflow execution and duplication of the agents session data. Moving forward this store will hold keys that are 1:1 mapped to an agents workflow `instance_id`. This will hold the **short term** memory for the agent for that specific workflow execution. Users should should be able to set a flag to persist short term memory. If set to `False` the key should be removed post execution. Default `True`.
- `agent-statestore` -> Renamed to `agent-memory`. Today this holds the session data for all executed workflows. Moving forward this store will be the **long term** memory and will hold the summarized content (rendered during `finalize_workflow` task) as well as an attempt of extracting general purpose bullets for "good value memory".
- `agent-registry` -> No changes.
- `agent-runtime` -> New statestore containing non-secret runtime configuration that the agent will load **prior** to finalizing instantiation. Examples could be the OpenTelemetry endpoint, llm-model etc.

We will rid the "session" concept as it provides little value conceptually. Instead the workflow execution _is_ the session. For chat-based clients we propose to await the work scoped for post v1.0 release which will introduce streaming (effectively `max_iterations=infinite` relying on workflow termination event).

The following depicts the multi-agent setup and how relying more natively on workflows woudl create stronger bindings also for parent/child relationship:

<img width="1435" height="733" alt="Image" src="https://github.com/user-attachments/assets/4638d2c2-15dc-463f-b211-75413ca10014" />

Co proposed / created with @sicoyle

yaron2:
> agent-runtime -> New statestore containing non-secret runtime configuration that the agent will load prior to finalizing instantiation. Examples could be the OpenTelemetry endpoint, llm-model etc.

[...]

PR #316, opened 2025-12-18
Feat/310 default agent conf

CasperGN:
# Description

This PR allows default configuration and default secrets to be passed to the agent through `agent-secretstore` and `agent-runtime`. This PR also covers automatically setting up OTel tracing if the appropriate configuration parameters are passed through `agent-runtime` statestore. 
This allows operators to automate configuration setup as well as externalize the lifecycle of these configurations.

**NB:** This carries a breaking change with dropping support for Python 3.10.

## Issue reference

Please reference the issue this PR closes: #310 

## Checklist

* [x] Created/updated tests

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -28,12 +28,9 @@ jobs:
       uses: astral-sh/setup-uv@v3
       with:
         version: "latest"
-    - name: Install dependencies
-      run: |
-        uv sync --extra=dev
     - name: Run Autoformatter
       run: |
-        uv run tox -e ruff
+        uv run --no-sync --with tox tox -e ruff
         statusResult=$(git status -u --porcelain)
         if [ -z "$statusResult" ]
         then
@@ -42,9 +39,15 @@ jobs:
             echo "Source files are not formatted correctly. Run 'tox -e ruff' to autoformat."
             exit 1
         fi
+    - name: Clean up tox env
+      if: always()
+      run: rm -rf .tox/ruff
     - name: Run Linter
       run: |
-        uv run tox -e flake8
+        uv run --no-sync --with tox tox -e flake8
+    - name: Clean up tox env
+      if: always()
+      run: rm -rf .tox/flake8
 
   build:
     needs: lint
--- a/tox.ini
+++ b/tox.ini
@@ -17,21 +17,21 @@ pip_pre = True
 
 [testenv:flake8]
 basepython = python3
-usedevelop = False
+skip_install = True
 deps = flake8
 commands =
     flake8 dapr_agents tests --ignore=E501,F401,W503,E203
 
 [testenv:ruff]
 basepython = python3
-usedevelop = False
+skip_install = True
 deps = ruff==0.2.2
 commands =
     ruff format
 
 [testenv:type]
 basepython = python3
-usedevelop = False
+skip_install = True
 deps = 
     -e .[dev]
     mypy
@@ -40,5 +40,7 @@ commands =
 
 [testenv:pytest]
 basepython = python3
-usedevelop = True
+skip_install = True
+deps = 
+    -e .[test]
 commands = pytest tests -m "not integration"


CasperGN:
@sicoyle before I get too far it would be good to align on the assumptions I take on naming:

**Secrets:**
- `OTEL_TOKEN` -> Bearer token for OTLP auth

**Conf:**
- `OTEL_ENDPOINT` -> Str: endpoint to otel collectors
- `OTEL_ENABLED` -> Bool: whether to init otel
- `OTEL_TRACING_ENABLED` -> Bool: Whether to init tracing
- `OTEL_LOGGING_ENABLED` -> Bool: Whether to init logging
- `OTEL_TRACING_EXPORTER` -> Enum: Type of exporter (`otlp_grpc`, `otlp_http`, `zipkin`, `console`, defaults to latter)
- `OTEL_LOGGING_EXPORTER` -> Enum: Type of exporter (`otlp_grpc`, `otlp_http`, `console`, defaults to latter)

I was thinking we could also allow users to specify this via env variables on equal terms.

dapr_agents/agents/components.py
```diff
@@ -67,38 +98,80 @@ def __init__(
...
+            components: Sequence[RegisteredComponents] = resp.registered_components
+            for component in components:
+                if (
+                    "state" in component.type
+                    and component.name == "agent-wfstatestore"
+                    and state is None
+                ):
+                    state = AgentStateConfig(
+                        store=StateStoreService(store_name=component.name),
+                        state_key=f"{name.replace(' ', '-').lower() if name else 'default'}:workflow_state",
```
sicoyle:
```suggestion
                        state_key=f"{name.replace(' ', '-').lower() if name else 'default'}:agent_workflow",
```

https://github.com/dapr/dapr-agents/issues/312

dapr_agents/agents/components.py
```diff
@@ -67,38 +98,80 @@ def __init__(
...
+                    and registry is None
+                ):
+                    registry = AgentRegistryConfig(
+                        store=StateStoreService(store_name="agent-registry"),
+                        team_name="default",
+                    )
+                if "state" in component.type and component.name == "agent-runtime":
+                    raw_runtime_conf: StateResponse = _client.get_state(
+                        store_name=component.name,
+                        key="agent_runtime_configuration",
```
sicoyle: if above we have agent-wfstatestore with key agent_workflow, then should we do agent-runtimestatestore with key agent_runtime?

dapr_agents/agents/components.py
```diff
@@ -67,38 +98,80 @@ def __init__(
...
+                            json.loads(raw_runtime_conf.data)
+                            if raw_runtime_conf.data
+                            else {}
                         )
-                    if component.name == "agent-registry" and registry is None:
-                        registry = AgentRegistryConfig(
-                            store=StateStoreService(store_name="agent-registry"),
-                            team_name="default",
+                        for key, value in self._runtime_conf.items():
+                            logger.info(f"Runtime configuration: {key}={value}")
```
sicoyle:
```suggestion
                            logger.debug(f"Runtime configuration: {key}={value}")
```

dapr_agents/agents/components.py
```diff
@@ -67,38 +98,80 @@ def __init__(
...
-                            agent_topic=f"{name.replace(' ', '-').lower()}.topic",
-                            broadcast_topic="agents.broadcast",
+                if (
+                    "pubsub" in component.type
+                    and component.name == "agent-pubsub"
+                    and pubsub is None
+                ):
+                    logger.info(f"topic: {name}.topic")
+                    pubsub = AgentPubSubConfig(
+                        pubsub_name="agent-pubsub",
```
sicoyle:
```suggestion
                        pubsub_name=component.name,
```

dapr_agents/agents/components.py
```diff
@@ -67,38 +98,80 @@ def __init__(
...
+                        agent_topic=f"{name.replace(' ', '-').lower()}.topic",
+                        broadcast_topic="agents.broadcast",
```
sicoyle: what if someone has set agent-pubsub and we're picking it up, but they set a diff broadcast_topic? Same for the state components... then, we would be overriding their value.

dapr_agents/agents/components.py
```diff
@@ -67,38 +98,80 @@ def __init__(
...
+                    )
+                if (
+                    "secretstores" in component.type
+                    and component.name == "agent-secretstore"
+                ):
+                    try:
+                        agent_secrets: GetBulkSecretResponse = _client.get_bulk_secret(
+                            store_name="agent-secretstore"
+                        )
+                        logger.info(
```
sicoyle:
```suggestion
                        logger.debug(
```

dapr_agents/agents/components.py
```diff
@@ -788,6 +861,110 @@ def _get_entry_container(self) -> Optional[dict]:
...
+        # OTel setup
+        if self._runtime_conf.get("OTEL_ENABLED", "false").lower() == "true":
+            # Set resource name for tracing and logging
+            resource = Resource(
+                attributes={"service.name": self.name.replace(" ", "-").lower()}
+            )
+
+            otlp_headers: dict[str, str] = {}
+
+            otel_token = self._runtime_secrets.get("OPEN_TELEMETRY_TOKEN", "")
```
sicoyle:
```suggestion
            otel_token = self._runtime_secrets.get("OTEL_TOKEN", "")
```

dapr_agents/agents/components.py
```diff
@@ -788,6 +861,110 @@ def _get_entry_container(self) -> Optional[dict]:
...
+                            endpoint=_endpoint, headers=otlp_headers
+                        )
+                    case "otlp_http":
+                        tracing_exporter = OTLPHTTPSpanExporter(
+                            endpoint=_endpoint, headers=otlp_headers
+                        )
+                    case "zipkin":
+                        tracing_exporter = ZipkinExporter(endpoint=_endpoint)
+                    case _:
+                        tracing_exporter = ConsoleSpanExporter()
```
sicoyle: is jaeger captured within this too? https://github.com/dapr/dapr-agents/tree/main/quickstarts/09-agent-observability

dapr_agents/agents/components.py
```diff
@@ -788,6 +861,110 @@ def _get_entry_container(self) -> Optional[dict]:
...
+
+                span_processor = BatchSpanProcessor(tracing_exporter)
+                tracer_provider.add_span_processor(span_processor)
+                trace.set_tracer_provider(tracer_provider)
+
+            instrumentor = DaprAgentsInstrumentor()
+            instrumentor.instrument(
+                tracer_provider=tracer_provider,
+                logger_provider=logger_provider,
+                skip_dep_check=True,
```
sicoyle: what does skip_dep_check do? can we rm?

sicoyle:
> @sicoyle before I get too far it would be good to align on the assumptions I take on naming:

Looks great so far!

Instead of an env var equivalent what are your thoughts on if we add an otelConfig dataclass instead as an optional secondary setup to reduce the amt of env vars we support.

CasperGN:
> > @sicoyle before I get too far it would be good to align on the assumptions I take on naming:
> 
> Looks great so far!
> 
> Instead of an env var equivalent what are your thoughts on if we add an otelConfig dataclass instead as an optional secondary setup to reduce the amt of env vars we support.

@sicoyle I think this is a better approach. We'd still have to support the "usual suspects" (i.e., endpoint & token) but it would substantially lower the amount!

dapr_agents/agents/components.py
CasperGN: @sicoyle I belive it was a left over from previously. I've removed it and also removed it from all the quickstarts.

In the same swoop I also removed the conditional checks on dependencies in the observability stack as the otel, wrapt and openinference deps are now part of the core dependencies

dapr_agents/agents/components.py
CasperGN: Jaeger supports OTLP so we can ship to Jaeger backend with the `otlp_grpc` and `otlp_http` ✅

dapr_agents/agents/components.py
CasperGN: Uf, this is a good point. Let me try to handle this scenario better!

dapr_agents/agents/components.py
CasperGN: Fixed!

dapr_agents/agents/components.py
CasperGN: Actually, shouldn't that be covered by the conditional:
```python
if (
    "pubsub" in component.type
    and component.name == "agent-pubsub"
    and pubsub is None
):
  ...
```

If they've added:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: agent-pubsub
spec:
  type: pubsub.redis
  version: v1
  metadata:
    - name: redisHost
      value: localhost:6379
```

But they have not added it in the agent instantiation then this does exactly what we want it to, no?
`"pubsub" in component.type` -> We found the component
`and component.name == "agent-pubsub"` -> It has the right default naming
`and pubsub is None` -> And they didn't pass any during instantiation

Then the agent would pick it up as the default pubsub

dapr_agents/observability/instrumentor.py
```diff
@@ -381,6 +359,18 @@ def _uninstrument(self, **kwargs: Any) -> None:
         Args:
             **kwargs: Uninstrumentation configuration (unused)
         """
+        logger.info("Uninstrumenting Dapr Agents OpenTelemetry instrumentation")
```
sicoyle:
```suggestion
        logger.debug("Uninstrumenting Dapr Agents OpenTelemetry instrumentation")
```

dapr_agents/observability/instrumentor.py
```diff
@@ -381,6 +359,18 @@ def _uninstrument(self, **kwargs: Any) -> None:
...
+
+        try:
+            tracer_provider: TracerProvider = trace_api.get_tracer_provider()
+            if hasattr(tracer_provider, "force_flush"):
+                tracer_provider.force_flush(timeout_millis=5000)  # type: ignore
+                logger.info("Flushed tracer provider spans")
```
sicoyle:
```suggestion
                logger.debug("Flushed tracer provider spans")
```

dapr_agents/observability/instrumentor.py
```diff
@@ -381,6 +359,18 @@ def _uninstrument(self, **kwargs: Any) -> None:
...
+        except Exception:  # noqa: BLE001
+            logger.info("Error while shutting down tracer provider", exc_info=True)
```
sicoyle:
```suggestion
            logger.exception("Error while shutting down tracer provider", exc_info=True)
```

dapr_agents/workflow/runners/agent.py
```diff
@@ -605,6 +605,12 @@ def shutdown(self, agent: Optional[AgentComponents]) -> None:
                 if agent in self._managed_agents:
                     agent.stop()  # This is safe as they'll return None if not started
                     self._managed_agents.remove(agent)
+                if len(self._managed_agents) == 0:
+                    try:
+                        self.unwire_pubsub()
+                    finally:
+                        self._close_dapr_client()
+                        self._close_wf_client()
```
sicoyle: should we close wf client before dapr client?

sicoyle (commented):
few comments so far and will complete my review in my morning. Thank you!!

dapr_agents/workflow/runners/agent.py
CasperGN: Agreed and done ✅

dapr_agents/agents/base.py
```diff
@@ -115,6 +155,104 @@ def __init__(
...
+                        except json.JSONDecodeError:
+                            logger.warning(
+                                "Failed to decode agent runtime configuration JSON. Using empty configuration."
+                            )
+                    if (
+                        "pubsub" in component.type
+                        and component.name == "agent-pubsub"
+                        and pubsub is None
+                    ):
+                        logger.info(f"topic: {name}.topic")
```
```suggestion
                        logger.debug(f"topic: {name}.topic")
```

PR #398, opened 2026-01-22
feat: Introduce '03-standalone-agent-tool-call' quickstart, including…

0Xuser100:
# Bug Report & Proposed Fixes

This document summarizes the issues found in the `03-standalone-agent-tool-call` quickstart and the technical fixes implemented to enable a successful run on Windows and ensure YAML validation.

## 1. YAML Syntax Error (Indentation)
**Issue:** Several component YAML files had invalid indentation. The list items (`- name:`) under the `metadata:` block were aligned vertically with the `metadata` key itself.
**Impact:** Dapr failed to start with the error: `yaml: line 9: did not find expected key`.
**Fix:** Indented all list items under `metadata:` by 2 extra spaces (total 4 spaces from the start).

**Affected Files:**
- `components/openai.yaml`
- `components/historystore.yaml`
- `components/agentstatestore.yaml`
- `components/workflowstatestore.yaml`

## 2. Missing Windows (PowerShell) Support
**Issue:** The README only provided Bash commands (`export`, `grep`, `xargs`). These do not work in native Windows CMD or PowerShell.
**Impact:** Windows users encountered `'export' is not recognized` errors.
**Fix:** Added a dedicated "Windows (PowerShell)" section in the README with compatible commands for loading `.env` files and managing variables.

## 3. Undocumented Requirement: `dapr init`
**Issue:** The README did not mention that the Dapr runtime must be initialized before the first run.
**Impact:** Users received `fork/exec ... daprd.exe: The system cannot find the path specified`.
**Fix:** Added `dapr init` as a core prerequisite in the README.

## 4. Virtual Environment Isolation
**Issue:** Running `dapr run ... python script.py` often defaulted to the system Python, missing the `dapr-agents` library installed via `uv`.
**Impact:** `ModuleNotFoundError: No module named 'dapr_agents'`.
**Fix:** Updated all execution commands in the README to use `uv run`. This ensures the command executes within the correct virtual environment with all dependencies available.

## 5. Relative Path Cleanup
**Issue:** The README referenced `../../.env` but the `.env` file is commonly placed inside the quickstart folder itself for local testing.
**Fix:** Updated instructions to reference the local `.env` and added troubleshooting tips for path errors.


0Xuser100:
Bug Report & Proposed Fixes

quickstarts/03-standalone-agent-tool-call/README.md
```diff
@@ -32,20 +34,39 @@ OPENAI_API_KEY=your_api_key_here
...
 
 # Run your dapr command with the temporary resources
-dapr run --app-id weatheragent --resources-path $temp_resources_folder -- python standalone_weather_agent_dapr.py
+uv run dapr run --app-id weatheragent --resources-path $temp_resources_folder -- python standalone_weather_agent_dapr.py
 
 # Clean up when done
 rm -rf $temp_resources_folder
 ```
 
+#### Windows (PowerShell)
```
sicoyle: Thank you so much for bridging the gap for us and powershell here. This is something we will need rolled out to the rest of the quickstarts if you are up to applying this same change in another PR to the rest of the quickstarts pls? 🙏

sicoyle:
Please click on the DCO build step failure and run the commands it suggests to make it pass 🙏

quickstarts/03-standalone-agent-tool-call/README.md
```diff
@@ -32,20 +34,39 @@ OPENAI_API_KEY=your_api_key_here
...
 # Get the environment variables from the .env file:
-export $(grep -v '^#' ../../.env | xargs)
+export $(grep -v '^#' .env | xargs)
 
 # Create a temporary resources folder with resolved environment variables
 temp_resources_folder=$(../resolve_env_templates.py ./components)
 
 # Run your dapr command with the temporary resources
-dapr run --app-id weatheragent --resources-path $temp_resources_folder -- python standalone_weather_agent_dapr.py
+uv run dapr run --app-id weatheragent --resources-path $temp_resources_folder -- python standalone_weather_agent_dapr.py
```
CasperGN: Adding `uv run` here isn't necessary

quickstarts/03-standalone-agent-tool-call/README.md
```diff
@@ -75,7 +96,7 @@ This example shows how to create tools and an agent that can use them.
 
 Run the weather agent:
 ```bash
-python standalone_weather_agent_openai.py
+uv run python standalone_weather_agent_openai.py
```
CasperGN: Same applies here and on line 196, 250 and 255. 
As we source the `.venv` during environment setup the packages should be readily available - unless this works differently on Windows?

quickstarts/03-standalone-agent-tool-call/README.md
0Xuser100: ok i do that

quickstarts/03-standalone-agent-tool-call/README.md
0Xuser100: yes it diffenetly in window you have 2 option
1-activate venv
2- use uv to run

CasperGN: The same applies on nix versions though. As long as you've activated (sourced) the venv the dependencies should be updated and the `PYTHONPATH` should be pointing to your sourced `.venv` which has the dependencies installed.

It does raise an interesting point though. We can lean towards more heavily relying on `uv` (which we already kinda started when removing tox) or we can stay in this middle ground.

The more I reflect on this the more I may be inclined to favor your approach, generally speaking. If we let `uv` handle this we can remove some of the steps with setting up the virtual environment etc (since `uv` will do that with `uv run ...`).

@sicoyle what's your thoughts here? I might be inclined to agree with @0Xuser100 and push all quickstarts to use `uv run ..`. Will we have any dapr specifics here that'll cause problems?

sicoyle: yes i am in favor of us leaning in more on `uv` and like this change

quickstarts/05-multi-agent-workflow-k8s/services/frodo/Dockerfile
sicoyle, line 20: can you pls rebase and see if this change included in your PR goes away? 🙏

sicoyle (commented):
can you pls rebase or grab the latest from main one last time and resolve the conflicts and we can merge this shortly after. It looks good overall, just need the build all green 👏 Thank you for your contributions and patience! This is on my radar to check back asap so we can avoid you constantly updating the branch 😄

PR #406, opened 2026-01-27
build: fix uv support and update quickstart setup instructions

bibryam:
# Description

This fixes two issues:
 - The instructions in the quickstart readme didn't work, such as these ones: uv sync --active
 - Also it doesn't tell user to run the instructions on top level
 - But at top level, the pyproject.toml also fail to install with uv, so fixed that.
  
  
 If this is the right fix, I  can update all quickstart readmes with this change.


bibryam:
Updating UV fixed the error, this PR is not needed

PR #422, opened 2026-02-02
Demonstrate how durable agent can orchestrate other durable agents

bibryam:
# This demonstrates how to use durable agents to orchestrate other durable agents. 

The durable agent needs access to the workload context in order to call other durable agents as child workloads. 
The current implementation has the following limitations:

- Errors in inline tools fail the workflow directly (no ToolResult wrapping like activity tools).
- Detection via _workflow_ctx parameter name is brittle and easy to miss/typo.
- Inline tools run sequentially and can reduce parallelism vs activity‑based tools.
- Exposes full workflow context to tool code; safe only for trusted tools.

### DIFF
--- a/quickstarts/04-agent-based-workflows/manager_agen_with_wf.py
+++ b/quickstarts/04-agent-based-workflows/manager_agen_with_wf.py
@@ -1,6 +1,9 @@
 from __future__ import annotations
 
+import json
 import logging
+import uuid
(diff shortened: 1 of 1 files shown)


sicoyle:
@bibryam is this still WIP or can we close it?

sicoyle:
closing for now... we can reopen in future if we want :)

PR #438, opened 2026-02-05
refactor: state revamp

sicoyle:
# Description

One workflow instance = one state entry now = the new concept of a "session".

Mappings for “instances” within workflow_state are gone now:
- ensure_instance_exists
- entry_container
- ensure_instance_row
- AgentWorkflowState and LLMWorkflowState wrappers as they are no longer needed
 
Workflow changes:
- No more load state calls everywhere as that is no longer needed bc we can get the workflow instance from the activity context and agent base loads state for us.
- Activities use the workflow context for the current instance instead of reloading full state every time; state is loaded only for the instance in scope.

Keys
- Workflow state: agentname_workflow_<workflow_id>
- Memory: agentname_memory_<workflow_id>.

Keys should all be lowercased and if spaces those are transformed to dashes.

There is no longer the concept of a session_id that users can set, and no duplicate workflow metadata stored within our repo. IE I dropped from our state anything Dapr already exposes via the workflow client: input_value, output, start_time, end_time, status, workflow_instance_id, workflow_name. Users can use our fast API endpoint for that. I also added in the endpoints from Caspers PR for purge and terminate and updated various state stores accordingly.

Memory & summarization
- Long-term memory uses a summarization activity and stores a summary per workflow. Memory store API now requires workflow_instance_id on add/get/reset; agent_name is required and set by the framework.

NOTE: I did verify that a resumed workflows state is as expected.

TODO: I think I can maybe delete load_state in favor of get_state now that I'm typing this out..? TBD reminder to self lol.

## Visuals:
Before high level:
image

Before data content:
image

After high level (ignore the blue pls):
image

After data content (ignore the pink and black pls):
image

## Issue reference

https://github.com/dapr/dapr-agents/issues/312

## Checklist

* [ ] Created/updated tests

### DIFF
--- a/dapr_agents/llm/dapr/chat.py
+++ b/dapr_agents/llm/dapr/chat.py
@@ -381,13 +381,32 @@ class DaprChatClient(DaprInferenceClientBase, ChatClientBase):
             if "structured_mode" in params:
                 api_params["structured_mode"] = str(params["structured_mode"])
 
+            # Convert tool_choice from dict format to string of 'none', 'auto', or 'required'
+            tool_choice_param = params.get("tool_choice")
+            if isinstance(tool_choice_param, dict):
+                # When a specific function is requested (dict format), use 'required'
+                # to force tool calling, since Dapr doesn't support function-specific selection
+                tool_choice_param = "required"
+            elif isinstance(tool_choice_param, str) and tool_choice_param not in (
+                "none",
+                "auto",
+                "required",
+            ):
+                # If it's a string but not a valid Dapr value (e.g., a function name),
+                # default to 'required' to force tool calling
+                logger.warning(
+                    f"tool_choice value '{tool_choice_param}' is not supported by Dapr. "
+                    "Using 'required' instead."
+                )
+                tool_choice_param = "required"
+
             raw = self.client.chat_completion_alpha2(
                 llm=llm_component or self._llm_component,
                 inputs=conv_inputs,
                 scrub_pii=scrubPII,
                 temperature=temperature,
                 tools=params.get("tools"),
-                tool_choice=params.get("tool_choice"),
+                tool_choice=tool_choice_param,
                 parameters=api_params or None,
             )
             normalized = self.translate_response(
--- a/dapr_agents/tool/utils/function_calling.py
+++ b/dapr_agents/tool/utils/function_calling.py
@@ -162,7 +162,7 @@ def to_function_call_definition(
     fmt = format_type.lower()
 
     # OpenAI‑style wrapper schema:
-    if fmt in ("openai", "nvidia", "huggingface"):
+    if fmt in ("openai", "nvidia", "huggingface", "dapr"):
         return to_openai_function_call_definition(
             name, description, args_schema, use_deprecated
         )
--- a/tests/integration/quickstarts/conftest.py
+++ b/tests/integration/quickstarts/conftest.py
@@ -927,6 +927,9 @@ def _run_multi_app_with_completion_detection(
                 elif "workflow_agents" in yaml_name or "09_workflow" in yaml_name:
                     # 09_workflow_agents: main workflow is support_workflow; ignore broadcast_listener
                     orchestrator_workflow_name = "support_workflow"
+                elif "sequential" in yaml_name:
+                    # Sequential workflow quickstart: main workflow is chained_planner_workflow
+                    orchestrator_workflow_name = "chained_planner_workflow"
                 break
 
     logger.info(f"Running multi-app command with completion detection: {' '.join(cmd)}")


sicoyle:
this should be g2g for review pls @CasperGN NOTE: I did not test orchestrators and will do that after we merge yours and then update my branch accordingly 😄 
also pls dont let me forget the todo above in my description

dapr_agents/agents/orchestrators/llm/base.py
```diff
@@ -90,22 +90,19 @@ def __init__(
         if self._memory.store is None and state is not None:
             self._memory.store = ConversationDaprStateMemory(
                 store_name=state.store.store_name,
-                session_id=f"{self.name}-session",
+                agent_name=self.name,
             )
         self.memory = self._memory.store or ConversationListMemory()
+        # TODO(@sicoyle): if we like this then i need to update this for all other memory stores.
```
CasperGN, line 96: Good with me ✅

dapr_agents/agents/base.py
```diff
@@ -674,6 +668,84 @@ def get_last_message_if_user(
...
+            f"{conversation_text}\n\n"
+            "Tool calls/results:\n"
+            f"{json.dumps(tool_list, default=str)}"
+        )
+        llm_messages: List[Dict[str, Any]] = [
+            {"role": "user", "content": task},
+        ]
+
+        try:
+            response: LLMChatResponse = self.llm.generate(messages=llm_messages)
```
CasperGN: Should we consider a specific `response_format` here making it easier to work with the memory?

dapr_agents/agents/components.py
```diff
@@ -287,65 +322,61 @@ def save_state(self) -> None:
...
-
-    def ensure_instance_exists(
-        self,
-        *,
-        instance_id: str,
-        input_value: Any,
-        triggering_workflow_instance_id: Optional[str],
-        time: Optional[datetime] = None,
-    ) -> None:
+    def purge_state(self, workflow_instance_id: str) -> None:
```
CasperGN, line 328: Should this not also look at the long term memory and purge the entry for the `workflow_instance_id`?

dapr_agents/agents/components.py
```diff
@@ -738,21 +772,21 @@ def _ensure_registry_initialized(self, *, key: str, meta: Dict[str, str]) -> Non
...
-            the underlying state model does not expose such a container.
-
-        Notes:
-            Prefer a caller-provided hook via `AgentStateConfig.entry_container_getter`.
-            Falls back to `model.instances` for legacy/default shapes.
-        """
-        if self._entry_container_getter:
-            return self._entry_container_getter(self._state_model)
-        return getattr(self._state_model, "instances", None)
+    # def _get_entry_container(self) -> Optional[dict]:
```
CasperGN: Should this not just be removed?

dapr_agents/agents/standalone.py
```diff
@@ -234,14 +218,35 @@ def construct_messages(
...
-        self.load_state()
         active_instance = instance_id or self._generate_instance_id()
+        self._infra.load_state(active_instance)
         chat_history = self._reconstruct_conversation_history(active_instance)
         return self.prompting_helper.build_initial_messages(
             user_input=input_data,
             chat_history=chat_history,
         )
 
+    def _summarize(self, instance_id: str) -> Dict[str, Any]:
```
CasperGN, line 229: This isn't going to work for the `Agent` as it don't run with workflow execution.

This is probably a good time to make a decision: Remove non-durable `Agent` type or no?
cc @yaron2, @Cyb3rWard0g

dapr_agents/memory/daprstatestore.py
```diff
@@ -13,77 +12,59 @@
...
-        return values
-
     def model_post_init(self, __context: Any) -> None:
         """
         Initializes the Dapr state store after validation.
         """
         self.dapr_store = DaprStateStore(store_name=self.store_name)
         logger.info(
-            f"ConversationDaprStateMemory initialized with session ID: {self.session_id}"
+            "ConversationDaprStateMemory initialized (store=%s, agent_name=%s)",
```
CasperGN: Can we make this f-string instead?

dapr_agents/memory/daprstatestore.py
```diff
@@ -123,43 +104,61 @@ def add_message(self, message: Union[Dict[str, Any], BaseMessage]) -> None:
             except Exception as exc:
                 if attempt == max_attempts:
                     logger.exception(
-                        f"Failed to add message to session {self.session_id} after {max_attempts} attempts: {exc}"
+                        "Failed to add message to workflow instance %s after %s attempts: %s",
```
CasperGN: Same with f-string and one below

dapr_agents/memory/daprstatestore.py
```diff
@@ -123,43 +104,61 @@ def add_message(self, message: Union[Dict[str, Any], BaseMessage]) -> None:
...
+        Adds multiple messages to the memory and saves each one to the Dapr state store.
 
         Args:
-            messages (List[Union[Dict[str, Any], BaseMessage]]): A list of messages to add to the memory.
+            messages: A list of messages to add to the memory.
+            workflow_instance_id: Workflow instance id for these messages.
         """
-        logger.info(f"Adding {len(messages)} messages to session {self.session_id}")
+        logger.info(
+            "Adding %s messages to workflow instance %s",
```
CasperGN: Same with f-string

dapr_agents/memory/daprstatestore.py
```diff
@@ -175,45 +174,57 @@ def _decode_message(self, message_data: Union[bytes, str]) -> Dict[str, Any]:
...
         """
-        response = self.query_messages(session_id=self.session_id)
+        response = self.query_messages(workflow_instance_id)
         if response and hasattr(response, "data") and response.data:
             raw_messages = json.loads(response.data)
             if raw_messages:
                 messages = raw_messages[:limit]
                 logger.info(
-                    f"Retrieved {len(messages)} messages for session {self.session_id}"
+                    "Retrieved %s messages for workflow instance %s",
```
CasperGN: f-string please

dapr_agents/workflow/runners/agent.py
```diff
@@ -418,8 +418,8 @@ def serve(
         host: str = "0.0.0.0",
         port: int = 8001,
         expose_entry: bool = True,
-        entry_path: str = "/run",
-        status_path: str = "/run/{instance_id}",
+        entry_path: str = "/agent/run",
+        status_path: str = "/agent/instances/{instance_id}",
```
CasperGN, line 422: Remember this in the docs pr and the returned json on the invocation call

dapr_agents/agents/components.py
sicoyle: I will add the fix for this in a separate follow up PR. I feel ok adding the response format change without testing but this one I want to test before committing. Will create follow up issue to ensure I track for another PR when I get my env back

sicoyle: https://github.com/dapr/dapr-agents/issues/445

dapr_agents/agents/standalone.py
Cyb3rWard0g: I believe we should remove the standalone agent and make the whole project durable first :)

CasperGN: Sounds good to me!

Issue #135, opened 2025-06-12
Enable Agents as Tools (In-process call)


bibryam:
**Background** The current Dapper Agents platform supports multi-agent interactions via asynchronous messaging, which can be overkill for tightly-coupled workflows. This feature introduces the ability for agents to invoke other agents synchronously (through in-process call) as tools. It allows modular agent design where a central agent can delegate tasks to specialized agents dynamically, with direct, inline responses.

<img src="https://uploads.linear.app/5ff56eb9-d5ff-4cd7-a2c9-fc0dbcb5cc2d/794905f9-3abb-44e9-99f6-e04cee91392d/d2e0628d-4ffe-4df0-ad60-72574b82b998?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzVmZjU2ZWI5LWQ1ZmYtNGNkNy1hMmM5LWZjMGRiY2I1Y2MyZC83OTQ5MDVmOS0zYWJiLTQ0ZTktOTlmNi1lMDRjZWU5MTM5MmQvZDJlMDYyOGQtNGZmZS00ZGYwLWFkNjAtNzI1NzRiODJiOTk4IiwiaWF0IjoxNzU2ODI5OTQ5LCJleHAiOjMzMzI3Mzg5OTQ5fQ.JM2OmkCDj4WciHBE-NMqFrPTe8f6dD_ogYMqgeFhh0o " alt="5.png" width="1532" data-linear-height="826" />

**Scenarion**
You define a parent agent (the Orchestrator) responsible for handling user input and delegating tasks. Separately, you define two specialized agents:

* `FlightSearchAgent`, a flight search expert that uses a search_flights tool to find flight options.
* `HotelSearchAgent`, a hotel booking expert that uses a search_hotels tool to find hotel options.
  Each specialized agent is defined with its own purpose, tools, and instructions. You then wrap both agents using agent_tool() and include them in the Orchestrator agent’s tool list. This setup allows the Orchestrator to synchronously invoke the specialized agents at runtime, just like any other tool, enabling tightly-coupled, multi-agent workflows.

**Acceptance Criteria**
Support wrapping an agent as a callable tool
Allow an agent to include other agents in its toolset.
Invocation is synchronous and returns the child agent's final result.

**Pseudo Code**

```
 from dapr_agents import Agent, tool, agent_tool
from pydantic import BaseModel
from typing import List

# Schemas
[...]

PR #448, opened 2026-02-15
feat: enable hot-reloading of DurableAgent through Dapr Configuration Store with subscribing

CasperGN:
# Description

This PR introduces `AgentConfigurationConfig` which allows a `DurableAgent` to subscribe to a Dapr [Configuration Store](https://docs.dapr.io/reference/components-reference/supported-configuration-stores/). The agent hot-reloads (and updates agent metadata!) the following:

`agent_role`
`agent_goal`
`agent_instruction`
`agent_system_prompt`
`llm_api_key` or `openai_api_key`
`llm_provider` -> update ref to component
`llm_model`
`state_store` -> update ref to component
`registry_store` -> update ref to component
`memory_store` -> update ref to component

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR closes: #364 

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Created/updated tests
* [x] Tested this change against all the quickstarts
* [x] Extended the documentation
    * [ ] Created the [dapr/docs](https://github.com/dapr/docs) PR: TODO

**Note:** We expect contributors to open a corresponding documentation PR in the [dapr/docs](https://github.com/dapr/docs/) repository. As the implementer, you are the best person to document your work! Implementation PRs will not be merged until the documentation PR is opened and ready for review.

PR #455, opened 2026-02-26
feat: agents as tools

sicoyle:
# Description

Enables one DurableAgent to call another as a Dapr child workflow via the LLM tool interface.

This means I've added an AgentWorkflowTool class with a converter func called agent_to_tool that under the hood leverages WorkflowContextInjectedTool to schedule another agent's workflow as a child workflow without exposing ctx within the LLM schema. 

In order for a DurableAgent to be used as a tool to another DurableAgent, you can:
3. not specify any tools on the calling agent, and so long as the agents are in the same registry team, then the calling agent auto registers the callable (agent as a tool) as a tool available

This means from a DX, there is now a new activity call `load_tools` to within runtime load agent tools and in future we can load more tool types here at runtime leveraging the registry.

I added an example to showcase this as it felt more advanced than just the fundamentals.

I also updated our workflow name to be <agent_name>_agent_workflow but am happy to rename

## Issue reference

Please reference the issue this PR closes: https://github.com/dapr/dapr-agents/issues/135

## Checklist

* [ ] Created/updated tests


sicoyle:
ignore the schema metadata comment as that ran when I pushed without my branch up-to-date with everything in main 😅

examples/09-agents-as-tools/dapr-cross-app.yaml
```diff
@@ -0,0 +1,21 @@
+version: 1
+common:
+  resourcesPath: ./resources
```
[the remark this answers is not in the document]
sicoyle: i updated the tests to support both components and resources dir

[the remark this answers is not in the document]
CasperGN: This is also part of the #448 pr

dapr_agents/agents/base.py
```diff
@@ -835,16 +846,7 @@ def get_llm_tools(self) -> List[Union[AgentTool, Dict[str, Any]]]:
...
-        for tool in self.tools:
-            if isinstance(tool, AgentTool):
-                llm_tools.append(tool)
-            elif callable(tool):
-                try:
-                    llm_tools.append(AgentTool.from_func(tool))
-                except Exception as exc:  # noqa: BLE001
-                    logger.warning("Failed to convert callable to AgentTool: %s", exc)
-        return llm_tools
+        return list(self.tool_executor._tools_map.values())
```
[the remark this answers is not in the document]
CasperGN: @sicoyle can we do as copilot suggest here and add a public accessor function?

examples/09-agents-as-tools/resources/agent-runtime.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: agent-runtimestatestore
```
CasperGN:
```suggestion
  name: agent-runtime
```

quickstarts/09_workflow_agents.py
```diff
@@ -8,17 +8,19 @@
 @wfr.workflow(name="support_workflow")
 def support_workflow(ctx: wf.DaprWorkflowContext, request: dict) -> str:
     """Process a support request through triage and expert agents."""
-    # Each step is durable and can be retried
+    # Each step is durable and can be retried.
+    # Workflow names follow the "{agent_name}_agent_workflow" convention
+    # set by DurableAgent.register_workflows().
     triage_result = yield ctx.call_child_workflow(
-        workflow="agent_workflow",
+        workflow="triage_agent_agent_workflow",
```
CasperGN: Same q here

dapr_agents/agents/durable.py
```diff
@@ -673,12 +787,12 @@ def orchestration_workflow(self, ctx: wf.DaprWorkflowContext, message: dict):
 
             agent_appid = agent_entry["agent"]["appid"]
 
-            result = yield ctx.call_child_workflow(
-                workflow="agent_workflow",
-                input={"task": instruction},
-                app_id=agent_appid,
-                retry_policy=self._retry_policy,
+            _agent_tool = agent_to_tool(
```
CasperGN, line 795: Wait, this means orchestration will also invoke agents as tools?

schemas/agent-metadata/latest.json
```diff
@@ -19,6 +19,12 @@
           "title": "Orchestrator",
           "type": "boolean"
         },
+        "is_tool": {
```
CasperGN: Assuming we'll roll this into `0.13.1` (since the `0.13.0` schema already exists in main), can you generate with the `--version 0.13.1` flag?

dapr_agents/agents/durable.py
sicoyle: I made it to where orchestrators can still use regular non-agent tools but not agents as tools :) Since orchestrators use child workflows to trigger subagents then they still and only communicate out to the other agents via their orchestrator strategy, notttt via the tool executor that normal agents can use..

CasperGN:
@sicoyle this lgtm. Can you generate the v0.13.1 schema and push up then it's good to go! 🚀

CasperGN (approved):
LGTM

PR #469, opened 2026-03-06
Feat/hot reload otel vars

CasperGN:
# Description

Implement hot-reloading of otel configuration & add example for all 4 different mechanisms of otel instrumentation & configuring.

## Issue reference

Please reference the issue this PR closes: N/A

## Checklist

* [x] Created/updated tests

### DIFF
--- a/dapr_agents/agents/base.py
+++ b/dapr_agents/agents/base.py
@@ -1,3 +1,16 @@
+#
+# Copyright 2026 The Dapr Authors
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
 from __future__ import annotations
 
 import asyncio
--- a/dapr_agents/agents/configs.py
+++ b/dapr_agents/agents/configs.py
@@ -1,3 +1,16 @@
+#
+# Copyright 2026 The Dapr Authors
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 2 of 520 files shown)


dapr_agents/agents/base.py
```diff
@@ -1680,103 +1754,189 @@ def _setup_agent_runtime_configuration(self) -> None:
...
 
-            _endpoint = config.endpoint or ""
+        otlp_headers: dict[str, str] = dict(config.headers or {})
 
-            logger_provider = None
-            if config.logging_enabled:
-                logger_provider = LoggerProvider(resource=resource)
+        otel_token = config.auth_token
+        if otel_token and otel_token != "":
+            otlp_headers["authorization"] = f"Bearer {otel_token}"
```
sicoyle, line 1806: do we support case sensitivity here?

dapr_agents/agents/configs.py
```diff
@@ -236,6 +237,30 @@ def validate_tool_choice(v: str) -> str:
...
+            f"Valid options: {[e.value for e in AgentTracingExporter]}"
+        )
+    return v
+
+
+def validate_otel_exporter_logging(v: str) -> str:
+    """Validate that the logging exporter is a known AgentLoggingExporter value."""
+    try:
+        AgentLoggingExporter(v)
+    except (ValueError, KeyError):
```
sicoyle, line 269: would we get any other exception type here that we need to raise?

examples/demo-otel-k8s/seed-runtime-store.sh
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env bash
```
sicoyle, line 1: can you add a comment somewhere even in the readme for this one pls that th ex wont work out of the box for windows?

dapr_agents/agents/configs.py
CasperGN: It should only be those 2 as any wrong input will raise, most probably `ValueError`

dapr_agents/agents/base.py
CasperGN: It shouldn't matter for the header as http headers are case insensitive so we set `authorization: Bearer {whatever was in the var}`

CasperGN:
@sicoyle pushed up the note about windows, added comment + resolved the other 2 ✅

sicoyle (approved):
thank you!!

PR #483, opened 2026-03-06
fix: handle summarize bug + concurrent registrations

sicoyle:
# Description

This PR fixes the concurrent write issue for agent registration by removing  the ensure_registry_initialized func as our registration function itself already handles cases where the key doesn't exist as well as when the key exists and we need an update.

It also fixes this one from Alice
```
Activity task #7 failed: LLM summarize failed: Extraction failed: No tool_calls found for function_call mode
```
If the LLM returns plain JSON, then we should support JSON here too.

## Issue reference

## Checklist

* [ ] Created/updated tests


sicoyle:
ignore me ^ just seeing where it's at again 🙃

PR #513, opened 2026-03-11
Fix/bug long lived dapr chat client

CasperGN:
# Description

_Please explain the changes you've made_

## Issue reference

## Checklist

* [ ] Created/updated tests

### DIFF
--- a/dapr_agents/agents/durable.py
+++ b/dapr_agents/agents/durable.py
@@ -1376,10 +1376,7 @@ class DurableAgent(AgentBase):
                 new_status = u["status"]
 
                 logger.debug(
-                    "Updating step %s/%s to status '%s'",
-                    step_id,
-                    sub_id,
-                    new_status,
+                    f"Updating step {step_id}/{sub_id} to status '{new_status}'"
                 )
                 target = find_step_in_plan(plan, step_id, sub_id)
                 if not target:


dapr_agents/agents/durable.py
```diff
@@ -966,6 +966,8 @@ def orchestration_workflow(self, ctx: wf.DaprWorkflowContext, message: dict):
                     f"Agent '{next_agent}' has invalid agent metadata. "
                     f"Available agents: {list(agents_metadata.keys())}"
                 )
+            agent_appid = agent_entry["agent"]["appid"]
+            agent_type = agent_entry["agent"].get("type", "agents").lower()
```
sicoyle, line 970: can you make type -> framework pls? I would hate for down stream consumers of this registry data to think this is an actual object type they can work with, and "type" is a keyword kinda in python so would be nice to avoid it pls 🙏

dapr_agents/agents/durable.py
```diff
@@ -1380,9 +1383,12 @@ async def update_plan_internal(
                 )
                 target = find_step_in_plan(plan, step_id, sub_id)
                 if not target:
-                    msg = f"Step {step_id}/{sub_id} not present in plan."
-                    logger.error(msg)
-                    raise ValueError(msg)
+                    logger.warning(
+                        "Step %s/%s not present in plan — skipping status update.",
```
sicoyle, line 1384: f strings pls

dapr_agents/agents/durable.py
CasperGN: This is the actual type though, like you'd get from `type(agent)`

PR #526, opened 2026-03-16
feat: add agent.workflow start method

sicoyle:
# Description

Users might want a workflow only means of starting an agent to not expose any ports or use a pubsub. This PR creates `runner.workflow(agent)` so agents used with this start method start the workflow runtime without any pubsub wiring or HTTP routes, and makes it only accessible via the workflow API or from other workflows.

I've made `trigger_agent` a standalone blocking helper that hides all of the WorkflowRuntime lifecycling and enables a fire-and-wait way to trigger agents without any workflow boilerplate code. Alternatively someone can use `call_agent` inside of a workflow as a yieldable helper similar to naming conventions/usage of `call_activity` that delegates to `call_child_workflow` under the hood. Both of these helpers use the registry to resolve fields without needing them explicitly and handle timeouts, warning if expired clientside but workflow still is running.

I also moved the agent name sanitization helper to a better helper location.

## Issue reference

## Checklist

* [ ] Created/updated tests

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1602,24 +1602,26 @@ wheels = [
 
 [[package]]
 name = "hf-xet"
-version = "1.2.0"
+version = "1.4.2"
 source = { registry = "https://pypi.org/simple" }
-sdist = { url = "https://files.pythonhosted.org/packages/5e/6e/0f11bacf08a67f7fb5ee09740f2ca54163863b07b70d579356e9222ce5d8/hf_xet-1.2.0.tar.gz", hash = "sha256:a8c27070ca547293b6890c4bf389f713f80e8c478631432962bb7f4bc0bd7d7f", size = 506020, upload-time = "2025-10-24T19:04:32.129Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/09/08/23c84a26716382c89151b5b447b4beb19e3345f3a93d3b73009a71a57ad3/hf_xet-1.4.2.tar.gz", hash = "sha256:b7457b6b482d9e0743bd116363239b1fa904a5e65deede350fbc0c4ea67c71ea", size = 672357, upload-time = "2026-03-13T06:58:51.077Z" }
 wheels = [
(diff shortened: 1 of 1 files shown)


dapr_agents/workflow/utils/core.py
```diff
@@ -43,6 +46,193 @@ def _set_stop(*_: object) -> None:
...
+        "workflow": workflow_name,
+        "input": input,
+        "app_id": app_id,
+    }
+    if instance_id is not None:
+        kwargs["instance_id"] = instance_id
+    return ctx.call_child_workflow(**kwargs)
+
+
+def trigger_agent(
```
CasperGN, line 154: Would it make sense to also have a util func to create an agents workflow name to be called?
I'm thinking the util being callable when you don't know the workflow name but you know either the agent name or app id?

dapr_agents/workflow/utils/names.py
```diff
@@ -0,0 +1,34 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import re
+
+
+def sanitize_agent_name(name: str) -> str:
```
CasperGN, line 17: Perhaps that's what we're getting here?

dapr_agents/workflow/utils/core.py
sicoyle: This should be captured by `agent_workflow_id` in `tool/workflow/agent_tool.py`. I am happy to rename this function if you have feedback or want things moved around :)

dapr_agents/workflow/utils/names.py
sicoyle: this one is the lower level func used within `agent_workflow_id` in `tool/workflow/agent_tool.py`

PR #537, opened 2026-03-19
fix: durable agent concurrency, broadcast memory, and pub/sub dedup

sicoyle:
# Description

1. OpenAI HTTP 400 "tool call id out of order" fixes:

Replace shared _last_etag with per-key _etag_cache: Dict[str, Optional[str]] in AgentComponents — eliminates cross-contamination between concurrent activity threads reading/writing state
Add entry= parameter to save_state so each activity serializes its own locally-held state entry, not the shared _state_model
Remove duplicate workflow_meta.append in agent_workflow that caused zip(workflow_meta, results) misalignment when ≥2 AgentWorkflowTool calls ran — tool responses were assigned to wrong tool_call_ids

2. Broadcast / team context sharing:

Rename broadcast_workflow → on_broadcast; add record_broadcast activity that persists received peer broadcasts to memory under a stable "broadcast" session key ({agent_name}:_memory_broadcast)
Inject last 3 broadcasts from memory into _reconstruct_conversation_history so workers have peer context on every LLM call
Move broadcast_to_team in orchestration_workflow from initial plan creation to post-loop, so workers receive final plan state (with completed step statuses) rather than an all-not_started snapshot
Restrict broadcast_to_team to orchestrators only in agent_workflow
Include broadcast memory slot in purge() alongside per-instance memory
Pub/sub duplicate workflow prevention

3. Add TTLDedupeBackend (60s TTL, 4096-entry LRU using cachetools) implementing the existing DedupeBackend protocol
Wire it as the default in AgentRunner._wire_pubsub_routes — Dapr at-least-once redeliveries within the TTL window are dropped before schedule_new_workflow is called

## Issue reference

## Checklist

* [ ] Created/updated tests

### DIFF
--- a/quickstarts/06_workflow_agents.py
+++ b/quickstarts/06_workflow_agents.py
@@ -60,7 +60,11 @@ if __name__ == "__main__":
     )
     print(f"Workflow started: {instance_id}", flush=True)
 
-    state = client.wait_for_workflow_completion(instance_id, timeout_in_seconds=60)
+    try:
+        state = client.wait_for_workflow_completion(instance_id, timeout_in_seconds=120)
+    except TimeoutError:
+        print(f"Workflow {instance_id} timed out waiting for completion.")
+        state = None
     if not state:
         print("No workflow state returned.")
     elif state.runtime_status.name == "COMPLETED":


dapr_agents/workflow/runners/agent.py
```diff
@@ -388,6 +389,7 @@ def _wire_pubsub_routes(
             await_timeout=await_timeout,
             fetch_payloads=fetch_payloads,
             log_outcome=log_outcome,
+            deduper=TTLDedupeBackend(),
         )
```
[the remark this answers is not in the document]
CasperGN: @sicoyle can you make sure cachetools is in direct dep group?

dapr_agents/agents/configs.py
```diff
@@ -321,12 +321,12 @@ class AgentPubSubConfig:
...
         pubsub_name: Name of the Dapr pub/sub component to use for all agent traffic.
-        agent_topic: Primary topic for direct messages to the agent. Defaults to ``name``.
+        topic: Primary topic for direct messages to the agent. Defaults to ``name``.
         broadcast_topic: Optional topic shared by a team for broadcast messages.
     """
 
     pubsub_name: str
-    agent_topic: Optional[str] = None
+    topic: Optional[str] = None
     broadcast_topic: Optional[str] = None
```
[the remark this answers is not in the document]
CasperGN: @sicoyle is this rename necessary?

dapr_agents/agents/base.py
```diff
@@ -1271,6 +1271,15 @@ def purge(self, workflow_instance_id: str) -> None:
                     workflow_instance_id,
                     exc_info=True,
                 )
+            try:
+                self.memory.purge_memory("broadcast")
```
CasperGN, line 1275: Can you expand a little on this?
Why would we need to purge broadcast memory?

dapr_agents/agents/base.py
```diff
@@ -1271,6 +1271,15 @@ def purge(self, workflow_instance_id: str) -> None:
...
+            except Exception:  # noqa: BLE001
+                logger.warning(
+                    "Failed to purge broadcast memory for agent %s; "
```
CasperGN, line 1278: f-string pls

dapr_agents/agents/base.py
```diff
@@ -1508,7 +1517,27 @@ def _reconstruct_conversation_history(
...
+                if broadcasts:
+                    recent = broadcasts[-3:]
+                    broadcast_msgs = [
+                        {"role": m.get("role", "user"), "content": m.get("content", "")}
+                        for m in recent
+                    ]
+                    messages = broadcast_msgs + messages
+            except Exception:  # noqa: BLE001
+                logger.warning(
+                    "Failed to load broadcast memory for agent %s; continuing without it.",
```
CasperGN, line 1535: f-string

dapr_agents/agents/durable.py
```diff
@@ -558,14 +558,6 @@ def agent_workflow(self, ctx: wf.DaprWorkflowContext, message: dict):
...
                                     )
                                 workflow_tasks.append(tool_obj(**call_kwargs))
-                                workflow_meta.append(
-                                    {
-                                        "order": idx,
-                                        "tool_call": tc,
-                                        "child_instance_id": child_instance_id,
-                                        "dispatch_time": ctx.current_utc_datetime.isoformat(),
-                                    }
-                                )
```
CasperGN, line 568: Wasn't this what we discussed was a good idea to preserve in the metadata?

dapr_agents/agents/durable.py
```diff
@@ -1191,6 +1165,25 @@ def orchestration_workflow(self, ctx: wf.DaprWorkflowContext, message: dict):
...
+                    "content": json.dumps({"objects": plan}, indent=2),
+                }
+            else:
+                broadcast_msg = final_message
+            yield ctx.call_activity(
+                self.broadcast_to_team,
+                input={"message": broadcast_msg},
+                retry_policy=self._retry_policy,
+            )
+
```
CasperGN, line 1180: This will make it easier to implement the team orchestration mode. Nice ✅

dapr_agents/agents/durable.py
```diff
@@ -1530,7 +1531,36 @@ def record_initial_entry(
...
+        metadata = message.get("_message_metadata", {}) or {}
+        source = metadata.get("source") or "unknown"
+        content = message.get("content", "")
+        if self.memory is not None:
+            self.memory.add_message(
+                UserMessage(content=f"[Broadcast from {source}]: {content}"),
+                session_id,
+            )
+        logger.info(
+            "Agent %s recorded broadcast from %s in memory (session=%s)",
```
CasperGN, line 1553: f-string

dapr_agents/workflow/utils/subscription.py
```diff
@@ -57,6 +57,29 @@ def seen(self, key: str) -> bool: ...
...
+class TTLDedupeBackend:
+    """Thread-safe in-memory deduplication backend using a TTL cache.
+
+    Entries expire after `ttl` seconds, so memory is bounded and old IDs
+    are not retained indefinitely.  Suitable for Dapr at-least-once delivery
+    where duplicate messages typically arrive within a short retry window.
+    """
+
+    def __init__(self, maxsize: int = 4096, ttl: float = 60.0) -> None:
+        from cachetools import TTLCache
```
CasperGN: move imports to top and make sure cachetools is moved from `dev` group

dapr_agents/agents/base.py
sicoyle: this would be an ever growing key otherwise... but i can rm the purge here if you have another idea for this?

dapr_agents/agents/durable.py
sicoyle: there were two of these so I removed the duplicate one bc it was causing issues.

dapr_agents/agents/configs.py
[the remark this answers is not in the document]
sicoyle: I was looking at the examples and actually did like renaming personally when I was checking the 04 example for clarity when using both topics (one to trigger and the other to listen with for updates), but sure I can go back to what it was called before :)

dapr_agents/agents/durable.py
sicoyle: I'm not entirely sure if we need a team orchestator mode tbh as if they set the broadcast topic with any orchestrator mode then it will broadcast to whoever is listening so it unlocks new potential for all of the orchestrators this way :)

dapr_agents/agents/base.py
CasperGN: Hm. Not at the moment. Maybe something to keep in the back of our mind for later

sicoyle:
i updated the metadata to revert my rename on the agent topic field so the metadata comment above is outdated

PR #571, opened 2026-04-17
feat: add human-in-the-loop support

### LINKED ISSUE #132: Callback Hooks for Dapr Agents
**Context**
Dapr agents currently lack a built-in mechanism to intercept, observe, or modify execution flow at critical points, such as before/after LLM or tool calls. Introducing a standardized callback interface enhances extensibility, debugging, and control, enabling advanced use cases like auditing, response caching, guardrails, and dynamic modification of inputs/outputs.

**Acceptance Criteria**

- Support four callback types:
- before_llm_call(context): invoked before calling the LLM.
- after_llm_call(context, response): invoked after receiving LLM response.
- before_tool_call(context): invoked before invoking a tool.
- after_tool_call(context, result): invoked after tool execution.
- Each callback receives a context object with execution metadata (e.g., agent name, session ID, invocation details).

**Callbacks can:**

- Log, inspect, or modify request data.
- Return None to proceed or return a specific value to override execution.

**Callback return behavior:**

- Returning a result from a before_* callback skips execution and uses that value.
- Returning a result from an after_* callback replaces the original output.

### LINKED ISSUE #162: Add Human-in-the-Loop (HITL) Support to DurableAgent
**Description**:
Introduce a pattern in `DurableAgent` to allow human approval at critical decision points. The agent should pause and wait for external input before proceeding. This is essential for high-stakes workflows requiring manual oversight. Behind the scene, this will pause workflow execution and wait for an external event.

**When to use:**

* Human oversight is legally, operationally, or ethically required (e.g., financial approvals, production deployments)
* Pausing, resuming, or aborting tasks based on external input is critical
* You need to audit decision points and responses

**Example: Human Approval Loop**

> This is a conceptual example to illustrate intent. Implementation may vary.

```python
@tool(wait_for="approval_received")
async def dangerous_task(approval_event: ExternalEvent) -> str:
    if approval_event.response.strip().lower() == "yes":
        return "Dangerous task completed successfully."
    else:
        return "Dangerous task aborted."

durable_agent = DurableAgent(
    name="HumanApprovalBot",
    role="Human-in-the-loop Operations Coordinator",
    instructions=[
        "Handle critical workflows that require explicit human approval",
        "Pause and wait for human input before continuing risky operations",
        "Retain all state, inputs, and decisions across restarts or delays",
    ],
    llm=DaprChatClient(conversation_component="openai"),
    tools=[dangerous_task, mcp_tool_with_human_approval],
    memory=ConversationDaprStateMemory(store_name="statestore"),
    durable_execution=DaprWorkflowDurableExecution(
        state_store_name="statestore",
        state_key="human_approval_workflow",
    ),
)
```

**Event Loop Example**

```python
async for event in durable_agent.ctx.stream_events():
    if isinstance(event, ExternalEventRequired):
        response = input(event.prefix)
        durable_agent.ctx.raise_event(
            instance_id=event.instance_id,
            event_name="approval_received",
            event_data=asdict(response),
        )

response = durable_agent.run_async(user_msg="Proceed with this dangerous task.")
```

**Open Questions**

1. How should users define when human-in-the-loop is triggered: is it tool-only, or pre-LLM, post-LLM calls?
2. What is the best way to notify or signal the human? Can the agent emit events during specific lifecycle stages as described in #132 ?

**There are different ways of doing HITL with agents**

* [https://www.humanlayer.dev/docs/core/human-as-tool](https://www.humanlayer.dev/docs/core/human-as-tool)
* [https://docs.llamaindex.ai/en/stable/understanding/agent/human_in_the_loop/](https://docs.llamaindex.ai/en/stable/understanding/agent/human_in_the_loop/)
* [https://google.github.io/adk-docs/agents/multi-agents/#human-in-the-loop-pattern](https://google.github.io/adk-docs/agents/multi-agents/#human-in-the-loop-pattern)
* [https://github.com/google/adk-python/blob/main/contributing/samples/human_in_loop/main.py](https://github.com/google/adk-python/blob/main/contributing/samples/human_in_loop/main.py)
* [https://github.com/langchain-ai/agents-from-scratch/blob/main/notebooks/hitl.ipynb](https://github.com/langchain-ai/agents-from-scratch/blob/main/notebooks/hitl.ipynb)

1Ninad:
# Description
Adds human-in-the-loop support to DurableAgent. Tools marked `@tool(requires_approval=True)` pause the Dapr workflow, publish an approval request, and only run after a human sends back a decision.

## Issue reference
closes: #162

## Checklist
* [x] Created/updated tests

### DIFF
--- a/dapr_agents/agents/configs.py
+++ b/dapr_agents/agents/configs.py
@@ -400,6 +400,29 @@ class OrchestrationMode(StrEnum):
 
 
 @dataclass
+class AgentApprovalConfig:
+    """
+    Configuration for human-in-the-loop approval before tool execution.
+
+    When enabled, any tool decorated with requires_approval=True will pause the
+    workflow and wait for a human decision before running. Tools without that flag
+    are unaffected and execute immediately as usual.
+
+    Attributes:
+        enabled: When False (the default) approval checks are skipped entirely.
+        pubsub_name: Dapr pub/sub component where ApprovalRequiredEvent is published.
+        topic: Topic name for outbound approval requests.
+        default_timeout_seconds: Seconds to wait before auto-denying when a tool does
+            not specify its own approval_timeout_seconds.
+    """
+
+    enabled: bool = False
+    pubsub_name: str = "dapr-agents-pubsub"
+    topic: str = "agent-approval-requests"
+    default_timeout_seconds: int = 300
+
+
+@dataclass
 class AgentExecutionConfig:
     """
     Dials to configure the agent execution.
@@ -411,6 +434,7 @@ class AgentExecutionConfig:
     tool_choice: Optional[str] = "auto"
     tool_execution_mode: ToolExecutionMode = ToolExecutionMode.PARALLEL
     orchestration_mode: Optional[OrchestrationMode] = None
+    approval: AgentApprovalConfig = field(default_factory=AgentApprovalConfig)
 
 
 @dataclass
--- a/dapr_agents/agents/configs.py
+++ b/dapr_agents/agents/configs.py
@@ -404,18 +404,28 @@ class AgentApprovalConfig:
     """
     Infrastructure configuration for human-in-the-loop approval.
 
-    This tells the agent where to publish ApprovalRequiredEvent messages when a
+    This tells the agent how to deliver ApprovalRequiredEvent messages when a
     hook returns RequireApproval. The gate for whether approval runs is the hook
     itself — if no hook returns RequireApproval, this config is never used.
 
+    Delivery modes:
+        - pubsub_name set: publishes the event to the configured Dapr pub/sub topic.
+          Use this when the agent is running in subscribe() or serve() mode and a
+          pub/sub component is available (e.g. a Slack bot or dashboard is listening).
+        - pubsub_name None (default): no pub/sub publish. The event is held in memory
+          and exposed via GET /hitl/approvals when the agent is running in serve() mode.
+          For workflow-only agents, submit responses directly via the Dapr sidecar:
+          POST <sidecar>/v1.0-beta1/workflows/dapr/{instance_id}/raiseEvent/approval_response_{id}
+
     Attributes:
-        pubsub_name: Dapr pub/sub component where ApprovalRequiredEvent is published.
-        topic: Topic name for outbound approval requests.
+        pubsub_name: Optional Dapr pub/sub component for outbound approval events.
+            Set to None (default) to disable pub/sub delivery and use HTTP polling instead.
+        topic: Topic name used when pubsub_name is set.
         default_timeout_seconds: Seconds to wait before auto-denying when a
             RequireApproval decision does not specify its own timeout_seconds.
     """
 
-    pubsub_name: str = "dapr-agents-pubsub"
+    pubsub_name: Optional[str] = None
     topic: str = "agent-approval-requests"
     # None means wait indefinitely — the workflow suspends until a human responds, with no automatic denial. Use a positive int (seconds) to auto-deny after that window elapses when approvers may be unavailable.
     default_timeout_seconds: Optional[int] = 300
--- a/dapr_agents/agents/configs.py
+++ b/dapr_agents/agents/configs.py
@@ -402,21 +402,19 @@ class OrchestrationMode(StrEnum):
 @dataclass
 class AgentApprovalConfig:
     """
-    Configuration for human-in-the-loop approval before tool execution.
+    Infrastructure configuration for human-in-the-loop approval.
 
-    When enabled, any tool decorated with requires_approval=True will pause the
-    workflow and wait for a human decision before running. Tools without that flag
-    are unaffected and execute immediately as usual.
+    This tells the agent where to publish ApprovalRequiredEvent messages when a
+    hook returns RequireApproval. The gate for whether approval runs is the hook
+    itself — if no hook returns RequireApproval, this config is never used.
 
     Attributes:
-        enabled: When False (the default) approval checks are skipped entirely.
         pubsub_name: Dapr pub/sub component where ApprovalRequiredEvent is published.
         topic: Topic name for outbound approval requests.
-        default_timeout_seconds: Seconds to wait before auto-denying when a tool does
-            not specify its own approval_timeout_seconds.
+        default_timeout_seconds: Seconds to wait before auto-denying when a
+            RequireApproval decision does not specify its own timeout_seconds.
     """
 
-    enabled: bool = False
     pubsub_name: str = "dapr-agents-pubsub"
     topic: str = "agent-approval-requests"
     default_timeout_seconds: int = 300
--- a/dapr_agents/agents/configs.py
+++ b/dapr_agents/agents/configs.py
@@ -417,7 +417,8 @@ class AgentApprovalConfig:
 
     pubsub_name: str = "dapr-agents-pubsub"
     topic: str = "agent-approval-requests"
-    default_timeout_seconds: int = 300
+    # None means wait indefinitely — the workflow suspends until a human responds, with no automatic denial. Use a positive int (seconds) to auto-deny after that window elapses when approvers may be unavailable.
+    default_timeout_seconds: Optional[int] = 300
 
 
 @dataclass
--- a/dapr_agents/agents/durable.py
+++ b/dapr_agents/agents/durable.py
@@ -53,6 +53,7 @@ from dapr_agents.agents.base import AgentBase
 from dapr_agents.agents.configs import (
     OrchestrationMode,
     ToolExecutionMode,
+    AgentApprovalConfig,
     AgentExecutionConfig,
     AgentMemoryConfig,
     AgentPubSubConfig,
@@ -66,6 +67,8 @@ from dapr_agents.agents.configs import (
 from dapr_agents.agents.prompting import AgentProfileConfig
 from dapr_agents.agents.schemas import (
     AgentWorkflowMessage,
+    ApprovalRequiredEvent,
+    ApprovalResponseEvent,
     BroadcastMessage,
     TriggerAction,
 )
@@ -507,6 +510,35 @@ class DurableAgent(AgentBase):
                                 turn,
                             )
 
+                        # approval pass:
+                        # for each tool that requires approval, pause the workflow
+                        # and collect the human decision before we build any task lists.
+                        # approvals are processed one at a time because each blocks
+                        # until a human responds (or the timeout elapses).
+                        approval_decisions: Dict[str, bool] = {}
+                        if self.execution.approval.enabled:
+                            for tc in tool_calls:
+                                fn_name_check = tc["function"]["name"]
+                                check_obj = self.tool_executor.get_tool(fn_name_check)
+                                if check_obj and getattr(
+                                    check_obj, "requires_approval", False
+                                ):
+                                    approved = yield from self._request_approval(
+                                        ctx, ctx.instance_id, tc
+                                    )
+                                    approval_decisions[tc["id"]] = approved
+                                    if not ctx.is_replaying:
+                                        logger.debug(
+                                            "Approval for tool '%s': %s (instance=%s)",
+                                            fn_name_check,
+                                            "approved" if approved else "not approved",
+                                            ctx.instance_id,
+                                        )
+
+                        ordered: List[Optional[Dict[str, Any]]] = [None] * len(
+                            tool_calls
+                        )
+
                         workflow_tasks: List[Any] = []
                         workflow_meta: List[Dict[str, Any]] = []
                         activity_tasks: List[Any] = []
@@ -514,6 +546,31 @@ class DurableAgent(AgentBase):
 
                         for idx, tc in enumerate(tool_calls):
                             fn_name = tc["function"]["name"]
+
+                            # inject denial message and skip execution for any tool
+                            # whose approval was not granted or timed out
+                            if (
+                                tc["id"] in approval_decisions
+                                and not approval_decisions[tc["id"]]
+                            ):
+                                denial_msg = ToolMessage(
+                                    content=(
+                                        f"Tool '{fn_name}' was not executed: "
+                                        "approval was not granted."
+                                    ),
+                                    role="tool",
+                                    name=fn_name,
+                                    tool_call_id=tc["id"],
+                                )
+                                ordered[idx] = denial_msg.model_dump()
+                                if not ctx.is_replaying:
+                                    logger.info(
+                                        "Skipping tool '%s': approval was not granted (instance=%s)",
+                                        fn_name,
+                                        ctx.instance_id,
+                                    )
+                                continue
+
                             tool_obj = self.tool_executor.get_tool(fn_name)
 
                             # WorkflowContextInjectedTool instances get executed inline as workflow tasks so they can receive the workflow context.
@@ -591,10 +648,6 @@ class DurableAgent(AgentBase):
                         else:
                             results: List[Any] = yield wf.when_all(all_tasks)
 
-                        ordered: List[Optional[Dict[str, Any]]] = [None] * len(
-                            tool_calls
-                        )
-
                         for meta, res in zip(
                             workflow_meta, results[: len(workflow_tasks)]
                         ):
@@ -634,6 +687,18 @@ class DurableAgent(AgentBase):
                                 for meta in activity_meta
                             }
                         )
+                        # include denied tool calls so save_tool_results can log them
+                        # in tool_history for observability
+                        for tc in tool_calls:
+                            if (
+                                tc["id"] in approval_decisions
+                                and not approval_decisions[tc["id"]]
+                            ):
+                                tool_calls_by_id[tc["id"]] = {
+                                    "tool_call": tc,
+                                    "is_agent_call": False,
+                                    "dispatch_time": ctx.current_utc_datetime.isoformat(),
+                                }
                         yield ctx.call_activity(
                             self.save_tool_results,
                             input={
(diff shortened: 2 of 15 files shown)


1Ninad:
Hi @bibryam, could you please review the PR when you have time?

bibryam:
That next step is for @sicoyle & @CasperGN

CasperGN (changes requested):
Thank you @1Ninad for taking this on. I've been in design-land with my trusted advisor (Claude rn) on how we could expand the HITL setup with some of the upcoming features we're planning and it's gotta be a hook-like system. I've tried to get the design session I had + the context of this PR condensed down to the following feedback, let me know if the below makes sense!

---

## What this gets right

- `ApprovalRequiredEvent` / `ApprovalResponseEvent` are well-shaped.
- `publish_approval_request` wrapped as an activity is the right pattern for replay determinism.
- Deterministic `uuid5(NAMESPACE_DNS, f"{instance_id}:{tool_call_id}")` for the request ID is correct.
- `when_any([wait_for_external_event, create_timer])` is the right timeout primitive.
- Serial approvals + parallel execution of approved tools strikes a good balance.
- `approval_requests` dedup map on `AgentWorkflowEntry` — good foresight for replay dedup.

## The gap

The only way to mark a step for approval is `@tool(requires_approval=True)`. This misses the primary use case: **MCP tools, OpenAPI tools, and other runtime-loaded tools do not go through the `@tool` decorator**, so a deployer who wants to gate `delete_repository` from a GitHub MCP server has no way to do so.

Concrete examples that cannot be expressed with this PR:

- "Require approval for any tool from `mcp:github` whose name starts with `delete_`"
- "Require approval when `send_email`'s `to` argument ends with `@external.com`"
- "Deny `drop_table` outright without involving a human"
- "Require approval for LLM calls above a certain estimated cost"

These all need a **runtime policy** that inspects the step at dispatch time, not a decorator applied (or not) at tool-definition time.

## Suggested direction: unify with #132

The companion issue [#132](https://github.com/dapr/dapr-agents/issues/132) ("Callback Hooks for Dapr Agents") was closed stale but specified exactly the interface HITL needs:

> - `before_tool_call(context)` / `after_tool_call(context, result)`
> - `before_llm_call(context)` / `after_llm_call(context, response)`
> - Returning a value from `before_*` skips execution and uses that value
> - Returning a value from `after_*` replaces the original output

HITL is literally the case where `before_tool_call` says *"skip execution, pause, and wait for a human."* OpenAI's Agents SDK takes the same shape — tool approvals halt the run with `interruptions[]` and a resumable `state`, and input/output guardrails share the same inspect-and-decide contract ([guardrails & approvals docs](https://developers.openai.com/api/docs/guides/agents/guardrails-approvals)). OpenAI Codex's `PreToolUse` / `PostToolUse` hooks are structurally identical ([codex hooks](https://developers.openai.com/codex/hooks)).

### Proposed unified surface

```python
from dapr_agents.hooks import Hooks, HookContext, HookDecision
from dapr_agents.hooks import Proceed, Skip, Modify, RequireApproval, Deny

def before_tool(ctx: HookContext) -> HookDecision:
    # Works for decorated tools, MCP, OpenAPI, anything — inspection at runtime
    if ctx.source == "mcp" and ctx.step_name.startswith("delete_"):
        return RequireApproval(
            timeout_seconds=3600,
            instructions=f"Confirm deletion of {ctx.payload.get('path')}",
        )
    if ctx.step_name == "drop_table":
        return Deny(reason="schema changes go through DBA review")
    return Proceed()   # `None` also coerces to this

agent = DurableAgent(
    ...,
    hooks=Hooks(
        before_tool_call=before_tool,
        after_llm_call=redact_pii,
        # before_llm_call, after_tool_call also available
    ),
)
```

### Decision variants

| Decision | Semantics |
|---|---|
| `Proceed()` | Run the step normally (default; `None` coerces to this) |
| `Skip(result=...)` | Skip execution, use `result` instead. Covers caching, short-circuit, guardrail trips |
| `Modify(payload=...)` / `Modify(result=...)` | Run the step with mutated input (`before_*`) or replace output (`after_*`) |
| `RequireApproval(timeout_seconds, instructions, reason, approvers, on_denial)` | Pause the workflow, publish an `ApprovalRequest`, wait for a human (HITL) |
| `Deny(reason=...)` | Block without involving a human; synthesize a denial message |

HITL becomes **one decision among five**. This PR's plumbing becomes the backend that fires when a hook returns `RequireApproval`.

### Capability matrix

| Use case | Hook | Decision |
|---|---|---|
| HITL on specific tools (AI-461) | `before_tool_call` | `RequireApproval(...)` |
| HITL on MCP tools | `before_tool_call` | same — hook inspects `ctx.source == "mcp"` at runtime |
| HITL on LLM calls | `before_llm_call` | `RequireApproval(...)` |
| Response caching | `before_llm_call` | `Skip(result=cached)` |
| Output redaction / guardrails | `after_llm_call` | `Modify(result=redacted)` |
| Prompt injection check | `before_tool_call` | `Skip(result=safe_default)` or `Deny(...)` |
| Cost ceiling | `before_llm_call` | `Deny(reason="budget exceeded")` |
| Audit logging | any hook | log + `Proceed()` |

## What to salvage from this PR

Most of the implementation stays — only the opt-in surface changes.

**Keep:**
- `ApprovalRequiredEvent` / `ApprovalResponseEvent` (generalize `tool_name` → `step_name`, add `step_kind`, `source`)
- `publish_approval_request` activity (generalize payload shape)
- Deterministic `uuid5` request ID
- `when_any` timer race
- `raise_approval_event` helper (recommend moving off the instance to a module-level function so external approvers don't need a `DurableAgent` to send the response)
- Serial-approval + parallel-execution ordering
- `approval_requests` dedup map on `AgentWorkflowEntry`

**Drop:**
- `AgentTool.requires_approval`
- `@tool(requires_approval=True, approval_timeout_seconds=...)` decorator extension
- `AgentApprovalConfig.enabled` — the gate becomes "did a registered hook return `RequireApproval`?"

**Replace:**
- The approval-specific branch in `agent_workflow` (`durable.py:510`) with a general `_run_with_hooks(ctx, step_kind, payload, activity_call)` helper used at both the `call_llm` site and each tool call site.

## Determinism note

Hook callbacks are invoked during workflow replay, so they must be pure functions of their `HookContext`. Any I/O inside a hook breaks replay — users who need I/O-backed policies (DB lookup, external HTTP) should either:

1. Return `RequireApproval` and put the I/O in the approval service, or
2. Wait for async hook support (v2 — hook returns a generator that can `yield` activities).

This should be called out in the module docstring and in the example.

## Tests

Current tests only cover Pydantic model construction and config defaults. Workflow-integration tests are needed for:

- Approve path → tool runs, `tool_history` records APPROVED
- Deny decision (no human) → synthesized `ToolMessage`, no `run_tool` call
- `RequireApproval` → approval event → tool runs
- `RequireApproval` → timer wins → denial `ToolMessage`, `timed_out=True`
- LLM-level denial → workflow terminates with synthesized assistant message
- Mixed parallel calls: approved + denied + require-approval → correct ordering in `save_tool_results`
- MCP tool (mocked) gated by hook → same behavior as local tool
- Replay determinism: the hook is called twice with identical inputs and produces identical decisions

The fixture at `tests/agents/durableagent/test_durable_agent.py:152` already mocks `wait_for_external_event` and `create_timer` — reuse that pattern.

1Ninad:
Thanks @CasperGN for the suggestions. I have implemented them for issue #162.

1Ninad:
Hi @CasperGN, let me know your feedback when you're available.

dapr_agents/agents/durable.py
```diff
@@ -726,6 +866,140 @@ def agent_workflow(self, ctx: wf.DaprWorkflowContext, message: dict):
...
+                approval_request_id,
+                approval_config.topic,
+                fn_name,
+                instance_id,
+            )
+
+        # suspend until the external approval event arrives or the timeout elapses;
+        # wait_for_external_event has no timeout parameter in this SDK version so we
+        # race the event against a create_timer using when_any instead.
+        from durabletask import task as dt
```
CasperGN: Can you please move imports to the top of the file

dapr_agents/agents/configs.py
```diff
@@ -399,6 +399,27 @@ class OrchestrationMode(StrEnum):
...
+    Attributes:
+        pubsub_name: Dapr pub/sub component where ApprovalRequiredEvent is published.
+        topic: Topic name for outbound approval requests.
+        default_timeout_seconds: Seconds to wait before auto-denying when a
+            RequireApproval decision does not specify its own timeout_seconds.
+    """
+
+    pubsub_name: str = "dapr-agents-pubsub"
+    topic: str = "agent-approval-requests"
+    default_timeout_seconds: int = 300
```
CasperGN: How does a user set infinite wait?

dapr_agents/agents/durable.py
```diff
@@ -1922,6 +2204,41 @@ def save_tool_results(
...
+        Args:
+            payload: Keys 'event' (serialized ApprovalRequiredEvent dict),
+                'pubsub_name' (pub/sub component name), 'topic' (topic name).
+        """
+        event_data = payload["event"]
+        pubsub_name = payload["pubsub_name"]
+        topic = payload["topic"]
+
+        async def _publish() -> None:
+            from dapr_agents.workflow.utils.pubsub import publish_message
```
CasperGN: Imports to the top please

dapr_agents/agents/durable.py
```diff
@@ -1922,6 +2204,41 @@ def save_tool_results(
...
+
+            await publish_message(
+                pubsub_name=pubsub_name,
+                topic_name=topic,
+                message=event_data,
+            )
+
+        self._run_asyncio_task(_publish())
+        logger.info(
+            "Published approval request %s for tool '%s' to topic '%s'",
```
CasperGN: Can you convert this to f-string

dapr_agents/agents/durable.py
```diff
@@ -2293,6 +2610,53 @@ def save_plan(
...
+        Args:
+            instance_id: Workflow instance ID from the ApprovalRequiredEvent.
+            approval_request_id: Unique request ID from the ApprovalRequiredEvent.
+            approved: True to allow the tool to run, False to block it.
+            reason: Optional human-provided explanation for the decision.
+
+        Raises:
+            Exception: If the Dapr workflow client cannot deliver the event.
+        """
+        from dapr.ext.workflow import DaprWorkflowClient
```
CasperGN: Import to the top

CasperGN (changes requested):
Should we also create an endpoint if serving with http?
Similarly, what if the agent is running as workflow only?

I think we should be supporting the different ways for serving long running agents without requiring pub/sub

1Ninad:
> Should we also create an endpoint if serving with http? Similarly, what if the agent is running as workflow only?
> 
> I think we should be supporting the different ways for serving long running agents without requiring pub/sub

Agreed. Would you be okay if we address these in a follow-up PR to keep this one focused on the core HITL mechanism?

dapr_agents/agents/configs.py
1Ninad: Updated. 
Changed `default_timeout_seconds` to `Optional[int]`, where None means wait indefinitely with no auto-deny. Updated` _request_approval` to skip the timer entirely when `None`, so the workflow just suspends until a human explicitly responds

CasperGN:
> > Should we also create an endpoint if serving with http? Similarly, what if the agent is running as workflow only?
> > I think we should be supporting the different ways for serving long running agents without requiring pub/sub
> 
> Agreed. Would you be okay if we address these in a follow-up PR to keep this one focused on the core HITL mechanism?

I think it's better to include in this one tbh. I'd rather we come out with the HITL being feature complete on the different options for serving the agent

1Ninad:
> I think it's better to include in this one tbh. I'd rather we come out with the HITL being feature complete on the different options for serving the agent

Done, I have implemented it. Please review when you're available.

dapr_agents/agents/durable.py
```diff
@@ -2327,6 +2666,80 @@ def save_plan(
...
+        wf_client.raise_workflow_event(
+            instance_id=instance_id,
+            event_name=event_name,
+            data=response.model_dump(mode="json"),
+        )
+
+        self._pending_approvals.pop(approval_request_id, None)
+
+        logger.info(
+            "Raised approval event '%s' for instance %s: %s",
```
CasperGN: f-string please

dapr_agents/agents/durable.py
```diff
@@ -1954,6 +2244,55 @@ def save_tool_results(
...
+            self._run_asyncio_task(_publish())
+            logger.info(
+                "Published approval request %s for step '%s' to topic '%s'",
+                approval_request_id,
+                event_data.get("step_name"),
+                topic,
+            )
+        else:
+            logger.info(
+                "Stored approval request %s for step '%s' (no pub/sub configured; poll GET /hitl/approvals or use Dapr sidecar raiseEvent API)",
```
CasperGN: f-string please

dapr_agents/agents/durable.py
```diff
@@ -1954,6 +2244,55 @@ def save_tool_results(
...
+            async def _publish() -> None:
+                await publish_message(
+                    pubsub_name=pubsub_name,
+                    topic_name=topic,
+                    message=event_data,
+                )
+
+            self._run_asyncio_task(_publish())
+            logger.info(
+                "Published approval request %s for step '%s' to topic '%s'",
```
CasperGN: f-string please

dapr_agents/agents/durable.py
```diff
@@ -726,6 +873,141 @@ def agent_workflow(self, ctx: wf.DaprWorkflowContext, message: dict):
...
+                logger.warning(
+                    "Could not parse approval response for request %s: %s — auto-denying",
+                    approval_request_id,
+                    exc,
+                )
+            return False
+
+        if not ctx.is_replaying:
+            logger.info(
+                "Approval decision for request %s, tool '%s': %s (instance=%s)",
```
CasperGN: f-string please

dapr_agents/agents/durable.py
```diff
@@ -726,6 +873,141 @@ def agent_workflow(self, ctx: wf.DaprWorkflowContext, message: dict):
...
+                return False
+
+        # event won the race — read the human decision
+        try:
+            response_data = event_task.get_result()
+            response = ApprovalResponseEvent(**response_data)
+        except Exception as exc:
+            if not ctx.is_replaying:
+                logger.warning(
+                    "Could not parse approval response for request %s: %s — auto-denying",
```
CasperGN: f-string please

dapr_agents/agents/durable.py
```diff
@@ -726,6 +873,141 @@ def agent_workflow(self, ctx: wf.DaprWorkflowContext, message: dict):
...
+            yield event_task
+        else:
+            # Race the approval event against a timer
+            timer_task = ctx.create_timer(timedelta(seconds=timeout_seconds))
+            winner = yield dt.when_any([event_task, timer_task])
+
+            if winner is timer_task:
+                if not ctx.is_replaying:
+                    logger.warning(
+                        "Approval request %s timed out for tool '%s' (instance=%s) — auto-denying",
```
CasperGN: f-string please

dapr_agents/agents/durable.py
```diff
@@ -726,6 +873,141 @@ def agent_workflow(self, ctx: wf.DaprWorkflowContext, message: dict):
...
+                "event": approval_event.model_dump(mode="json"),
+                "pubsub_name": approval_config.pubsub_name,
+                "topic": approval_config.topic,
+            },
+            retry_policy=self._retry_policy,
+        )
+
+        if not ctx.is_replaying:
+            logger.info(
+                "Approval request %s published to topic '%s' (tool='%s', instance=%s)",
```
CasperGN: f-string please

dapr_agents/agents/durable.py
```diff
@@ -726,6 +873,141 @@ def agent_workflow(self, ctx: wf.DaprWorkflowContext, message: dict):
...
+            source=getattr(tool_obj, "source", "local"),
+            tool_call_id=tool_call_id,
+            tool_arguments=tool_args,
+            timeout_seconds=timeout_seconds,
+            instructions=decision.instructions,
+        )
+
+        if not ctx.is_replaying:
+            logger.info(
+                "Requesting approval: tool='%s' approval_request_id=%s instance=%s timeout=%ds",
```
CasperGN: f-string please

dapr_agents/agents/durable.py
```diff
@@ -507,13 +529,126 @@ def agent_workflow(self, ctx: wf.DaprWorkflowContext, message: dict):
...
+                                skip_msg = ToolMessage(
+                                    content=skip_content,
+                                    role="tool",
+                                    name=fn_name,
+                                    tool_call_id=tc["id"],
+                                )
+                                ordered[idx] = skip_msg.model_dump()
+                                if not ctx.is_replaying:
+                                    logger.info(
+                                        "Skipping tool '%s' with hook-provided result (instance=%s)",
```
CasperGN: f-string please

dapr_agents/agents/durable.py
```diff
@@ -507,13 +529,126 @@ def agent_workflow(self, ctx: wf.DaprWorkflowContext, message: dict):
...
+                                denial_msg = ToolMessage(
+                                    content=f"Tool '{fn_name}' was not executed: {block_reason}.",
+                                    role="tool",
+                                    name=fn_name,
+                                    tool_call_id=tc["id"],
+                                )
+                                ordered[idx] = denial_msg.model_dump()
+                                if not ctx.is_replaying:
+                                    logger.info(
+                                        "Skipping tool '%s': %s (instance=%s)",
```
CasperGN: f-string please

dapr_agents/agents/durable.py
```diff
@@ -507,13 +529,126 @@ def agent_workflow(self, ctx: wf.DaprWorkflowContext, message: dict):
...
+                                    hook_decisions[tc["id"]] = (
+                                        Proceed()
+                                        if approved
+                                        else Deny(
+                                            reason="approval was not granted or timed out"
+                                        )
+                                    )
+                                    if not ctx.is_replaying:
+                                        logger.debug(
+                                            "RequireApproval for tool '%s': %s (instance=%s)",
```
CasperGN: f-string please

examples/02-durable-agent-tool-call/durable_agent_hitl.py
```diff
@@ -0,0 +1,254 @@
...
+        role="Operations Assistant",
+        goal="Help the user manage data and fetch information.",
+        instructions=[
+            "Use delete_old_data when asked to clean up or remove data.",
+            "Use get_weather for weather queries.",
+        ],
+        llm=OpenAIChatClient(
+            model="openai/gpt-5-nano",
+            base_url="https://openrouter.ai/api/v1",
+        ),
```
CasperGN: Can we instead use the component as the other examples do?

examples/02-durable-agent-tool-call/durable_agent_hitl.py
```diff
@@ -0,0 +1,254 @@
...
+from dapr_agents.llm.openai import OpenAIChatClient
+
+load_dotenv()
+
+logger = logging.getLogger(__name__)
+
+# both uvicorn and main() run in the same event loop (via asyncio.create_task),
+# so put_nowait() from the fastapi handler immediately wakes up get() in main()
+_approval_queue: asyncio.Queue = asyncio.Queue()
+_dapr_app = FastAPI()
```
CasperGN: Why do we need this when the agent uses `.run()`?

examples/02-durable-agent-tool-call/durable_agent_hitl.py
```diff
@@ -0,0 +1,254 @@
...
+    print("\n--- starting workflow ---\n")
+
+    instance_id = await runner.run(
+        agent,
+        payload={"task": prompt},
+        wait=False,
+    )
+    print(f"workflow started: instance_id = {instance_id}\n")
+
+    print("waiting for approval request from pub/sub (workflow is paused) ...\n")
```
CasperGN: We're not using approval from pub/sub in this example

CasperGN (changes requested):

dapr_agents/agents/durable.py
1Ninad: Got it, will switch to f-strings. Should I also add this to `agents.md` under Code Standards?

1Ninad:
Hi @CasperGN, pushed a commit addressing all comments. Apologies for the delayed turnaround.

dapr_agents/agents/durable.py
```diff
@@ -311,6 +328,11 @@ def __init__(
         self._runtime_owned = runtime is None
         self._registered = False
         self._started = False
+        self._hooks: Optional[Hooks] = hooks
+        # In-memory store of active approval requests, keyed by approval_request_id.
+        # Populated by publish_approval_request activity; consumed by raise_approval_event.
+        # Used by GET /hitl/approvals when serving over HTTP without pub/sub.
+        self._pending_approvals: Dict[str, Dict[str, Any]] = {}
```
CasperGN, line 332: Since this is only in-memory, how will the agent handle a crash with pending approvals?

dapr_agents/agents/durable.py
```diff
@@ -23,6 +23,9 @@
 from os import getenv
 from dapr_agents.tool.utils.function_calling import sanitize_openai_tool_name
 import dapr.ext.workflow as wf
+from dapr.ext.workflow import DaprWorkflowClient
+from dapr.ext.workflow.workflow_state import WorkflowStatus
+from durabletask import task as dt
```
CasperGN: Instead importing durable task directly and then using `dt.when_any..` can we instead use the workflow client?

dapr_agents/hooks.py
```diff
@@ -0,0 +1,145 @@
...
+        None
+    )
+    """called after a tool completes. return Modify(result=...) to replace the output."""
+
+    before_llm_call: Optional[Callable[[HookContext], Optional[HookDecision]]] = None
+    """called before every llm call."""
+
+    after_llm_call: Optional[Callable[[HookContext, Any], Optional[HookDecision]]] = (
+        None
+    )
```
CasperGN: Can we add something along the lines of
```python
BeforeHook = Callable[[HookContext], Optional[HookDecision]]
AfterHook = Callable[[HookContext, Any], Optional[HookDecision]]   
```
At the top of this file and then change these to:

```python
before_tool_call: List[BeforeHook] = field(default_factory=list)                                                                                          
after_tool_call: List[AfterHook] = field(default_factory=list)
before_llm_call: List[BeforeHook] = field(default_factory=list)                                                                                           
after_llm_call: List[AfterHook] = field(default_factory=list)
```

This allows us to reuse the hooks for plugins that can be chained. For this PR you can just use `for hook in self._hooks....` when evaluating. We'll pick up on order and what not in a follow up :-)

examples/02-durable-agent-tool-call/approval_sender.py
```diff
@@ -0,0 +1,78 @@
+#
```
CasperGN, line 1: Generally on these examples, can we simplify them a bit? I think this example tries too much in one go making it complicated to follow. I think it would be better if we had some smaller examples where 1 pauses and waits for approval on http, another through pub/sub and a third through the wf invocation.
That'll give the users examples of the various ways to raise approval events without complicating it more than it has to

CasperGN (changes requested):

dapr_agents/agents/durable.py
1Ninad: Added `_persist_pending_approvals()` which writes the dict to dapr state store on every write/remove. On start(), `_restore_pending_approvals()` reloads it and drops entries whose workflow is no longer running, so nothing is lost on crash.

dapr_agents/agents/durable.py
```diff
@@ -311,6 +327,11 @@ def __init__(
         self._runtime_owned = runtime is None
         self._registered = False
         self._started = False
+        self._hooks: Optional[Hooks] = hooks
+        # In-memory store of active approval requests, keyed by approval_request_id.
+        # Populated by publish_approval_request activity; consumed by raise_approval_event.
+        # Used by GET /hitl/approvals when serving over HTTP without pub/sub.
+        self._pending_approvals: Dict[str, Dict[str, Any]] = {}
```
CasperGN: Can you update the comment and also make this conditional to having any HITL hooks?

dapr_agents/agents/durable.py
```diff
@@ -23,6 +23,8 @@
 from os import getenv
 from dapr_agents.tool.utils.function_calling import sanitize_openai_tool_name
 import dapr.ext.workflow as wf
+from dapr.ext.workflow import DaprWorkflowClient
```
CasperGN: You have the `wf` just above this and can just use the client from that. It shouldn't be necessary to import the `DaprWorkflowClient` specifically

examples/02-durable-agent-tool-call/hitl_wf_event.py
```diff
@@ -0,0 +1,87 @@
...
+        reason=f"Sent via hitl_wf_event.py (decision={decision})",
+    )
+
+    print(
+        f"Sending decision: instance={instance_id}, "
+        f"request={approval_request_id}, approved={approved}"
+    )
+
+    client = DaprWorkflowClient()
+    client.raise_workflow_event(
```
CasperGN: This still uses `raise_workflow_event` instead of responding with pub/sub. 

The examples gotta be something along the lines of

Example 1:
Require HITL raising the event on http handler -> client responds with http request

Example 2:
Require HITL raising event on pub/sub handler -> client responds with pub/sub request

Example 3: 
Require HITL raising even on workflow handler -> client responds with raising workflow event

Else we don't really get to see and test that all 3 paths work

CasperGN (changes requested):

dapr_agents/agents/durable.py
```diff
@@ -2259,8 +2263,8 @@ def _restore_pending_approvals(self) -> None:
             )
             if not exists or not data:
                 return
-            wf_client = DaprWorkflowClient()
-            _ACTIVE = {WorkflowStatus.RUNNING, WorkflowStatus.SUSPENDED}
+            wf_client = wf.DaprWorkflowClient()
```
CasperGN: This client is created twice. Can we instead move it to `init` on the condition that we have hooks (which would be the only execution path we'd need it)?

CasperGN (changes requested):

CasperGN (approved):
Very nice work @1Ninad - this PR will pave the foundation to the plugin system we got on the roadmap 🚀 🪝

1Ninad:
> Very nice work @1Ninad - this PR will pave the foundation to the plugin system we got on the roadmap 🚀 🪝

Thank you @CasperGN 😃

yaron2 (approved):

1Ninad:
> Very nice work @1Ninad - this PR will pave the foundation to the plugin system we got on the roadmap 🚀 🪝

Hi @CasperGN, are there any upcoming features or enhancements for Dapr Agents that I could contribute to? I would love to help out.

PR #572, opened 2026-04-17
stop over-sanitizing valid names

cicoyle:
preserve hyphens, underscores, and casing in tool/agent names. 

`sanitize_openai_tool_name()` was silently TitleCasing all tool names 
ex: `get-xyz-count` -> `GetXyzCount` even though hyphens, underscores, & mixed casing are valid per both OpenAI && Anthropic tool-name specs (^[a-zA-Z0-9_-]{1,64}$).

Users expect names to remain as-is and not changed bc then they run into naming issues where tools or agent names are not found which is a hard to track down bug.

Now, I strip invalid chars, raise an err if the char length exceeds 64 chars, log a warning if the name is changed due to sanitization to surface this to users.
                                                               
Tool and agent wf names that previously got TitleCased will now preserve their original form
ex: `get-xyz-count` stays `get-xyz-count` instead of becoming `GetXyzCount`

This affects:                       
                                                                                                                                       
- Workflow names: `dapr.agents.GetXyzCount.workflow` -> `dapr.agents.get-xyz-count.workflow`                                                                                    
- Tool names sent to the LLM in the tool schema                                                                                      
                                                                                                                                       
Code that hardcoded the old TitleCased names will need to be updated. LLMs self-correct on the next request since they see the correct name in the tool schema. The old behavior was a bug that no one expects their tool names to be silently rewritten.

### DIFF
--- a/dapr_agents/agents/durable.py
+++ b/dapr_agents/agents/durable.py
@@ -2048,7 +2048,7 @@ class DurableAgent(AgentBase):
         registered: List[str] = []
 
         for name, meta in agents_metadata.items():
-            # Normalize name for lookup since AgentTool sanitizes names to TitleCase
+            # Normalize name for lookup: AgentTool strips spec-illegal chars at registration
             sanitized_name = sanitize_openai_tool_name(name)
             if not self.tool_executor.get_tool(sanitized_name):
                 # Defensive check: ensure meta is a dict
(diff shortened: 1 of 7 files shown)


sicoyle (approved):
amazing - thank you!!

CasperGN (approved):

PR #575, opened 2026-04-20
Fix/structured output resilience

CasperGN:
# Description

_Please explain the changes you've made_

## Issue reference

## Checklist

* [ ] Created/updated tests

### DIFF
--- a/dapr_agents/agents/configs.py
+++ b/dapr_agents/agents/configs.py
@@ -411,6 +411,11 @@ class AgentExecutionConfig:
     tool_choice: Optional[str] = "auto"
     tool_execution_mode: ToolExecutionMode = ToolExecutionMode.PARALLEL
     orchestration_mode: Optional[OrchestrationMode] = None
+    # Number of extra attempts when a structured-output LLM call returns empty
+    # or malformed content. Defaults to 1 retry, which is usually enough to
+    # recover from transient flakiness (e.g. a model emitting tool calls on
+    # the first turn instead of JSON) without masking real schema mismatches.
+    structured_output_retries: int = 1
 
 
 @dataclass
--- a/dapr_agents/agents/durable.py
+++ b/dapr_agents/agents/durable.py
@@ -1630,10 +1630,63 @@ class DurableAgent(AgentBase):
         if tools and self.execution.tool_choice is not None:
             generate_kwargs["tool_choice"] = self.execution.tool_choice
 
-        try:
-            response = self.llm.generate(**generate_kwargs)
-        except Exception as exc:  # noqa: BLE001
-            raise AgentError(str(exc)) from exc
+        # Structured-output calls are where orchestration failures manifest
+        # most often — the model returns empty content or tool calls instead
+        # of JSON. Retry once (configurable) with an explicit nudge so
+        # transient flakiness doesn't abort the whole orchestration.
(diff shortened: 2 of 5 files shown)


dapr_agents/llm/utils/structure.py
```diff
@@ -43,6 +43,20 @@
...
 
 
+def _preview_content(content: Any, limit: int = 200) -> str:
+    """Return a short preview of LLM content for error messages.
+
+    Keeps errors diagnostic without exploding log size when the model
+    returns large payloads.
+    """
+    if content is None:
+        return "<none>"
```
sicoyle, line 53: why?

dapr_agents/agents/durable.py
```diff
@@ -1662,10 +1662,63 @@ def call_llm(
...
+            if response_model is not None
+            else 0
+        )
+        attempt = 0
+        while True:
+            try:
+                response = self.llm.generate(**generate_kwargs)
+                break
+            except Exception as exc:  # noqa: BLE001
+                if response_model is not None and attempt < structured_retries:
```
sicoyle, line 1680: this kind of failure can come from a range of sources - provider or network issues, rate limiting, timeouts, as well as parsing errors or schema mismatches. None of those inherently indicate a structured output failure, yet that’s what this logic assumes. As a result, we risk masking real bugs by retrying under the assumption that the issue is related to structured output when it may not be.

dapr_agents/agents/durable.py
```diff
@@ -1662,10 +1662,63 @@ def call_llm(
...
+                        response_format_name,
+                        attempt,
+                        structured_retries,
+                        self.name,
+                        exc,
+                    )
+                    # Append a system nudge so the LLM is steered back to
+                    # emitting structured content. Rebuild the messages list
+                    # so we don't mutate the caller's reference.
+                    nudge = {
```
sicoyle, line 1694: This introduces non-deterministic behavior by mutating the prompt between retry attempts based on runtime failures. Even if this runs within an activity and is technically safe, it makes the system harder to reason about as identical inputs can lead to different prompts and outputs across attempts.

It also obscures what’s actually happening at the workflow level. If someone is inspecting the workflow to understand retries, it may appear that the activity executed once, when in reality multiple LLM calls were made internally. That mismatch between observed workflow behavior and actual execution can lead to confusion and make debugging more difficult.

dapr_agents/agents/configs.py
```diff
@@ -411,6 +411,11 @@ class AgentExecutionConfig:
     tool_choice: Optional[str] = "auto"
     tool_execution_mode: ToolExecutionMode = ToolExecutionMode.PARALLEL
     orchestration_mode: Optional[OrchestrationMode] = None
+    # Number of extra attempts when a structured-output LLM call returns empty
+    # or malformed content. Defaults to 1 retry, which is usually enough to
+    # recover from transient flakiness (e.g. a model emitting tool calls on
+    # the first turn instead of JSON) without masking real schema mismatches.
+    structured_output_retries: int = 1
```
sicoyle, line 418: optional pls and it needs revision either to rename or repurpose this bc we cannot guarantee that we are retrying due to structured output failures alone based on the conditionals checked.

sicoyle (changes requested):

PR #576, opened 2026-04-20
fix(agents): scope activity names per agent, guard run_tool

CasperGN:
# Description

Two related fixes for `DurableAgent` workflow plumbing that surface when
multiple agents share a `WorkflowRuntime` or when tool-executor state drifts
between dispatch and activity execution.

## 1. Scope activity registrations per agent

Workflows were already registered under agent-scoped names
(`dapr.<framework>.<sanitized_agent_name>.<kind>`), but activities were
registered under their bare method names (`run_tool`, `call_llm`, …).
Dapr's `register_activity` is last-write-wins by name, so two
DurableAgents on the same runtime would silently overwrite each other's
activity bindings — cross-agent activity calls would then execute against
the wrong `self`.

Activities are now registered (and called) under
`dapr.agents.<sanitized_agent_name>.<method>`, mirroring the existing
workflow-name convention. The new `_activity_name` helper accepts either
a string or a bound method reference, and three small helpers
(`_standard_activities`, `_agent_orchestration_activities`,
`_static_orchestration_activities`) keep registration sites and
call-site naming reading from the same source of truth so renames flow
through automatically. The shared `DAPR_AGENTS_NAMESPACE` constant in
`dapr_agents/tool/workflow/agent_tool.py` is reused for both workflow
and activity naming.

## 2. Guard `run_tool` against `WorkflowContextInjectedTool`

`WorkflowContextInjectedTool` instances (e.g. `AgentWorkflowTool` from
`agent_to_tool(...)`) must be routed inline by the workflow's dispatch
loop because they need a `DaprWorkflowContext`. If one reaches the
`run_tool` activity (e.g. due to tool-executor state drift between
workflow replay and activity execution), it previously failed deep in
the executor with the opaque `Missing workflow context. Pass it as
ctx=<DaprWorkflowContext>.` error, which reads like a framework bug.

`run_tool` now performs a sanity check and raises a clear `AgentError`
naming the tool and the structural cause, pointing the operator at the
dispatch loop rather than the framework.

## Issue reference

No linked issue — caught during debugging of co-located DurableAgent
setups and dispatch-loop state drift.

## Checklist

* [x] Created/updated tests
  * Cross-agent activity-name distinctness in
    `tests/agents/durableagent/test_agents_as_tools.py`
    (`test_activity_names_are_agent_scoped`,
    `test_two_agents_have_distinct_activity_names`).
  * `run_tool` guard regression in
    `tests/agents/durableagent/test_durable_agent.py`
    (`test_run_tool_rejects_workflow_context_tool`), plus updates to
    existing activity-name assertions so they tolerate the scoped names.
    * [ ] Created the [dapr/docs](https://github.com/dapr/docs) PR: _N/A — internal naming convention, no user-facing API change._


CasperGN:
~This needs #574 first~

dapr_agents/agents/durable.py
```diff
@@ -2431,19 +2451,71 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
+
+        Activities are registered under ``dapr.agents.<sanitized_name>.<method>``
+        so that multiple agents sharing a ``WorkflowRuntime`` (e.g. development
+        or test setups where agents are co-located) do not clobber each
+        other's bindings. Dapr's ``register_activity`` is last-write-wins by
+        name: without this scoping, the second agent's ``run_tool`` would
+        silently replace the first's, and cross-agent activity calls would
+        execute against the wrong ``self``.
+        """
+        return f"dapr.agents.{sanitize_openai_tool_name(self.name)}.{method_name}"
```
sicoyle: don't we have a const somewhere for the prefix dapr.agents. that you could pls use here?

dapr_agents/agents/durable.py
```diff
@@ -2431,19 +2451,71 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
+        "validate_step",
+        "evaluate_progress",
+        "save_plan",
+    )
+    _STATIC_ORCHESTRATION_ACTIVITY_METHODS = (
+        "initialize_orchestration",
+        "select_next_task",
+        "handle_response",
+        "check_completion",
+        "finalize_orchestration",
```
sicoyle: there is a way that you can reference the actual function objects name here instead of using magic strings where I fear we could create potential for drive on the string names vs if the func signatures ever change. Could you pls use that approach instead?

sicoyle (changes requested):
few comments for you pls 🙏 also if you could pls add a PR description that would be helpful :)

tests/agents/durableagent/test_durable_agent.py
```diff
@@ -1298,6 +1334,11 @@ def track_call_activity(activity, **kwargs):
                 activity_name = activity.__func__.__name__
             else:
                 activity_name = str(activity)
+            # Activities are registered with agent-scoped names like
+            # dapr.agents.<AgentName>.<method>; match on the trailing
+            # method segment so this test works regardless of scoping.
+            if "." in activity_name:
+                activity_name = activity_name.rsplit(".", 1)[-1]
```
seherv: nit: These could be a helper method at the top of the module

sicoyle (commented):
overall LGTM - pls resolve the conflicts and we can send it 🚀

sicoyle:
pls address conflicts and we can send this one!

sicoyle (approved):

PR #581, opened 2026-04-26
Add context-aware structured logger (#225)

### LINKED ISSUE #225: feat: bring in context aware logger

q404365631:
New module `pkg/logger/context_logger.py` providing:

- **ContextLogger** class with thread-safe local storage
- `set_context(**kwargs)` / `clear_context()` for metadata injection
- `@contextmanager bind(**kwargs)` for scoped context
- Auto-formatted log output: `message | key=value | extra=k=v`
- Pre-configured `log` instance for direct use
- Compatible with standard `logging` module

Usage:
```python
from pkg.logger.context_logger import log
log.set_context(run_id="abc123", agent="chatbot")
log.info("Processing request", user_id=42)
# Output: 2025-01-01 [INFO]: Processing request | run_id=abc123 agent=chatbot | user_id=42
```

Closes #225

### DIFF
--- a/pkg/logger/context_logger.py
+++ b/pkg/logger/context_logger.py
@@ -1,0 +1,76 @@
+"""Context-aware structured logger for Dapr Agents.
+
+Provides automatic correlation ID injection, log level management,
+and structured output compatible with observability platforms.
+"""
+
+import logging
+import sys
+import threading
+from contextlib import contextmanager
+from typing import Any, Optional
+
+
+class ContextLogger:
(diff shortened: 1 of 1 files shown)


sicoyle:
Closing in favor of https://github.com/dapr/dapr-agents/pull/584.

Please be sure to check whether an issue is already assigned before picking one up. Thank you!! And thank you for your interest in Dapr Agents! Lmk if you need help finding another issue to contribute to :)

PR #582, opened 2026-04-26
Add Context-Aware Structured Logger v2 (#225)

### LINKED ISSUE #225: feat: bring in context aware logger

q404365631:
StructuredLogger v2 with ISO timestamps, thread-local context, configurable format, stderr routing for errors. Closes #225


sicoyle:
Closing in favor of https://github.com/dapr/dapr-agents/pull/584.

Please be sure to check whether an issue is already assigned before picking one up. Thank you!! And thank you for your interest in Dapr Agents! Lmk if you need help finding another issue to contribute to :)

PR #584, opened 2026-04-26
feat(logger): implement ContextAwareLogger to suppress workflow replay duplicates

### LINKED ISSUE #225: feat: bring in context aware logger

Rishabh-git10:
# Description

Implements a `ContextAwareLogger` to suppress duplicate log pollution during Dapr workflow state rehydration. 

By utilizing `contextvars` and wrapping the `@workflow_entry` decorator, the system automatically tracks the `is_replaying` state globally per execution. This allows the `WorkflowReplayFilter` to drop records during replays without requiring changes to user function signatures.

## Issue reference

Please reference the issue this PR closes: #225

## Checklist

* [x] Created/updated tests
    * [x] Created the [dapr/docs](https://github.com/dapr/docs) PR: https://github.com/dapr/docs/pull/5149

---

### Note on Testing
Verified via unit tests in `tests/utils/test_logger.py`. The `02_durable_agent_trigger.py` quickstart was attempted but found to be blocked by a pre-existing `TypeError` in `dapr_agents/workflow/utils/core.py` unrelated to this PR.

### DIFF
--- a/dapr_agents/utils/logger.py
+++ b/dapr_agents/utils/logger.py
@@ -58,6 +58,12 @@ def with_logger_context(func: Callable[..., Any]) -> Callable[..., Any]:
                 ctx = arg
                 break
 
+        if ctx is None:
+            for val in kwargs.values():
+                if hasattr(val, "is_replaying"):
+                    ctx = val
+                    break
+
         if ctx is not None:
             token = workflow_replaying_ctx.set(ctx.is_replaying)
             try:
--- a/dapr_agents/utils/logger.py
+++ b/dapr_agents/utils/logger.py
@@ -1,0 +1,70 @@
+#
+# Copyright 2026 The Dapr Authors
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import logging
+import contextvars
+from functools import wraps
+from typing import Any, Callable
+
+workflow_replaying_ctx: contextvars.ContextVar[bool] = contextvars.ContextVar(
+    "workflow_replaying_ctx", default=False
+)
+
+
+class WorkflowReplayFilter(logging.Filter):
+    """
+    A logging filter that suppresses log records if the Dapr workflow
+    is currently replaying. This prevents duplicate logs during state rehydration.
+    """
+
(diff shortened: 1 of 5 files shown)


Rishabh-git10:
@CasperGN, addressed the reviews. Ready for another look.

dapr_agents/utils/logger.py
```diff
@@ -0,0 +1,76 @@
...
+            for val in kwargs.values():
+                if hasattr(val, "is_replaying"):
+                    ctx = val
+                    break
+
+        if ctx is not None:
+            token = workflow_replaying_ctx.set(ctx.is_replaying)
+            try:
+                return func(*args, **kwargs)
+            finally:
```
sicoyle, line 79: do we need to capture an exception here?

dapr_agents/utils/logger.py
```diff
@@ -0,0 +1,76 @@
...
+import contextvars
+from functools import wraps
+from typing import Any, Callable
+
+workflow_replaying_ctx: contextvars.ContextVar[bool] = contextvars.ContextVar(
+    "workflow_replaying_ctx", default=False
+)
+
+
+class WorkflowReplayFilter(logging.Filter):
```
sicoyle, line 24: there has to be some way of opting into this from an user experience imo. Some people will want the logs replayed. Others may not.

sicoyle (changes requested):
few comments so far - thank you!!

sicoyle:
a special big thank you @Rishabh-git10 for another contribution to Dapr Agents btw - you rock! 🎉 🙌

dapr_agents/utils/logger.py
Rishabh-git10: I thought the `try...finally` block ensures the context variable is safely reset even if the workflow fails. Any exception will directly go to the Dapr engine to handle retries/faults. Let me know if you prefer we log or wrap the exception here first.

dapr_agents/utils/logger.py
Rishabh-git10: Added the `suppress_replay_logs` toggle here. Let me know if the default `True` is ok to prevent duplicate logging by default, or if we should default to `False`.

Rishabh-git10:
@sicoyle, addressed the review comments. Let me know your thoughts.

PS - Just learning a thing or two. Thanks for the compliments! 🙌

sicoyle (commented):
can you pls update the agent in durable.py now to leverage this? We have a bunch of checks in the workflow with:
```
if not ctx.is_replaying
```
that i think need to pick up on the value that a user sets here pls. And yes, by default we do suppress so a default setting of suppressing the replay logs is great!

Rishabh-git10:
Done with the changes. Please check if it's good-to-go. Thank You!

sicoyle:
> Done with the changes. Please check if it's good-to-go. Thank You!

looking now! thank you!!

sicoyle (approved):
LGTM - thank you!!

sicoyle:
will wait for e2e tests to pass and then send it 😎 🤗

PR #590, opened 2026-04-29
feat: add health endpoints

### LINKED ISSUE #375: Add health endpoints to Dapr Agent for Kubernetes and Dapr probes
When the agent is run in standalone mode (exposed over HTTP and or PubSub), it should expose standard health endpoints to support Kubernetes and Dapr health checks.

Add the following endpoints:
* `/healthz`: Overall application health.
* `/livez`: Kubernetes liveness probe, indicates whether the process should be restarted.
* `/readyz`: Kubernetes readiness probe, indicates whether the agent can accept traffic.

If all 3 endpoints behave the same way, IMO a single healthz would be sufficient.

Additionally, Dapr always probes `/dapr/config` for subscriptions. Instead of returning 404, the agent should return an empty list to avoid surfacing errors when no subscriptions are defined.

Reference: [https://docs.dapr.io/operations/resiliency/health-checks/app-health/](https://docs.dapr.io/operations/resiliency/health-checks/app-health/)

JeffreyJPZ:
# Description

## Changes
- Exposes `/livez` and `/readyz` endpoints when agents are hosted as a service.
- Refactors runtime config updates and agent config resolution so that they use the same coerce/validate/transform/apply logic.
- Creates a agent execution `ToolChoice` enum for consistency.
- Move config resolution logic from the agent to the config classes.
- Extracts agent execution defaults to named constants.

## Issue reference

Please reference the issue this PR closes: #375

## Checklist

* [x] Created/updated tests

TODO:
- [ ] Update docs
- [ ] Update `03-message-router-workflow` example


dapr_agents/workflow/runners/agent.py
```diff
@@ -506,6 +507,8 @@ def serve(
         expose_entry: bool = True,
         entry_path: str = "/agent/run",
         status_path: str = "/agent/instances/{instance_id}",
+        health_check_path: str = "/livez",
+        readiness_check_path: str = "/readyz",
```
JeffreyJPZ: Should Dapr Agent app health endpoints point to `/readyz` rather than `/livez` since it better aligns with [Dapr health checks goals](https://docs.dapr.io/operations/resiliency/health-checks/app-health/#health-check-paths) (i.e. "can it accept traffic?")

dapr_agents/workflow/runners/agent.py
```diff
@@ -597,6 +612,94 @@ def _normalize_path(path: str) -> str:
...
+
+            # Ensure workflow runtime is started and agent workflows/activites are registered
+            if not agent.is_started:
+                return False
+
+            # Ensure subscription consumers are running and able to process messages
+            if getattr(agent, "pubsub", None):
+                if not self._wired_pubsub or not all(
+                    is_ready() for is_ready in self._pubsub_consumer_status_functions
+                ):
```
JeffreyJPZ, line 710: Since streaming subscriptions are pull-based, there's natural backpressure. Would checking for "progress" be useful if throughput is controlled by the app?

dapr_agents/workflow/utils/registration.py
```diff
@@ -397,9 +400,10 @@ def register_message_routes(
         await_timeout: Optional wait timeout in seconds.
         fetch_payloads: Include workflow payloads when waiting for completion.
         log_outcome: Log COMPLETED/FAILED status (either inline or via detached task).
-
     Returns:
-        List of closers that unsubscribe handlers and cancel async workers.
+        A tuple of lists:
+        - List of closers that unsubscribe handlers and cancel async workers.
+        - List of status functions indicating pub/sub consumer readiness.
```
JeffreyJPZ, line 403: Needed a way to surface the consumer statuses to the runner, an alternative could be to keep a "registry" and pass that as an arg, but this is probably better since subscriptions are hidden from the runner?

dapr_agents/workflow/utils/subscription.py
```diff
@@ -635,6 +636,22 @@ def _close() -> None:
...
+        def _make_status_func(subscription: Any) -> Callable[[], bool]:
+            def _is_ready() -> bool:
+                """Check if a stream consumer is able to process messages.
+
+                Returns:
+                    True if a consumer's stream is active.
+                    False if a non-recoverable error or caller-initiated shutdown occurred,
+                        or if the stream is currently reconnecting.
+                """
+                return subscription._is_stream_active()
```
JeffreyJPZ: From my understanding, transient errors and non-recoverable errors should all result in a `subscription.close()` call (through `dapr.clients.grpc.subscription.reconnect_stream()` and `_run_consumer_loop()` respectively) which sets the subscription's internal `_stream_active` flag to `False`. If this is the right approach, should `subscription._is_stream_active()` be made public?

dapr_agents/workflow/runners/agent.py
```diff
@@ -506,6 +507,8 @@ def serve(
         expose_entry: bool = True,
         entry_path: str = "/agent/run",
         status_path: str = "/agent/instances/{instance_id}",
+        health_check_path: str = "/livez",
+        readiness_check_path: str = "/readyz",
         workflow_component: str = "dapr",
```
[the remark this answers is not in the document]
CasperGN: No copilot, you clearly didn't read the comments from Josh or myself.

dapr_agents/workflow/runners/agent.py
CasperGN: You've done the right thing here. In k8s lingo it's `/readyz` and `/livez`. We should update the default dapr to deprecate `/healtz`. Keep as-is in this pr :-)

CasperGN:
@JeffreyJPZ thank you for this one!

Can you move your comments to address the copilot review as it's a little confusing to me which belongs where?

dapr_agents/workflow/utils/subscription.py
```diff
@@ -635,6 +636,22 @@ def _close() -> None:
...
+
```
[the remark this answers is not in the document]
JeffreyJPZ: From my understanding, transient errors and non-recoverable errors should all result in a subscription.close() call (through dapr.clients.grpc.subscription.reconnect_stream() and _run_consumer_loop() respectively) which sets the subscription's internal _stream_active flag to False. If this is the right approach, should subscription._is_stream_active() be made public?

dapr_agents/workflow/utils/subscription.py
```diff
@@ -671,7 +688,7 @@ def subscribe_message_bindings(
     await_timeout: Optional[int],
     fetch_payloads: bool,
     log_outcome: bool,
-) -> List[Callable[[], None]]:
+) -> tuple[List[Callable[[], None]], List[Callable[[], bool]]]:
     """Set up streaming subscriptions for message route bindings.
```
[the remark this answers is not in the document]
JeffreyJPZ: Needed a way to surface the consumer statuses to the runner, an alternative could be to keep a "registry" and pass that as an arg, but this is probably better since subscriptions are hidden from the runner?

dapr_agents/workflow/runners/agent.py
JeffreyJPZ: Since both [app health checks](https://docs.dapr.io/operations/resiliency/health-checks/app-health/#configuring-app-health-checks) and [K8s health checks](https://kubernetes.io/docs/concepts/configuration/liveness-readiness-startup-probes/#configure-probes) are opt-in, should there also be a flag to determine if these endpoints should be exposed (similar to expose_entry)?

dapr_agents/workflow/runners/agent.py
```diff
@@ -597,6 +617,91 @@ def _normalize_path(path: str) -> str:
...
+        readiness_check_path: str,
+        expose_entry: bool,
+        entry_path: str,
+        status_path: str,
+    ) -> None:
+        if health_check_path not in self._default_http_paths:
+            self._default_http_paths.add(health_check_path)
+
+            async def _get_health_status() -> dict[str, str]:
+                return {"status": "ok"}
```
CasperGN: As done with the `_is_ready()` func, can we also include a check for the health endpoint so we know that the agent is actually running healthy?

dapr_agents/workflow/runners/agent.py
```diff
@@ -597,6 +617,91 @@ def _normalize_path(path: str) -> str:
...
+
+            fastapi_app.add_api_route(
+                health_check_path,
+                _get_health_status,
+                methods=["GET"],
+                summary="Get agent health",
+                tags=["health"],
+            )
+
+            logger.info("Mounted health endpoint at %s", health_check_path)
```
CasperGN: Can you convert all the `%s` into f-strings please?

dapr_agents/workflow/runners/agent.py
CasperGN: Yes that would be great. If you can add both an ENV variable + a setting in the [agent execution config](https://github.com/dapr/dapr-agents/blob/main/dapr_agents/agents/configs.py#L403-L413). 
You can see the [agent observability config](https://github.com/dapr/dapr-agents/blob/main/dapr_agents/agents/configs.py#L497). We should let the priority be agent execution config > env variable > default

dapr_agents/workflow/runners/agent.py
```diff
@@ -597,6 +617,91 @@ def _normalize_path(path: str) -> str:
...
+            # Ensure subscription consumers are running and able to process messages
+            if getattr(agent, "pubsub", None):
+                if not self._wired_pubsub or not all(
+                    is_ready() for is_ready in self._pubsub_consumer_status_functions
+                ):
+                    return False
+
+            # Ensure agent routes are mounted
+            if not self._wired_http:
+                return False
```
CasperGN: This will cause all agents where the runner isn't using `.serve()` to return false. 
Should add a check for the attr [fastapi_app](https://github.com/JeffreyJPZ/dapr-agents/blob/0a95ef87f6cc83aa68ce8fac49c8805048e4e04f/dapr_agents/workflow/runners/agent.py#L540)

dapr_agents/workflow/utils/subscription.py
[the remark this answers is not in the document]
CasperGN: This is fine. Both return paths return tuples as far as I can tell

CasperGN (changes requested):
A few comments for changes

dapr_agents/workflow/utils/subscription.py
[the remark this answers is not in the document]
CasperGN: You can rely on the non-public method for now. If it becomes an issue we can look at what needs to be done

dapr_agents/workflow/runners/agent.py
JeffreyJPZ: Just out of curiosity, is f-string logging a ["bad practice"](https://docs.astral.sh/ruff/rules/logging-f-string/)? Here it probably doesn't matter since these lines are evaluated once, but wouldn't hot paths like [`_schedule_workflow`](https://github.com/dapr/dapr-agents/blob/bab804abcc622cb420ff973e79a8a51de3b24df6/dapr_agents/workflow/utils/subscription.py#L372) result in lots of unnecessary evaluations? This is probably a premature optimization though.

dapr_agents/workflow/runners/agent.py
JeffreyJPZ: Would this be the same as `_is_ready` excluding checking pub/sub consumers and HTTP routes (and shutdown signals)?

dapr_agents/workflow/runners/agent.py
JeffreyJPZ: If i'm not misunderstanding something, most CLI flags/K8s annotations aren't automatically made available as env variables in the app? Should we expect the env variable names to follow the [CLI flags/K8s annotations](https://docs.dapr.io/reference/arguments-annotations-overview/) (e.g. ENABLE_APP_HEALTH_CHECK, APP_HEALTH_CHECK_PATH)?

dapr_agents/workflow/utils/registration.py
```diff
@@ -444,7 +448,7 @@ def register_http_routes(
     targets: Optional[Iterable[Any]] = None,
     routes: Optional[Iterable[HttpRouteSpec]] = None,
     loop: Optional[asyncio.AbstractEventLoop] = None,
-) -> List[Callable[[], None]]:
+) -> tuple[List[Callable[[], None]], List[Callable[[], bool]]]:
```
seherv, line 448: I couldn't find any places where this return value of this `register_http_routes()` was actually used 👀 

Is this being done only to keep consistency with `register_message_routes()`, or  will these status functions be used in the future?

dapr_agents/workflow/runners/agent.py
CasperGN: I'd go with `ENABLE_APP_HEALTH_CHECK` and `ENABLE_APP_READY_CHECK`.
You are correct that CLI flags and annotations aren't available as env vars.
In the real deployment model you'd either serve them as env vars on the manifest or you'd mount a configmap as env vars with the details. 

We shouldn't let the path be overwritten imo.

When running locally you can just run `export ENABLE_APP_HEALTH_CHECK` in the terminal for it to be available as env var.

dapr_agents/workflow/runners/agent.py
CasperGN: Hm. I think I was more referring to checking the internals of the agent. The agent could in theory be stuck in a workflow or throwing unrecoverable errors while this would still return `ok`

dapr_agents/workflow/utils/registration.py
```diff
@@ -359,9 +360,11 @@ async def endpoint(body: Any = Body(...)) -> Any:
         )
 
         closers.append(lambda: None)
+        status_functions.append(lambda: True)
+
         logger.info("Mounted HTTP route %s %s -> %s", _method, _path, _name)
```
seherv: Can you please update these (and any other `%` formats in the affected functions) to f-strings too?

dapr_agents/workflow/runners/agent.py
CasperGN: The tiny overhead is not worth the optimization versus the readability gained. f-strings are so much easier to read and review and maintain. It gives consistency across the string formattings

dapr_agents/workflow/utils/subscription.py
```diff
@@ -635,6 +636,40 @@ def _close() -> None:
...
+
+                if not callable(is_stream_active):
+                    return False
+
+                try:
+                    return bool(is_stream_active())
+                except Exception:
+                    logger.exception(
+                        f"Error checking stream consumer {ps_name}:{t_name} status.",
+                        exc_info=True,
```
seherv: `logger.exception()` already includes `exc_info` by default, no need for this :)

dapr_agents/workflow/runners/agent.py
```diff
@@ -597,6 +617,95 @@ def _normalize_path(path: str) -> str:
...
+            ):
+                return False
+
+            return True
+
+        async def _get_ready_status() -> dict[str, str]:
+            if _is_ready(fastapi_app, agent, expose_entry, entry_path, status_path):
+                return {"status": "ok"}
+            raise HTTPException(
+                status_code=503, detail="Agent is not ready, check logs for details"
```
seherv: This might be out of scope for this PR, but `_is_ready()` knows exactly why the readiness check failed and I think it would be useful to have that reason included in the details of this 503.

CC @CasperGN

dapr_agents/workflow/utils/registration.py
JeffreyJPZ: Yeah, I did it for consistency reasons. My reasoning is that pub/sub consumers are stateful, whereas these HTTP handlers can be considered stateless.

dapr_agents/workflow/utils/registration.py
JeffreyJPZ: ~~At this point should I refactor the entire module(s) or just the affected functions?~~

Nvm, I'm guessing refactoring the entire module is probably out of scope.

dapr_agents/workflow/runners/agent.py
JeffreyJPZ: Thats fair. My initial thoughts are on some sort of heartbeat mechanism/watcher thread. I'm also looking at [the unrecoverable error path in the subscription code](https://github.com/dapr/dapr-agents/blob/bab804abcc622cb420ff973e79a8a51de3b24df6/dapr_agents/workflow/utils/subscription.py#L611-L615), it looks like the consumer thread just logs the error and exits? Should there be self-healing by recreating consumer threads, or should it just report unhealthy and have K8s restart the container?

Some questions building off of my initial thoughts:
- What if the watcher thread goes down? Can it be reactivated somehow (maybe reminders?)
- How can one distinguish between a really long workflow and a "stuck" workflow, if that even is possible? AFAIK there's no "global" timeout for workflows (which I guess would defeat the point of a workflow), but maybe this can be configured on the agent somehow?

CasperGN: Regarding the subscription, I think it's fine to report unhealthy and restart. We gotta be careful not to interfere with the expected behavior of retries etc as that comes from the retry policies.

@JoshVanL do you have some input here about how we could probe the wf runtime to infer if it's "running healthy" or not?

dapr_agents/workflow/utils/registration.py
CasperGN: > Nvm, I'm guessing refactoring the entire module is probably out of scope. 

Yeah, we'll get to that bit by bit as we go by. Don't spend time and effort on this PR on it, just as long as we make sure new strings are formatted as f-strings

dapr_agents/workflow/runners/agent.py
CasperGN: Concur, at least surfacing it out to the caller would be beneficial

tests/workflow/test_agent_runner.py
```diff
@@ -0,0 +1,329 @@
...
+    with patch(
+        "dapr_agents.workflow.runners.agent.register_http_routes",
+        return_value=([MagicMock()], [MagicMock()]),
+    ):
+        runner.serve(agent, app=mock_fastapi_app, expose_entry=False)
+
+    client = _MockClient(mock_fastapi_app)
+    livez = await client.get("/livez")
+
+    assert livez.status_code == 200
```
JeffreyJPZ, line 122: Will add some more tests to ensure the config resolution works

dapr_agents/agents/base.py
```diff
@@ -1724,6 +1728,117 @@ def _coerce_datetime(value: Optional[Any]) -> datetime:
...
+                    max_iterations = 10
+
+            tool_choice: Optional[ToolChoice] = None
+            if tool_choice_str := self._runtime_conf.get("TOOL_CHOICE"):
+                try:
+                    tool_choice = ToolChoice(tool_choice_str)
+                except (ValueError, KeyError):
+                    tool_choice = ToolChoice.AUTO
+
+            tool_execution_mode: Optional[ToolExecutionMode] = None
```
JeffreyJPZ: Will there be support for runtime config hot-reloading of tool execution modes? Definitely not in the scope of this PR though

dapr_agents/agents/configs.py
```diff
@@ -369,6 +369,18 @@ class AgentProfileConfig:
...
+class ToolChoice(StrEnum):
+    """
+    Enumeration of supported tool choice strategies for durable agents.
+
+    AUTO: The agent decides when to use tools based on the prompt and context.
+        This is the default and recommended setting for most use cases,
+        as it allows the agent to leverage tools when beneficial while avoiding unnecessary calls.
+    """
+
+    AUTO = "auto"
```
JeffreyJPZ: Added a `ToolChoice` enum for consistency with the other config settings, but looking at some of the LLM interfaces, this might not work for all situations since `tool_choice` can be a [dict when forcing a specific tool](https://github.com/dapr/dapr-agents/blob/3f848e46f337cc9e5872ceb50645aec43f4b27e2/dapr_agents/types/llm.py#L262-L264)?

dapr_agents/agents/configs.py
```diff
@@ -403,15 +415,79 @@ class OrchestrationMode(StrEnum):
...
+    @classmethod
+    def from_env(cls) -> "AgentExecutionConfig":
+        """Create execution config from environment variables."""
+
+        max_iterations: Optional[int] = None
+        if max_iterations := getenv("MAX_ITERATIONS"):
+            try:
+                max_iterations = max(1, int(max_iterations))
+            except ValueError:
+                max_iterations = 10
```
JeffreyJPZ: Should the default max number of iterations be a named constant? Its repeated in quite a few places

CasperGN: Yep, good catch.

dapr_agents/agents/base.py
CasperGN: I think this is a good idea for a later PR yes

dapr_agents/agents/configs.py
CasperGN: Hm. This might need a bit more consideration. It looks like most providers actually use `tool_choice: Optional[Union[str, Dict[str, Any]]]` which leads me to think we should have that structure as well if we wanted to make this stronger typed.

Let's leave this for a follow up

dapr_agents/agents/configs.py
JeffreyJPZ: I put constants for agent execution defaults in the agent config module. I also thought about making a separate constants file like the one for observability but I ran into issues with circular imports (we need the enum at runtime so most workarounds don't apply), so I went back to the first approach.

dapr_agents/agents/configs.py
JeffreyJPZ: Added a TODO for this

dapr_agents/workflow/runners/agent.py
JeffreyJPZ: Looks like there's only the [`wait_for_worker_ready` method](https://github.com/dapr/python-sdk/blob/a56c7b178482d24ddd1214e5b942a75556d39c18/ext/dapr-ext-workflow/dapr/ext/workflow/workflow_runtime.py#L226), but that's probably not enough

dapr_agents/agents/base.py
```diff
@@ -1724,6 +1729,112 @@ def _coerce_datetime(value: Optional[Any]) -> datetime:
                 pass
         return datetime.now(timezone.utc)
 
+    def _resolve_execution_config(self) -> AgentExecutionConfig:
+        """
+        Resolve the execution configuration for the agent in the following order:
+        1. Passed through instantiation (highest priority)
+        2. Environment variables
+        3. Default statestore runtime config (lowest priority)
+
```
sicoyle: hiiii 👋 can you pls update to revise the order here to be
1. statestore runtime config so we can hotreload here
2. explicit instantiation args making sure there is not a None default check
3. env vars

dapr_agents/agents/configs.py
```diff
@@ -369,6 +369,19 @@ class AgentProfileConfig:
     module_overrides: Dict[str, PromptSection] = field(default_factory=dict)
 
 
+class ToolChoice(StrEnum):
```
sicoyle: there should also be "required" and "none" if you can pls add those here

dapr_agents/agents/configs.py
```diff
@@ -435,16 +453,80 @@ class AgentApprovalConfig:
...
+    tool_choice: Optional[ToolChoice] = AGENT_DEFAULT_TOOL_CHOICE
+    tool_execution_mode: ToolExecutionMode = AGENT_DEFAULT_TOOL_EXECUTION_MODE
     orchestration_mode: Optional[OrchestrationMode] = None
     approval: AgentApprovalConfig = field(default_factory=AgentApprovalConfig)
 
+    app_health_check_enabled: Optional[bool] = None
+    app_ready_check_enabled: Optional[bool] = None
+
+    @classmethod
+    def from_env(cls) -> "AgentExecutionConfig":
```
sicoyle, line 641: are there helpers you can use/create here so you can share this logic with the base.py changes above for these checks?

dapr_agents/workflow/runners/agent.py
```diff
@@ -568,7 +584,30 @@ def serve(
                 fetch_status_payloads=fetch_status_payloads,
             )
 
+<<<<<<< health-endpoints
```
sicoyle: looks like this needs clean up from a merge conflict pls

dapr_agents/workflow/runners/agent.py
```diff
@@ -568,7 +584,30 @@ def serve(
...
+            self._mount_ready_endpoint(
+                fastapi_app=fastapi_app,
+                agent=agent,
+                expose_entry=expose_entry,
+                entry_path=entry_path,
+                status_path=status_path,
+            )
+=======
         self._mount_hitl_routes(fastapi_app=fastapi_app, agent=agent)
+>>>>>>> main
```
sicoyle: here too pls

dapr_agents/workflow/runners/agent.py
```diff
@@ -605,6 +644,109 @@ def _normalize_path(path: str) -> str:
             path = f"/{path}"
         return path
 
+    def _mount_health_endpoint(
```
sicoyle, line 645: can you mv these funcs to a agent_health.py or something along those lines pls? make sure to include the license header too

dapr_agents/workflow/runners/agent.py
```diff
@@ -605,6 +644,109 @@ def _normalize_path(path: str) -> str:
...
+        self,
+        *,
+        fastapi_app: FastAPI,
+        agent: DurableAgent,
+    ):
+        app_health_check_path = "/livez"
```
sicoyle, line 651: this should be configurable i believe pls

dapr_agents/workflow/runners/agent.py
```diff
@@ -605,6 +644,109 @@ def _normalize_path(path: str) -> str:
...
+    def _mount_ready_endpoint(
+        self,
+        *,
+        fastapi_app: FastAPI,
+        agent: DurableAgent,
+        expose_entry: bool,
+        entry_path: str,
+        status_path: str,
+    ):
+        app_ready_check_path = "/readyz"
```
sicoyle, line 682: can this be configurable pls with this default

dapr_agents/workflow/runners/agent.py
```diff
@@ -605,6 +644,109 @@ def _normalize_path(path: str) -> str:
...
+        if app_ready_check_path in self._default_http_paths:
+            logger.debug(
```
sicoyle:
```suggestion
            logger.warn(
```

dapr_agents/workflow/runners/agent.py
```diff
@@ -605,6 +644,109 @@ def _normalize_path(path: str) -> str:
...
+                f"Readiness endpoint already mounted at {app_ready_check_path}"
+            )
+            return
+
+        self._default_http_paths.add(app_ready_check_path)
+
+        def _get_agent_readiness(
```
sicoyle, line 691: pls can we avoid nested funcs?

sicoyle (changes requested):
i really like the direction this PR is going with the clean up with the feature. Thank you! Few comments from me pls.

dapr_agents/workflow/runners/agent.py
JeffreyJPZ: Should the paths really be overridable? We were discussing earlier about this:

> I'd go with `ENABLE_APP_HEALTH_CHECK` and `ENABLE_APP_READY_CHECK`. You are correct that CLI flags and annotations aren't available as env vars. In the real deployment model you'd either serve them as env vars on the manifest or you'd mount a configmap as env vars with the details.
> 
> We shouldn't let the path be overwritten imo.
> 
> When running locally you can just run `export ENABLE_APP_HEALTH_CHECK` in the terminal for it to be available as env var.

The above would make sense if `/livez` and `/readyz` were required, but from what I'm seeing they're just conventions?

dapr_agents/agents/configs.py
JeffreyJPZ: Claude, Mistral, and others use "any" instead of "required" (I think)?

mypy.ini
```diff
@@ -12,7 +12,7 @@
 #
 
 [mypy]
-python_version = 3.10
+python_version = 3.11
```
JeffreyJPZ, line 15: Needed to bump the version since `StrEnum` is only supported in >=3.11, and some tests rely on `str(enum)` being equal to the string value which only works in >=3.11. I think it was stale anyway since the pyproject file specifies >=3.11.

dapr_agents/agents/configs.py
JeffreyJPZ: A lot of the logic relates to coercing/validating/applying config values, I think `ConfigFieldDescriptor` can be reused here and on the observability side too (currently it only seems to be used for runtime config).

dapr_agents/agents/base.py
```diff
@@ -841,109 +851,48 @@ def _apply_config_update(self, key: str, value: Any) -> bool:
...
+            applied_value = apply_config_update(
+                target_obj=self,
+                key=normalized_key,
+                value=processed_value,
+                descriptor=descriptor,
+            )
+        except RuntimeError as e:
+            # Fall through if the agent could not be updated but the value is otherwise valid
+            logger.debug(f"Agent {self.name}: {e}")
+            applied_value = None
```
JeffreyJPZ, line 887: Was this behaviour intended from the beginning? Had to split the config update in two which leads to a bit of redundant processing because of this, not sure if there's another way.

JeffreyJPZ:
keep-alive

JeffreyJPZ:
keep-alive

PR #599, opened 2026-05-12
fix(598): allow passing gRPC options for message size either through AgentExecutionConfig or DAPR_GRPC_MAX_INBOUND_MESSAGE_SIZE_BYTES env var

### LINKED ISSUE #598: Support for raising the max payload size
Override the max payload size property to allow for larger payloads then 4MB. This should override the dapr python sdk env var default and allow for complex workflow scenarios.

Depends on https://github.com/dapr/python-sdk/issues/1023.

CasperGN:
# Description

Allow setting dapr client gRPC payload size through either `AgentExecutionConfig` or env var 

## Issue reference

Please reference the issue this PR closes: #598 

## Checklist

* [x] Created/updated tests


dapr_agents/utils/dapr_client_factory.py
```diff
@@ -0,0 +1,135 @@
...
+from typing import Any, Dict, Optional
+
+from dapr.aio.clients import DaprClient as AsyncDaprClient
+from dapr.clients import DaprClient
+
+INBOUND_MESSAGE_SIZE_ENV: str = "DAPR_GRPC_MAX_INBOUND_MESSAGE_SIZE_BYTES"
+
+logger = logging.getLogger(__name__)
+
+_override_inbound_size_bytes: Optional[int] = None
```
sicoyle: is there a way we can avoid a global var? something like an explicit config obj
```

@dataclass(frozen=True)
class DaprClientConfig:
    max_grpc_message_length: Optional[int] = None
```

then
```
def dapr_client_kwargs(
    config: Optional[DaprClientConfig] = None,
    **explicit_kwargs,
):
    ...
```

CasperGN: This was a good call @sicoyle. The initial solution was lazy. I've now wired the `DaprClientConfig` through the durable agent instead

dapr_agents/agents/base.py
```diff
@@ -491,10 +525,18 @@ def __init__(
...
             )
         self.memory = self._memory.store or ConversationListMemory()
         if hasattr(self.memory, "agent_name"):
             self.memory.agent_name = self.name
+        # Forward the agent's Dapr client config to any state-store dependent
+        # collaborators that support it.
+        self._propagate_dapr_client_config_to_collaborators(
+            state=state,
+            registry=registry,
+            memory_store=self.memory,
```
sicoyle: the pubsub class will need this too pls

dapr_agents/agents/base.py
```diff
@@ -996,6 +1050,51 @@ def stop(self) -> None:
...
 
+    # ------------------------------------------------------------------
+    # Dapr client config propagation
+    # ------------------------------------------------------------------
+    @property
+    def dapr_client_config(self) -> Optional[DaprClientConfig]:
+        """Return the resolved Dapr client tuning for this agent, if any."""
+        return self._dapr_client_config
+
+    def _propagate_dapr_client_config_to_collaborators(
```
sicoyle: you're already using a client factory in durable.py and i think we have in the async pubsub call sites, so can the storage/memory/llm classes take the client factory in the same way instead of a DaprClientConfig field? Then we can this func and the refresh_client() calls since there is only one factory rather than N compies of a config we try to keep in sync pls

sicoyle (changes requested):
still have many more files to review, but first pass from me - thank you!!

dapr_agents/agents/base.py
```diff
@@ -381,8 +389,43 @@ def __init__(
...
+            except ValueError as exc:
+                raise ValueError(
+                    "Invalid value for "
+                    "'execution.max_grpc_inbound_message_size_bytes': "
+                    f"{exc}"
+                ) from exc
+
+        # Single canonical client factory shared with every collaborator
+        # (memory, state, registry, LLM). Replaces the previous per-component
+        # ``DaprClientConfig`` copies and the propagation/refresh plumbing.
```
```suggestion
```

sicoyle (commented):
one minor comment pls and this is g2g!! thank you!

sicoyle (approved):
thank you!

PR #605, opened 2026-05-15
feat: Custom client-side @message_router filters

### LINKED ISSUE #561: Client-side filtering for @message_router
## Overview
`@message_router` is a really nice pythonic way to trigger workflows from pub/sub, but it’s currently quite limited when it comes to event filtering. It supports filtering based on message models and basic deduping/routing via CloudEvents id and type respectively, but it’s not very flexible at the moment. Any filtering based on values/conditions requires the models themselves to be modified (e.g. Pydantic post validation). Schemas are prioritized if their [class name matches the CloudEvents type](https://github.com/dapr/dapr-agents/blob/1b35f8dfc04b1f674ffd691b8ded6a372f57f2e7/dapr_agents/workflow/utils/subscription.py#L211), which doesn’t really work if events use the reverse-DNS convention (since class names can’t contain periods). [Deduping relies on the event ID/payload](https://github.com/dapr/dapr-agents/blob/1b35f8dfc04b1f674ffd691b8ded6a372f57f2e7/dapr_agents/workflow/utils/subscription.py#L456-L458) which is suitable for missed acks but doesn’t always work at the app level (e.g. say a user accidentally clicks a button twice, this might generate two events with different IDs). Most importantly, there’s no easy way to “semantically validate” events against external systems without a lot of boilerplate (e.g. checking against a DB), this should generally be kept separate from model validation ([see this issue comment from Pydantic](https://github.com/pydantic/pydantic/issues/857#issuecomment-537570546)). 

## Proposal
Add an optional “pre-validation” hook to `@message_router` that takes in an event payload and metadata:
```python
def pre_validation(event_data: Any, metadata: dict):
```
Add an optional “post-validation” hook to `@message_router` that takes in a validated message model (Pydantic/dataclass instance):
```python
def post_validation(model: Any):
```
Arguments should probably be copies to avoid caller mutation.

## Alternatives Considered
- Filter at the beginning of a workflow, but this results in wasted compute and can lead to cluttered workflow history/logs
- Support CEL (maybe as another argument to `@message_router`) similar to declarative and programmatic subscriptions, but CEL has limited syntax and can’t support complex logic (my understanding is that it’s best suited to metadata and “flat” payloads)

## Things to Consider

- Is there a better name to describe what these hooks are meant for (trigger-filter something?)

- What should users return to indicate that an event should be dropped (or not): a boolean? `TopicEventResponse`?

- Users might want to filter using specific values, how should these be passed? Via closure?

```python
def filter_name(name: str)
	def post_validation(model: Any):
		# Filter using name
		…
	return post_validation

@message_router(pubsub=”pubsub”, topic=”topic”, message_models=UserAddEvent, post_validation=filter_name(“Jimmy”))
def register_jimmy(ctx: DaprWorkflowContext, message: dict) -> str:
	# Triggers only when user is named “Jimmy”
	…
```
- Or via context?
```python
def filter_name(model: Any, **kwargs):
	name = kwargs.get("context", {}).get("name")
	# Filter using name
	…

@message_router(pubsub=”pubsub”, topic=”topic”, message_models=UserAddEvent, post_validation=filter_name, context={“name”: “Jimmy”})
def register_jimmy(ctx: DaprWorkflowContext, message: dict) -> str:
	# Triggers only when user is named “Jimmy”
	…
```
- “Semantic validation” can involve a lot of I/O. Each (pub/sub, topic) subscription maps to [a separate background thread](https://github.com/dapr/dapr-agents/blob/1b35f8dfc04b1f674ffd691b8ded6a372f57f2e7/dapr_agents/workflow/utils/subscription.py#L617-L621), and [messages are evaluated against bindings sequentially](https://github.com/dapr/dapr-agents/blob/1b35f8dfc04b1f674ffd691b8ded6a372f57f2e7/dapr_agents/workflow/utils/subscription.py#L476-L542), this might result in lower throughput for certain subscriptions if pre/post-validation isn’t offloaded (to the main thread?).
- Related to the above, what kinds of dependencies should users have access to and how should they be passed to the hooks (DaprClient is thread-safe at least)?

If there is interest in this, I wouldn’t mind having a crack at it.

-------------------------------------------

acceptance criteria:
- [ ] investigate current limitations of message_router decorator to understand gaps
- [ ] identify and suggestion options (hooks as suggested, context manager?, anything else?)
- [ ] suggestion format can be via issue comment or 150/30min meeting with team to discuss (with decision posted as comment)
- [ ] implement decided solution
- [ ] update docs
- [ ] update any relevant quickstarts/examples using this decorator

seherv:
# Description

Allows users to accept/drop the messages that trigger workflows through `@message_router` with custom **synchronous** callables. These filters can run either before Pydantic model validation (with `payload_filter`) or on the data created by the validated model (with `model_filter`).

Also 
- refactored many deeply nested `workflow.utils.subscription` functions into a shared-state class to make the file more readable
- fixed a bug where the dedupers would mark a message as seen *before* the workflow ran, which made `RETRY` messages always be considered dups 
 
## Issue reference

Please reference the issue this PR closes: https://github.com/dapr/dapr-agents/issues/561

## Checklist

* [x] Created/updated tests


dapr_agents/workflow/utils/registration.py
```diff
@@ -373,7 +387,6 @@ def register_message_routes(
     delivery_mode: Literal["sync", "async"] = "sync",
     queue_maxsize: int = 1024,
     deduper: Optional[DedupeBackend] = None,
-    scheduler: Optional[SchedulerFn] = None,
     wf_client: Optional[wf.DaprWorkflowClient] = None,
```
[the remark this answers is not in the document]
seherv: @dapr/maintainers-dapr-agents If my understanding of this situation is correct, this got removed in 1.0.0 when streaming support was introduced, and it is being deprecated and kept as a no-op shim. 

I deleted it because the next release would make it 3 releases since it was deprecated. What do you think?

dapr_agents/workflow/utils/subscription.py
```diff
@@ -317,344 +391,424 @@ def _shutdown_thread(
...
+                    f"Duplicate detected id={dedup_id} topic={topic_name}; dropping."
+                )
+                return TopicEventResponse(STATUS_SUCCESS)
+
+            response = self._route_to_binding(pairs, topic_name, event_data, metadata)
+
+            if dedup_id is not None and _normalize_status(response.status) in (
+                STATUS_SUCCESS,
+                STATUS_DROP,
+            ):
```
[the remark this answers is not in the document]
seherv: I don't think we should be skipping the dedup logic in the async case. 

If a sync message comes from pubsub and the workflow can't get started for whatever reason, we send pubsub a retry, and we don't ignore the second message when it comes back because it's not marked as seen yet. (The fix introduced in this PR)
 
If an async message comes from pubsub, it gets queued and the async worker takes care of it. The async worker cannot ask for a retry because it does not have access to pubsub, so enqueueing the message is as far as we can get in the code and we have to send the pubsub success from that point. This doesn't negate the need for a dedup cache for async, we could still have dups sent by the publisher and it still makes sense to catch those.

dapr_agents/workflow/utils/subscription.py
```diff
@@ -217,6 +263,51 @@ def _order_pairs_by_cloudevent_type(
...
+    """Run an optional message filter.
+
+    True means proceed, False means skip the binding.
+    An exception from a user-supplied filter is logged and treated as a filter rejection
+    to avoid an infinite retry loop.
+    """
+    if filter_fn is None:
+        return True
+    try:
+        return bool(filter_fn(value, msg_ctx))
```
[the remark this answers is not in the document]
seherv: Deepcopying every single incoming payload is probably too much. I'll just clarify in the docs that filters should not mutate things.

dapr_agents/workflow/utils/subscription.py
```diff
@@ -79,9 +81,51 @@ def mark(self, key: str) -> None:
...
+    Built once per incoming CloudEvent and passed as the second positional
+    argument to ``payload_filter`` and ``model_filter`` callables.
+
+    Attributes:
+        event: Parsed CloudEvent envelope (id, source, type, topic, headers, ...).
+        dapr_client: The DaprClient backing this subscription.
+    """
+
+    event: EventMessageMetadata
+    dapr_client: DaprClient
```
JeffreyJPZ: If filters should be synchronous and I/O free, I'm a bit confused on the purpose of the Dapr client here.

dapr_agents/workflow/utils/subscription.py
```diff
@@ -317,361 +414,453 @@ def _shutdown_thread(
...
+            logger.debug("Dedupe backend mark() error; continuing.", exc_info=True)
+
+    def _route_to_binding(
+        self,
+        pairs: list[BindingSchemaPair],
+        topic_name: str,
+        event_data: Any,
+        metadata: dict | None,
+    ) -> TopicEventResponse:
+        """Pick the first matching binding and dispatch; DROP when nothing matches."""
```
JeffreyJPZ, line 655: Is there anything preventing two separate bindings from having the same pub/sub and message model (seems like the first workflow callable given to [`register_message_routes`](https://github.com/dapr/dapr-agents/blob/da5c09e7de9fcfb9b74a771281159d6cd01cec7f/dapr_agents/workflow/utils/registration.py#L417) always wins out)? I'm assuming fan-out is better with a single orchestrator workflow so maybe there should at least be a warning?

JeffreyJPZ (commented):
Just a couple comments from me, hope you don't mind.

dapr_agents/workflow/utils/subscription.py
seherv: Ah yeah, that was put there before we decided against using I/O in filters, I should remove that now

dapr_agents/workflow/utils/subscription.py
seherv: We don't have anything to stop one binding from "stealing" another binding's message if they have the same message model. I'll add a warning

CasperGN (dismissed):

dapr_agents/types/workflow.py
```diff
@@ -41,13 +46,21 @@ class PubSubRouteSpec:
...
+            `handler_fn`'s `@message_router` decorator (if any).
     """
 
     pubsub_name: str
     topic: str
     handler_fn: Callable[..., Any]
-    message_model: Optional[Type[Any]] = None
-    dead_letter_topic: Optional[str] = None
+    message_model: type[Any] | None = None
+    dead_letter_topic: str | None = None
```
sicoyle, line 61: why are these no longer optional?

dapr_agents/workflow/utils/subscription.py
```diff
@@ -317,361 +440,452 @@ def _shutdown_thread(
...
+        if self.queue is not None:
+            assert self.loop is not None
+            for _ in range(max(1, len(bindings))):
+                self._worker_tasks.append(self.loop.create_task(self._async_worker()))
+
+        for (pubsub_name, topic_name), topic_bindings in _group_bindings_by_topic(
+            bindings
+        ).items():
+            pairs = _build_binding_schema_pairs(topic_bindings)
+            dead_letter_topic = topic_bindings[0].dead_letter_topic
```
sicoyle: why do we hard code the zeroth index here?

examples/03-message-router-workflow/README.md
```diff
@@ -172,6 +172,50 @@ The application entry point registers the workflow and sets up the pub/sub subsc
...
+also drops the binding — but logs the traceback — so a bug in a filter never
+pins a topic in a retry loop.
+
+> ⚠️ **Filters are on the critical path of message intake.** They run on the
+> per-topic consumer thread and **block further messages from being read**
+> until they return. Keep them in-memory, side-effect-free, and fast. If you
+> need an external check — database lookup, HTTP call, state-store read —
+> do it inside the workflow body and return early there.
+
+> 🚫 **`async def` filters are rejected at decoration time.** They look like
```
sicoyle: no emojis pls

examples/03-message-router-workflow/README.md
```diff
@@ -172,6 +172,50 @@ The application entry point registers the workflow and sets up the pub/sub subsc
 - The workflow function itself is passed as the target, not a separate handler function
 - When a message arrives, `register_message_routes` validates it and automatically schedules the workflow
 
+## Filtering messages (`payload_filter` / `model_filter`)
+
+`@message_router` accepts two optional callables that decide whether a message
+should actually trigger the workflow. Both receive a `MessageContext`
+(CloudEvent envelope + the active `DaprClient`) as an argument.
```
sicoyle: there is no dapr client on this dataclass for MessageContext from what I see pls revise the comment on this

dapr_agents/workflow/utils/subscription.py
```diff
@@ -317,361 +440,452 @@ def _shutdown_thread(
...
+        self,
+        pairs: list[BindingSchemaPair],
+        topic_name: str,
+        event_data: Any,
+        metadata: dict | None,
+    ) -> TopicEventResponse:
+        """Pick the first matching binding and dispatch; DROP when nothing matches."""
+        msg_ctx = MessageContext(
+            event=EventMessageMetadata.model_validate(metadata or {}),
         )
```
sicoyle: MessageContext is now built for every message even with no filters configured. If the model is strict about required/extra fields, then malformed but previously tolerated metadata would now raise inside the handle_message try/except and return STATUS_RETRY causing an infinite retry loop i believe. Can you look into this pls?

sicoyle (changes requested):
few comments from me - looks good overall though thank you!!

one nit: can you pls update all the if conditionals to use a switch instead for readability pls?

dapr_agents/types/workflow.py
seherv: Just thought I could use the new optional syntax because we are on 3.10+, and this is mostly new code anyway. 

I always use the new-style types in new code to gradually transition and avoid making gigantic "old-style to new-style" PRs, but honestly those are not too bad to review. If you think this clashes with the rest of the old-style `Optional[x]` I could rewrite these and make the new-style types PR separately 🙂

dapr_agents/workflow/utils/subscription.py
seherv: This code already existed before but I moved it to its own module. Seems like it was okay with any dead letter topic as long as it had the same topic as the binding, so it picked the 0th. 

I just made this a bit more robust 😃

dapr_agents/workflow/utils/subscription.py
```diff
@@ -700,25 +919,26 @@ def subscribe_message_bindings(
...
-        queue_maxsize=queue_maxsize,
         deduper=deduper,
         wf_client=resolved_wf_client,
+        queue_maxsize=queue_maxsize,
         await_result=await_result,
         await_timeout=await_timeout,
         fetch_payloads=fetch_payloads,
         log_outcome=log_outcome,
     )
+    return subscriber.subscribe_all(bindings)
```
sicoyle, line 944: can you pls add tests for all the helpers you created in this file?

examples/03-message-router-workflow/README.md
```diff
@@ -172,6 +172,49 @@ The application entry point registers the workflow and sets up the pub/sub subsc
...
+> **Filters are on the critical path of message intake.** They run on the
+> per-topic consumer thread and **block further messages from being read**
+> until they return. Keep them in-memory, side-effect-free, and fast. If you
+> need an external check (database lookup, HTTP call, state-store read), do
+> it inside the workflow body and return early there.
+
+> **`async def` filters are rejected at decoration time.** They look like
+> they'd run concurrently, but the consumer thread would still block waiting
+> for the result, so the library refuses them to avoid the misleading
+> appearance of concurrency.
```
sicoyle, line 216: why do we reject async filters..?

sicoyle (changes requested):
few comments pls

examples/03-message-router-workflow/README.md
seherv: IMO async is basically an invitation to do I/O. If we do filtering based on something like a DB query and that fails, we would have to manually let the pubsub know we couldn't schedule the workflow and that means duplicated pubsub retry logic in our filter code. With sync filters (that don't fail) we can just send an ack to the pubsub immediately.

And async also sounds like concurrency to me, and I wouldn't want to imply that workflows can trigger out of order or in any way that's not deterministic.

If there's a usecase for async filters I am missing I could create an issue and take a closer look at it in a followup PR 😄

sicoyle (approved):
lgtm - thank you!!

sicoyle:
@seherv can you pls check this PR in your morning 🌞 I'm rerunning the e2e tests which timed out in the last run, so just want to be sure there is nothing up with this PR's changes causing that timeout...

PR #612, opened 2026-05-19
feat(mcp): MCPServer auto-discovery via Dapr metadata API + workflow orchestration

sicoyle:
# Description

This PR adds first-class MCPServer support to `DurableAgent` via Dapr's native MCP integration. Agents now auto-discover MCPServers from the sidecar's metadata API and invoke tools through Dapr workflow orchestration, removing the need for clients to manually wire MCP servers.

Changes:
- **MCPServer auto-discovery** — `DurableAgent` queries the sidecar's metadata API to find configured MCPServer resources, calls the SDK's per-server `ListTools` workflow, and registers each MCP tool as a workflow-aware `AgentTool` automatically.
- **Workflow-based tool invocation** — MCP tool calls now schedule the SDK's per-tool `CallTool` child workflow (`dapr.internal.mcp.<server>.CallTool.<tool>`), inheriting durable execution, retries, and observability without per-agent plumbing.
- **`mcp_tool_def_to_workflow_tool`** — new conversion helper at [dapr_agents/tool/mcp/dapr_workflow_client.py](dapr_agents/tool/mcp/dapr_workflow_client.py) that turns SDK `MCPToolDef` objects into `WorkflowContextInjectedTool` instances ready for `AgentToolExecutor`.
- **`examples/10-mcpserver-all-transports/`** — runnable end-to-end example showing local-tools, remote-weather (HTTP/SSE), and middleware-workflows transports against a real Dapr sidecar.
- Updates us to the 1.18 RC in python SDK which meant making this an exclusion for our 7 day exclusion rules

## Issue reference

## Checklist

* [ ] Created/updated tests


dapr_agents/agents/durable.py
```diff
@@ -3261,6 +3271,78 @@ def stop(self) -> None:
...
+        self._mcp_tools_connected = True
+
+        connected = client.get_connected_servers()
+        if tools:
+            logger.debug(
+                "MCP auto-discovery: connected to %d MCPServer(s), %d tool(s) loaded.",
+                len(connected),
+                len(tools),
+            )
+        elif connected:
```
seherv: nit: I'd use `elif connected and not tools` for readability, even if it's redundant

dapr_agents/agents/durable.py
```diff
@@ -3261,6 +3271,78 @@ def stop(self) -> None:
...
+            logger.warning(
+                "MCP auto-discovery: connected to %d MCPServer(s) but 0 tools loaded "
+                "(servers may have no tools or allowed_tools filtered all out).",
+                len(connected),
+            )
+        else:
```
seherv: nit: I believe this branch cannot be reached

- If we can't connect to a server, an exception is raised before we even get here
- If we didn't raise while connecting, it means there is at least one connected server

dapr_agents/workflow/runners/agent.py
```diff
@@ -82,6 +82,25 @@ def __init__(
...
+        """Sync wrapper for MCP auto-discovery (for non-async entry points)."""
+        if getattr(agent, "_mcp_tools_connected", True):
+            return
+        try:
+            asyncio.get_running_loop()
+            # Already in an event loop — run in a new thread
+            self._run_coro_in_new_loop_thread(self._ensure_mcp_connected(agent))
+        except RuntimeError:
+            # No event loop — safe to use asyncio.run
+            asyncio.run(self._ensure_mcp_connected(agent))
```
seherv: `_run_coro_in_new_loop_thread()` could return a  different `RuntimeError` of its own that's not related to being in an event loop, I'd pull that out of this try:

```suggestion
        try:
            asyncio.get_running_loop()
            is_in_event_loop = True
        except RuntimeError:
            is_in_event_loop = False
        
        coroutine = self._ensure_mcp_connected(agent)
        if is_in_event_loop:
            self._run_coro_in_new_loop_thread(coroutine)
        else:
            asyncio.run(coroutine)
```

dapr_agents/tool/mcp/__init__.py
```diff
@@ -24,11 +24,11 @@
 
 __all__ = [
     "MCPClient",
+    "mcp_tool_def_to_workflow_tool",
     "start_stdio_session",
     "start_sse_session",
     "start_streamable_http_session",
     "start_websocket_session",
     "start_transport_session",
-    "create_pydantic_model_from_schema",
```
seherv, line 32: I saw this function in the last PR, is it okay to drop it from the API?

dapr_agents/tool/executor.py
```diff
@@ -57,8 +57,12 @@ def register_tool(self, tool: Union[AgentTool, Callable]) -> None:
         Args:
             tool (Union[AgentTool, Callable]): The tool to register.
 
+        Duplicate names are not an error: the first registration wins and
+        subsequent attempts log a warning and return.
+
```
seherv: Perhaps this is out of scope but can we do some basic disambiguation in `connect_mcpservers()` as a last resort? We have the client names there, maybe that's enough to break the tie naming it `f'{client}_{tool}'`.

It would move the logic a bit, and I am not sure if duplicate names are that frequent or important to be worth defending against

dapr_agents/tool/utils/serialization.py
```diff
@@ -14,14 +14,81 @@
...
+
+def _unwrap_mcp_call_tool_result(result: Any) -> Any:
+    """Unwrap an MCP CallToolResult into a plain text string.
+
+    The Dapr ``CallTool`` workflow returns a JSON-encoded MCP
+    `CallToolResult <https://modelcontextprotocol.io/specification/2024-11-05/server/tools#tool-result>`_::
+
+        {"isError": false, "content": [{"type": "text", "text": "..."}]}
+
+    Input may be a pre-parsed dict, a JSON string, or even double-JSON-encoded
```
seherv, line 43: I didn't know this was possible 😅

dapr_agents/tool/utils/serialization.py
```diff
@@ -14,14 +14,81 @@
...
+    content = parsed.get("content")
+    if not isinstance(content, list):
+        return "MCP tool call failed" if is_error else result
+
+    parts = [
+        text for item in content if (text := _extract_text_from_content_block(item))
+    ]
+
+    if is_error:
+        return "Error: " + " ".join(parts) if parts else "MCP tool call failed"
```
seherv: Are these errors caused by different reasons? I'd make the messages more explicit if that's the case

examples/10-mcpserver-all-transports/README.md
```diff
@@ -0,0 +1,167 @@
...
+This simulates an externally-hosted MCP server:
+
+```bash
+python weather_sse_server.py
+# Listening on http://0.0.0.0:8081/sse
+```
+
+### 2. Launch the agent with Dapr
+
+```bash
```
seherv: I like putting absolute `cd`s in readmes to make the root of the example very clear:

```suggestion
```bash
cd examples/10-mcpserver-all-transports/README.md
```

dapr_agents/agents/base.py
```diff
@@ -522,6 +523,27 @@ def __init__(
...
+                    logger.debug(
+                        "Discovered %d MCPServer resource(s) from sidecar: %s",
+                        len(self._discovered_mcpserver_names),
+                        self._discovered_mcpserver_names,
+                    )
+                else:
+                    logger.debug(
+                        "No MCPServer resources in sidecar metadata (has_attr=%s, value=%s).",
+                        _has_mcp,
+                        _mcp_val,
```
seherv, line 544: This should be fine for older versions of the runtime, but would this also execute if the runtime takes too long to be ready?

seherv (changes requested):
My comments so far 😄

dapr_agents/tool/mcp/__init__.py
sicoyle: yes this was from pythyon sdk

dapr_agents/tool/executor.py
sicoyle: that's a great call out. In dapr/dapr for MCPServers the mcpserver name is in the workflow name so this is never an issue. I updated to now raise an exception upon duplicate tool name registration.

dapr_agents/agents/base.py
sicoyle: i've updated these to all actually err out :)

pyproject.toml
```diff
@@ -27,9 +27,9 @@ authors = [
     {name = "Dapr Authors", email = "dapr@dapr.io"}
 ]
 dependencies = [
-    "dapr==1.17.3",
-    "dapr-ext-fastapi==1.17.3",
-    "dapr-ext-workflow==1.17.3",
+    "dapr==1.18.0rc0",
```
seherv: Maybe `dapr>=1.18.0rc0,<1.19.0` to avoid endless bumps while we are on RCs, then revert to fixed pin when 1.18 is final? 

`0.0.0rc` with `==` looks odd to me but I don't have a strong argument against it either

seherv (approved):
Looks good to me!

Curious to see what the tool name disambiguation looks like in the real world 👀

CasperGN (approved):

PR #642, opened 2026-06-12
feat: add config resolution for agent execution config

JeffreyJPZ:
# Description

#590 naturally resulted in some refactoring/extra functionality related to agent config resolution that got mixed in with the health endpoints implementation. Since health endpoints is blocked for now I thought I'd open a separate PR for the other stuff.

- Add instantiation/environment variable/state store config validation and resolution for `AgentExecutionConfig` (and `AgentObservabilityConfig`)
- Move config resolution for `AgentObservabilityConfig` and `AgentExecutionConfig` out of  `AgentBase` and `DurableAgent` into their respective classes
- Fix bug where `max_grpc_inbound_message_size_bytes` accepts booleans
- Create shared utilities for getting/coercing/validating/transforming/applying config values onto objects
- Replace some magic numbers/strings with named constants/enums
- Bump mypy to 3.11

## Issue reference

Cherry picks some commits from #590 

## Checklist

* [x] Created/updated tests

TODO:
- ~~[x] Add unit tests for config helpers~~
- ~~[x] Add unit tests for agent execution config~~


dapr_agents/agents/utils/models.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def is_pydantic_model(obj: Any) -> bool:
+    """Checks if the given object is a subclass of Pydantic's BaseModel."""
+    return isinstance(obj, type) and issubclass(obj, BaseModel)
+
+
+def is_supported_config_model(obj: Any) -> bool:
+    """Checks if an object is a supported configuration model (Pydantic, dataclass, or dict)."""
+    return obj is dict or is_dataclass(obj) or is_pydantic_model(obj)
```
JeffreyJPZ, line 27: This actually overlaps with workflow utils [here](https://github.com/dapr/dapr-agents/blob/2156600a2b8015b297e5ce2872ab6b33d48872c6/dapr_agents/workflow/utils/core.py#L236-L243), not sure if they have a reason to evolve independently so maybe a common utils would be better

dapr_agents/agents/configs.py
```diff
@@ -280,6 +343,263 @@ def validate_otel_exporter_logging(v: str) -> str:
...
+    if base.__class__ != override.__class__:
+        logger.warning(
+            f"Cannot merge models of different types: {base!r} and {override!r}"
+        )
+        return base
+
+    try:
+        # Infer model type from the base model
+        model_fields = get_model_fields(base)
+        model_factory = get_model_factory(base)
```
JeffreyJPZ: This also overlaps with [here](https://github.com/dapr/dapr-agents/blob/2156600a2b8015b297e5ce2872ab6b33d48872c6/dapr_agents/workflow/utils/core.py#L333-L369)

dapr_agents/agents/configs.py
```diff
@@ -47,6 +66,12 @@
 _JSON_SCHEMA_DRAFT_URL = "https://json-schema.org/draft/2020-12/schema"
 _JSON_SCHEMA_VERSION_KEY = "version"
 
+# Sentinel value for unset values (instead of None)
+UNSET = object()
```
JeffreyJPZ: It's hard for generic methods to distinguish between a value that is explicitly `None` (e.g. orchestration_mode = None means no orchestration) vs. a value that defaults to `None` because it's not set (e.g. getting env vars). I don't think Python has a built-in feature to support this (yet) so this was the next best thing I found.

JeffreyJPZ: Now that I think about it, it’s really just `OrchestrationMode` that suffers from this so maybe introducing a `'NONE'` state would be simpler (`None` could be internally converted to `'NONE'` for backward compatibility).

JeffreyJPZ: Not needed anymore

mypy.ini
```diff
@@ -79,6 +79,9 @@ ignore_errors = True
 [mypy-dapr_agents.observability.*]
 ignore_errors = True
 
+[mypy-dapr_agents.types.agent.*]
+ignore_errors = True
+
```
JeffreyJPZ: `StrEnum` was introduced in 3.11 and used elsewhere but mypy is stuck on 3.10, need to bump that in a separate PR

dapr_agents/agents/configs.py
```diff
@@ -326,6 +381,243 @@ def validate_otel_exporter_logging(v: str) -> str:
     return v
 
 
+# ---------------------------------------------------------------------------
+# Config helpers
+# ---------------------------------------------------------------------------
```
JeffreyJPZ: Will probably pull these out to a shared utils due to https://github.com/dapr/dapr-agents/pull/665#discussion_r3605338354

dapr_agents/agents/utils/models.py
JeffreyJPZ: done

mypy.ini
JeffreyJPZ: Nvm i think it should be fine tbh

tests/agents/durableagent/test_observability_config.py
```diff
@@ -819,164 +831,3 @@ def test_no_config_sources_returns_defaults(self, mock_llm, monkeypatch):
...
-        mock = Mock(spec=OpenAIChatClient)
-        mock.prompt_template = None
-        mock.__class__.__name__ = "MockLLMClient"
-        mock.provider = "MockOpenAIProvider"
-        mock.api = "MockOpenAIAPI"
-        mock.model = "gpt-4o-mock"
-        return mock
-
-    def test_merge_none_values_dont_override(self, mock_llm):
-        """Test that None values in override don't override base values."""
```
JeffreyJPZ, line 884: Should be safe to remove since its covered by the config tests

tests/agents/durableagent/test_observability_config.py
```diff
@@ -819,164 +831,3 @@ def test_no_config_sources_returns_defaults(self, mock_llm, monkeypatch):
...
-        )
-
-        merged = agent._merge_observability_configs(base, override)
-
-        assert merged.enabled is True  # From base
-        assert merged.service_name == "override-service"  # From override
-        assert merged.endpoint == "http://base-endpoint:4317"  # From base
-
-    def test_merge_boolean_fields_correctly(self, mock_llm):
-        """Test that boolean fields merge correctly with None handling."""
```
JeffreyJPZ, line 920: Same as above

tests/agents/durableagent/test_observability_config.py
```diff
@@ -819,164 +831,3 @@ def test_no_config_sources_returns_defaults(self, mock_llm, monkeypatch):
...
-        )
-
-        merged = agent._merge_observability_configs(base, override)
-
-        assert merged.enabled is False  # Override wins
-        assert merged.logging_enabled is True  # Base wins (override is None)
-        assert merged.tracing_enabled is True  # Override wins
-
-    def test_merge_empty_configs(self, mock_llm):
-        """Test merging two empty configs."""
```
JeffreyJPZ, line 956: This becomes redundant with the refactor

dapr_agents/agents/configs.py
```diff
@@ -617,6 +613,221 @@ def __post_init__(self) -> None:
...
+            ),
+            EnvConfigKey.MAX_GRPC_INBOUND_MESSAGE_SIZE_BYTES: ConfigFieldDescriptor(
+                target_type=Optional[int],
+                setter=lambda obj, v: setattr(
+                    obj, "max_grpc_inbound_message_size_bytes", v
+                ),
+                getter=lambda: getenv("DAPR_MAX_GRPC_INBOUND_MESSAGE_SIZE_BYTES"),
+                should_raise=False,
+            ),
+            # TODO: support orchestration_mode from env
```
JeffreyJPZ, line 658: Ran into tons of issues trying to support `orchestration_mode` since `None` is a valid override, which conflicts with a lot of the existing behaviour. Putting this off for now since runtime config/hot-reloading doesn't support it (yet) which is probably the main benefit.

dapr_agents/agents/configs.py
```diff
@@ -601,6 +597,10 @@ class AgentExecutionConfig:
     builtin_tools: List[str] = field(default_factory=list)
 
     def __post_init__(self) -> None:
+        # Treat None as an empty list to support template configs.
+        if self.builtin_tools is None:
+            self.builtin_tools = []
+
```
CasperGN: This should just return instead. It breaks downstream when using a streaming agent

CasperGN (changes requested):

JeffreyJPZ:
@CasperGN can you review when you have time, thanks

dapr_agents/agents/configs.py
```diff
@@ -601,21 +597,263 @@ class AgentExecutionConfig:
...
+            instantiated_config: Optional instantiated configuration.
+
+        Returns:
+            AgentExecutionConfig instance created from the instantiated configuration.
+
+        Raises:
+            ValueError: If a resolvable field in the instantiated configuration is invalid
+                or if a non-resolvable field is structurally invalid.
+        """
+        instantiated_config = instantiated_config or cls._template_config()
```
sicoyle, line 684: can you pls ensure that env vars are not shadowed when the user passes a partial agent exe config? The `or cls._template_config()` fallback only covers the `None` case, so any user-built config carries non-`None` dataclass defaults into the merge

sicoyle: alternatively, we could try setting the default consts further up to None and then make the ones in the `_base_config` set the the real defaults...

dapr_agents/agents/configs.py
```diff
@@ -285,20 +280,48 @@ def validate_non_empty_string(v: str) -> str:
...
+    """Validate that the tool execution mode is a known ToolExecutionMode value."""
+    try:
+        ToolExecutionMode(v.lower())
+    except (ValueError, KeyError):
+        raise ValueError(
+            f"Unknown tool execution mode '{v}'. "
+            f"Valid options: {[e.value for e in ToolExecutionMode]}"
+        )
+
+    return v
```
sicoyle, line 312:
```suggestion
    return ToolExecutionMode(v.lower())
```

dapr_agents/agents/configs.py
```diff
@@ -601,21 +597,263 @@ class AgentExecutionConfig:
...
+            ),
+            EnvConfigKey.TOOL_CHOICE: ConfigFieldDescriptor(
+                target_type=Optional[
+                    str  # Allow any string as tool choices are permissive
+                ],
+                setter=lambda obj, v: setattr(obj, "tool_choice", v),
+                getter=lambda: getenv("TOOL_CHOICE"),
+                validator=validate_tool_choice,
+            ),
+            EnvConfigKey.TOOL_EXECUTION_MODE: ConfigFieldDescriptor(
```
sicoyle, line 643: can you pls make sure this is case insensitive?

dapr_agents/agents/configs.py
sicoyle: we should probably also make sure we are case insensitive on the returns on these other funcs near here pls...

dapr_agents/agents/configs.py
```diff
@@ -601,21 +597,263 @@ class AgentExecutionConfig:
...
+                getter=lambda: instantiated_config.stream_listener,
+            ),
+            "builtin_tools": ConfigFieldDescriptor(
+                target_type=List[str],
+                setter=lambda obj, v: setattr(obj, "builtin_tools", v),
+                getter=lambda: instantiated_config.builtin_tools,
+            ),
+        }
+
+        apply_config_map(instantiated_config, config_field_map)
```
sicoyle, line 740: this mutates the callers config in place. Can you copy it first pls?

dapr_agents/agents/configs.py
```diff
@@ -601,21 +597,263 @@ class AgentExecutionConfig:
...
+    def resolve_config(
+        cls,
+        config: Optional["AgentExecutionConfig"],
+        runtime_config: Optional[Dict[str, Any]],
+    ) -> "AgentExecutionConfig":
+        """
+        Resolve the execution configuration for the agent in the following order:
+        1. State store runtime configuration (highest priority)
+        2. Passed through instantiation
+        3. Environment variables (lowest priority)
```
sicoyle, line 791: i think the observability section is set to the opposite of this order... is this intentional?

dapr_agents/agents/configs.py
```diff
@@ -727,62 +986,319 @@ class AgentObservabilityConfig:
...
+
+        instantiated_config = AgentObservabilityConfig.from_instantiation(config)
+        logger.debug(f"Instantiated observability config: {instantiated_config}")
+
+        resolved_config = functools.reduce(
+            merge_models,
+            [cls._base_config(), statestore_config, env_config, instantiated_config],
+        )
+
+        logger.debug(f"Final observability config: {resolved_config}")
```
sicoyle, line 1267: this would print things like an auth token and headers in plain text. Can you set `field(repr=False)` on the fields that should not print here plain text pls? The same applies to some of the other debug lines above...

dapr_agents/agents/configs.py
```diff
@@ -601,21 +597,263 @@ class AgentExecutionConfig:
...
+                validator=validate_positive_int,
+                raise_on_error=False,
+            ),
+            EnvConfigKey.TOOL_CHOICE: ConfigFieldDescriptor(
+                target_type=Optional[
+                    str  # Allow any string as tool choices are permissive
+                ],
+                setter=lambda obj, v: setattr(obj, "tool_choice", v),
+                getter=lambda: getenv("TOOL_CHOICE"),
+                validator=validate_tool_choice,
```
sicoyle, line 641: can you append the raise_on_error=False here for consistency pls

dapr_agents/agents/configs.py
```diff
@@ -601,21 +597,263 @@ class AgentExecutionConfig:
...
+
+        Returns:
+            AgentExecutionConfig instance created from state store runtime configuration.
+        """
+        runtime_config = runtime_config or {}
+        config_field_map = {
+            RuntimeConfigKey.MAX_ITERATIONS: ConfigFieldDescriptor(
+                target_type=Optional[int],
+                setter=lambda obj, v: setattr(obj, "max_iterations", v),
+                getter=lambda: runtime_config.get("MAX_ITERATIONS"),
```
sicoyle, line 763: can you address the mix of casing and make these consistent pls?

dapr_agents/utils/config.py
```diff
@@ -0,0 +1,268 @@
...
+        descriptor: An object describing how to process a value for a particular key.
+
+    Returns:
+        The processed value.
+
+    Raises:
+        ValueError: If no value can be retrieved or processing fails.
+    """
+    if not descriptor:
+        raise ValueError(f"Unrecognized config key: {key}.")
```
sicoyle, line 156: this can be removed since it is a frozen dataclass which is always truthy

dapr_agents/utils/config.py
```diff
@@ -0,0 +1,268 @@
...
+            descriptor.setter(target_obj, processed_value)
+        except Exception as exc:
+            raise RuntimeError(f"Could not apply setter for key '{key}'") from exc
+        return processed_value
+    except (ValueError, RuntimeError) as exc:
+        if descriptor.raise_on_error:
+            raise
+
+        if descriptor.fallback is None:
+            logger.debug(f"Ignoring failed config update for key '{key}': {exc}")
```
sicoyle, line 120: can you warn here pls so a bad env var gives signal

dapr_agents/utils/models.py
```diff
@@ -0,0 +1,143 @@
...
+        # Pydantic v1
+        return lambda vals: type(model).parse_obj(vals)
+
+    raise TypeError(f"Unsupported model type: {model!r}")
+
+
+def merge_models(base: T, override: T) -> T:
+    """
+    Merge two models of the same type, with override taking precedence.
+    Only override if the override value is not None.
```
sicoyle, line 86: this says it falls back to override, but the except at 141 always returns base.

sicoyle (changes requested):
few comments from me so far. Also, we will need a docs pr pls for the new env vars 🙏

PR #643, opened 2026-06-12
Feat/streaming agent

### LINKED ISSUE #80: Streaming LLM Output
### Context  
LLMs generate responses iteratively, allowing for intermediate results to be displayed as they are produced. Instead of waiting for a full response, users can see tokens as they're generated. Streaming LLM output from Dapr Agents will improving responsiveness in applications. 

In Dapr Agents, streaming should be supported across:
- **Agents** 
- **DaprChatClient** 
- **Dapr Conversation API**

CasperGN:
# Description

Implements streaming capabilities for durable agents.

## Issue reference

Please reference the issue this PR closes: #80 

## Checklist

* [x] Created/updated tests


dapr_agents/agents/configs.py
```diff
@@ -475,6 +511,14 @@ class AgentExecutionConfig:
...
     approval: AgentApprovalConfig = field(default_factory=AgentApprovalConfig)
     max_grpc_inbound_message_size_bytes: Optional[int] = None
+    streaming: bool = False
+    stream_listener: Optional[Dict[str, Any]] = None
+    # Built-in tools are opt-in. Default is an empty list to preserve strict
+    # backwards compatibility — upgrading agents do not suddenly see new tools
+    # in their LLM tool schema. To enable ``ask_user``, set
+    # ``builtin_tools=["ask_user"]`` *and* ``streaming=True``; the tool is
+    # registered only when both are true and the agent is not an orchestrator.
+    builtin_tools: List[str] = field(default_factory=list)
```
JeffreyJPZ, line 601: Would it be better if `execution.builtin_tools` is list of tool instances instead of tool names (mirroring agent, also users can leverage autocomplete)? Users would import them and then pass them to the execution config. For example, I'm imagining the `ask_user` tool could be decorated with something like `@tool` or `@workflow_context_injected_tool`, but I'm not sure whether it's possible to do everything before runtime.

Also, why not use the existing `tools` on the `DurableAgent` and look for built-ins there?

JeffreyJPZ (commented):
Will try to review in sections.

dapr_agents/agents/schemas.py
```diff
@@ -203,3 +203,13 @@ class AgentWorkflowEntry(BaseModel):
             "approval_request_id. Prevents duplicate publishes when the workflow replays."
         ),
     )
+    stream_context: Optional[Dict[str, Any]] = Field(
```
sicoyle: can we just make this optional str?

CHANGELOG.md
```diff
@@ -0,0 +1,117 @@
+# Changelog
```
sicoyle: can this be included in a separate PR pls? Also, note how small the width is on the text blocks. Can this be expanded a bit when the time comes?

dapr_agents/observability/wrappers/llm.py
```diff
@@ -344,11 +484,168 @@ def _handle_sync_execution(
...
+
+        # Wrap the generator in a small iterator class that supports
+        # attribute assignment so ``StreamEmitter.consume`` can detect the
+        # pre-populated accumulator and skip its own ingest pass.
+        return _AccumulatorBearingIterator(_wrap_iterator(result), accumulator)
+
+    # -- shared helpers for streaming branches ----------------------------
+
+    @staticmethod
+    def _looks_like_non_streaming_result(result: Any, *, async_mode: bool) -> bool:
```
sicoyle: can we have a separate file for the streaming helpers pls?

dapr_agents/llm/utils/request.py
```diff
@@ -167,6 +167,14 @@ def process_params(
                 **params,
             )
 
+        # When streaming for OpenAI-compatible providers, request usage on the
+        # terminal chunk so observability/accumulators can record token counts.
+        # Harmless when the provider ignores the flag.
+        if params.get("stream") and llm_provider in ("openai", "nvidia"):
```
sicoyle: are there any other providers that need to be put here? Should we make this a const somewhere in one of the streaming files so there is one place to update in future?

dapr_agents/llm/dapr/client.py
```diff
@@ -15,30 +15,73 @@
...
 import logging
+import threading
+import time
 
 logger = logging.getLogger(__name__)
 
+#: TTL on cached sidecar metadata. Component lists are near-immutable
+#: within a deployment; a minute-scale TTL keeps hot-reload observable
+#: while avoiding a 10-30 ms gRPC round-trip per ``generate()`` call.
+DEFAULT_METADATA_CACHE_TTL_SECONDS = 60.0
```
sicoyle, line 33: can you pls help me understand how this cache is diff from the prompt and response caching that we have in upstream oss dapr?

dapr_agents/llm/dapr/client.py
```diff
@@ -15,30 +15,73 @@
...
+        if cached is not None and (now - cached[0]) < self._metadata_cache_ttl:
+            return cached[1]
         with self._build_client() as client:
-            return client.get_metadata()
+            value = client.get_metadata()
+        with self._metadata_cache_lock:
+            self._metadata_cache = (time.monotonic(), value)
+        return value
+
+    def invalidate_metadata_cache(self) -> None:
```
sicoyle: when does this get called? If it is tests only then is there a diff approach vs having test specific logic internally?

dapr_agents/llm/dapr/chat.py
```diff
@@ -288,7 +288,11 @@ def generate(
         """
         Issue a non-streaming chat completion via Dapr.
 
-        - **Streaming is not supported** and setting `stream=True` will raise.
+        - **Streaming is opportunistic**: if ``stream=True`` and the Dapr runtime
+          supports streaming conversations, real deltas are returned. Otherwise
+          the call silently falls back to non-streaming and tags the response
+          metadata with ``dapr_streaming_fallback=True`` so callers can emit a
```
sicoyle: is there something clearer we can use instead of dapr_streaming_fallback=True bc we want it clear that the fallback is nonstreaming which that isn't 100% clear to me based on this key yet

sicoyle: do your tests simulate real streaming so we can check this logic given it always go to the fallback case? I want to make sure we don't have any big surprises when the time comes that conversation api supports streaming 😄

sicoyle (commented):
I'm at 28/39 reviewed, overall this is looking great! 👏 🙌 A few questions/comments:

1. Can you confirm hot-reloading `streaming` mid-session works as expected? The issue I'm seeing: `publish_stream_event` and `finalize_workflow`'s SESSION_COMPLETE only check `stream_ctx`, but `call_llm` gates on `stream_ctx and self.execution.streaming`. So if `streaming` flips to `False` mid-session, content deltas stop but control events + the terminal chunk keep flowing. The consumer gets a partial stream with pauses/resume/complete but no content in between. I think the fix is to gate `call_llm` on `stream_ctx` alone like everything else and drop `self.execution.streaming` from that predicate. The frozen context already means "this session streams," so re-reading the live flag is both redundant and the source of the divergence.

2. Any other edge cases here you can think of?

3. Can we gate `ask_user` emission to short-circuit when there's no `stream_ctx` at call time, or just not advertise it for sessions that can't satisfy it? Right now `_register_builtin_tools` advertises it based on the agent-global `streaming` flag, but a non-streaming pub/sub-triggered session (`.subscribe()` / `.serve()` non-streaming entry) has `stream_ctx=None`. If the LLM calls it, `publish_stream_event` no-ops, the workflow suspends on `wait_for_external_event`, no UI ever gets prompted, and it hangs until the 600s sentinel. The gate is checking the wrong scope (agent-global flag vs. per-session context).

4. On mid-execution interruptions: an in-flight `call_llm` at crash time re-executes on restart (never completed, so no cached result), which re-emits START + deltas + TURN_COMPLETE for that turn. Since chunk_id/sequence reset per emitter, consumers can't dedupe it. Is a restarted turn acceptable, or do we need an idempotency token?

5. Related HITL gap: the `USER_INPUT_REQUESTED` chunk that carried `request_id` to the client was a `publish_stream_event` activity, so it's cached and doesn't re-fire on replay. Pub/sub doesn't retain, so a client reconnecting via `GET {status_path}/stream` after a restart never sees the original prompt — and `/input` needs `request_id` + `target_instance_id`, which that client no longer has. So a workflow correctly parked on `ask_user` becomes un-answerable through the stream surface unless the client persisted `request_id` before the crash, and there's no "what input is this instance waiting on?" endpoint to recover it. If HITL-survives-restart is a goal, we'd need either a pending-input lookup on the status route or to re-emit outstanding `USER_INPUT_REQUESTED` on reattach from durable state.

6. One more, adjacent to restart is multi-replica handoff. A rolling restart can rehydrate a workflow on a different replica. In-process listener sessions are bound to the original process's registry, so emits after handoff land nowhere. But `_agent_has_cross_app_peers` only picks pubsub when there are cross-*app* tool targets — it doesn't detect multi-*replica* of the same app. So a horizontally-scaled single agent on the `in_process` default is unsafe across replica handoff, even though the listener docstring says multi-replica must use pubsub. The safe default isn't safe for plain scale-out.

7. Can you denote the schema field version as alpha so consumers can branch on stability, and give a warning when enabled that this is alpha? Are there any other ways we could denote this as alpha? Feels like we should denote this for big features moving forward given we're v1.0. Thoughts?

Sahib-Sawhney-WH:
We think this will be available in the next release?

yaron2 (dismissed):

CasperGN:
> We think this will be available in the next release?

@Sahib-Sawhney-WH I'm working on it yes!

yaron2 (approved):

Issue #631, opened 2026-06-01
feat: add Drasi extension to support agent workflow triggers from Drasi pubsub


JeffreyJPZ:
## Context

Drasi's Kubernetes offering has a pub/sub connector [`PostDaprPubSub`](https://drasi.io/drasi-kubernetes/how-to-guides/configure-reactions/configure-post-pubsub-reaction/) that forwards Drasi change events to pub/sub topics. These events can be consumed by any app that uses Dapr pub/sub (including agents), but requires [a significant amount of boilerplate](https://drasi.io/drasi-kubernetes/how-to-guides/configure-reactions/configure-post-pubsub-reaction/#how-dapr-microservices-consume-messages). This includes: redefining [Drasi event envelopes](https://drasi.io/drasi-kubernetes/how-to-guides/configure-reactions/configure-post-pubsub-reaction/#output-formats) (not a stable API since Drasi is a sandbox project); parsing and filtering in workflows for specific queries and operations, which mixes business logic and "boundary"-level concerns (and makes workflows harder to read); extra indirection as workflows must call agents as child workflows. All of this requires app developers to read Drasi documentation and understand Drasi internals.

## Solution

To reduce developer burden, we should add a `drasi_trigger` function that can route Drasi events directly to agent workflows, supporting the following parameters:
- The agent to target
- Pub/sub component
- Pub/sub topic
- Drasi query ID
- Drasi operation (insert/update/delete)
- Model representing the "view"/change data within the Drasi change event

A "barebones" implementation would need to validate the Drasi event, extract the "view"/change data, and transform it into the format expected by `DurableAgent` (`TriggerAction`).

This will also serve as a reference for developers building their own custom extensions.

## Acceptance Criteria
- [ ] `drasi_trigger` able to be imported as part of a namespace package `dapr_agents.ext.drasi`
- [ ] Agent workflows are triggered directly using `drasi_trigger`
- [ ] Unit tests
- [ ] Dapr docs update
- [ ] Development docs update
- [ ] CI update if necessary
- [ ] Demo E2E app

amansinghoriginal:
@JeffreyJPZ Originally we were considering a decorator as the title of this issue mentions, which would have looked like this:

```
 agent = DurableAgent(...)
  
  @drasi_trigger(agent, query_id="risky-orders")
  def to_task(event):                      
      return TriggerAction(task=f"Handle order {event.payload.after}")

  agent.run()
```

[...]

PR #646, opened 2026-06-18
feat: add mapper for arbitrary workflow inputs with pub/sub triggered workflows

JeffreyJPZ:
# Description

Adds an optional, synchronous `mapper` callable for `@message_router` and `PubSubRouteSpec` allowing users to transform validated message models to any arbitrary JSON-serializable model (which becomes the workflow input). Runs after schema validation and filters.

## Issue reference
Part of #631

## Checklist

* [x] Created/updated tests


dapr_agents/workflow/decorators/decorators.py
JeffreyJPZ, line 147: If type hints/annotations are used (other than `dict` since its the "widest" type) and `mapper` is present it might be confusing since it would suggest that the wf input matches that type when the actual wf input is the type returned by `mapper`.

If `mapper` is present I think it should look for the type of the first `mapper` param instead of the wf input for the `message_model`, since it's opt-in it shouldn't break existing workflows. I'll also add a comment to make this clear.

JeffreyJPZ: Since this is opt-in anyway, we decided that it's non-blocking, if it becomes an issue later we can take a second look.

dapr_agents/workflow/utils/subscription.py
```diff
@@ -85,27 +86,26 @@ def mark(self, key: str) -> None:
...
 ModelFilter = Callable[[Any, "MessageContext"], bool]
+Mapper = Callable[[Any, "MessageContext"], Any]
 
 
-def _validate_filter(
-    filter_fn: Callable[[Any, "MessageContext"], bool] | None,
-    filter_name: str,
+def validate_hook(
+    fn: Callable[[Any, "MessageContext"], Any] | None,
+    name: str,
```
CasperGN, line 94: This rename is perfectly fine to not have backwards compatibility on imo since it was a private function that no external consumer should be using

dapr_agents/workflow/utils/subscription.py
```diff
@@ -85,27 +86,26 @@ def mark(self, key: str) -> None:
...
-            f"`{filter_name}` must be callable, got {type(filter_fn).__name__}."
-        )
-    call_attr = getattr(filter_fn, "__call__", None)
-    is_async_callable = asyncio.iscoroutinefunction(
-        filter_fn
-    ) or asyncio.iscoroutinefunction(call_attr)
+    if not callable(fn):
+        raise TypeError(f"`{name}` must be callable, got {type(fn).__name__}.")
+    call_attr = getattr(fn, "__call__", None)
+    is_async_callable = asyncio.iscoroutinefunction(fn) or asyncio.iscoroutinefunction(
```
CasperGN: This can be simplified down to

```python
if asyncio.iscoroutinefunction(fn) or asyncio.iscoroutinefunction(call_attr):
  ...
```

Since the `boolean` value is only asserted once we don't need to assign it to a value ref. If we expand on the check later on we can always return it but for now it's cleaner to do the inline `if`.

dapr_agents/workflow/utils/subscription.py
```diff
@@ -338,6 +342,36 @@ def _filter_accepts(
...
+        return value
+    try:
+        result = mapper(value, msg_ctx)
+        if not is_supported_model_instance(result):
+            logger.exception(
+                f"mapper for binding '{binding_name}' returned non-JSON-serializable model; skipping binding."
+            )
+            return None
+        return result
+    except Exception:
```
CasperGN: I'm pondering whether it makes sense to catch the exception and re-raise rather than swallow and log it here. It's semantics I guess but makes it easier to understand on the consumer of `_safe_map` that the `is None` path is due to an exception.

CasperGN (changes requested):
Only one change requested. The last comment about the re-raise you can decide how you feel about. I'm leaning on re-raising as it's easier to read the execution flow from the caller side.

dapr_agents/workflow/utils/subscription.py
JeffreyJPZ: Good point, thinking about it now any exceptions in the mapper are highly likely to be due to bugs in user code (accessing an non-existent attribute, returning a non-JSON-serializable model) and not "expected" failures (e.g. validation/coercion failures or malformed payloads). I think makes sense for the caller to get the exception and decide what to do with that, plus the reasons you mentioned above.

dapr_agents/workflow/utils/subscription.py
JeffreyJPZ: Not quite sure what you meant by the first part, I just made the function public following convention since its used at decoration time in `decorators.py` `@message_router` and the `_collect_message_bindings` in `registration.py`.

Edit: misinterpreted it as a question instead of an observation, my bad

dapr_agents/workflow/decorators/decorators.py
```diff
@@ -101,25 +102,35 @@ def message_router(
...
     Returns:
         Callable[[Callable[..., Any]], Callable[..., Any]]:
             The decorated function.
 
     Raises:
-        TypeError: If `message_model` cannot be resolved, or if either filter is
-            an `async def` callable. Filters must be synchronous because they run
+        TypeError:
+            If `message_model` cannot be resolved, or if any hook is
+            an `async def` callable. Hook must be synchronous because they run
```
JeffreyJPZ:
```suggestion
            an `async def` callable. Hooks must be synchronous because they run
```
Small grammar fix

CasperGN (approved):

PR #653, opened 2026-06-18
fix: handle unexpected status types in _normalize_status

### LINKED ISSUE #652: [Bug] _normalize_status switch statement does not cover default case
The [_normalize_status](https://github.com/dapr/dapr-agents/blob/main/dapr_agents/workflow/utils/subscription.py#L300-L311) lacks the `default` case:

```python
def _normalize_status(status: str | TopicEventResponseStatus) -> str | None:
    """Coerce a `TopicEventResponse` status into a status constant if possible"""
    match status:
        case TopicEventResponseStatus():
            raw = status.name
        case str():
            raw = status
    lowered = raw.lower()
    for known in (STATUS_SUCCESS, STATUS_RETRY, STATUS_DROP):
        if known in lowered:
            return known
    return None
```

Which means a case can come through where `lowered = raw.lower()` throws `raw` as undefined.

SAY-5:
# Description

`_normalize_status` matches on `TopicEventResponseStatus()` and `str()` but has no wildcard case, so `raw` is left unbound when a status of any other type comes through. The following `raw.lower()` then raises `UnboundLocalError` instead of returning `None`. The async stream path already documents that unknown statuses should fall through to retry (and checks `status is None`), so returning `None` here matches the intended contract.

Added a `case _: return None` and a regression test for the non-str/non-enum input.

## Issue reference

Please reference the issue this PR closes: #652

## Checklist

* [x] Created/updated tests

### DIFF
--- a/dapr_agents/workflow/utils/subscription.py
+++ b/dapr_agents/workflow/utils/subscription.py
@@ -304,6 +304,8 @@ def _normalize_status(status: str | TopicEventResponseStatus) -> str | None:
             raw = status.name
         case str():
             raw = status
+        case _:
+            return None
     lowered = raw.lower()
     for known in (STATUS_SUCCESS, STATUS_RETRY, STATUS_DROP):
         if known in lowered:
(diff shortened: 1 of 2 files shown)


CasperGN (approved):
LGTM - thanks @SAY-5 for picking this one up!

Issue #630, opened 2026-06-01
[Tracking issue] Add first-class support for Drasi


JeffreyJPZ:
This issue tracks all of the Dapr Agents-related tasks for a Drasi integration with Dapr Agents, which helps build business event-driven agent workflows.

Project context: https://summerofcode.withgoogle.com/programs/2026/projects/qbtyohoJ
Drasi: https://drasi.io/
Drasi tracking issue: drasi-project/drasi-platform#435

- #646
- #665
- #708
- #734
- #742
- dapr/docs#5264
- #744
- #751

PR #665, opened 2026-06-25
feat(ext-drasi): add Drasi extension for DurableAgent

### LINKED ISSUE #631: feat: add Drasi extension to support agent workflow triggers from Drasi pubsub

JeffreyJPZ:
# Description

Initial Drasi extension exposing a `drasi_trigger` function that allows a `DurableAgent` to consume Drasi change events.

`drasi_trigger` takes in an agent and configuration, and registers an activation callback that wires pub/sub routes according to the configuration (with sensible defaults).

Incoming events:
- can be filtered based on the Drasi operation (insert, update, delete) and validated against user-defined models/"views"
- directly trigger agent workflows with customizable task messages as input

This lowers the barrier for building durable, business event-driven AI agents with Drasi as app devs don't need to delve into Dapr Agents internals and maintain their own extensions — the connector logic is reduced to a one-line `drasi_trigger` invocation.

## Issue reference

Closes #631 
Part of #630

## Checklist

* [x] Created/updated tests
    * [x] Created the [dapr/docs](https://github.com/dapr/docs) PR: dapr/docs#5264

## TODO:
* [x] Update extension README.md with example usage
* [x] Add AGENTS.md
* ~~[ ] Update development docs~~ This will be done in a followup
* [x] Verify test module resolution works
* [x] Installable as an extra i.e. `dapr-agents[drasi]`
* [x] Add e2e example in `examples/ext-drasi-pubsub-trigger-k8s`
* [x] Remove some TODOs


ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/activations.py
```diff
@@ -0,0 +1,352 @@
...
+    *,
+    query_id: str,
+    pubsub: str | None = None,
+    topic: str | None = None,
+    dead_letter_topic: str | None = None,
+    task_mapper: DrasiTaskMapper | None = None,
+    operations: DrasiOperation
+    | list[DrasiOperation]
+    | tuple[DrasiOperation]
+    | None = None,
```
JeffreyJPZ: I don't see tuples used often in other SDKs, so maybe its better to remove them to keep the interface simpler.

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/activations.py
```diff
@@ -0,0 +1,352 @@
...
+
+        logger.debug(
+            f"[drasi-trigger]: Activation callback fired for agent '{ctx.agent.name}' with "
+            f"query_id='{query_id}', "
+            f"pubsub='{pubsub}', "
+            f"topic='{topic}', "
+            f"dead_letter_topic='{dead_letter_topic}', "
+            f"operations={operations}, "
+            f"result_model={result_model}"
+        )
```
JeffreyJPZ, line 385: Is there a convention for debug log formatting? Varies quite a lot in the codebase.

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/activations.py
```diff
@@ -0,0 +1,352 @@
...
+
+            * `"i"` - Insert
+            * `"u"` - Update
+            * `"d"` - Delete
+            * `"x"` - Control (lifecycle)
+            * `"h"` - Reload header (snapshot begin)
+            * `"r"` - Reload item (snapshot item)
+
+            Control and reload operations are accepted but currently have no effect (no-op).
+            Defaults to `None` (no filtering).
```
JeffreyJPZ: Since non-change operations are valid and part of Drasi's contract I thought I may as well include them in case they are useful somewhere down the line. That said, I can't think of a clear use case for agents atm (other than deduping/some sort of cache invalidation?), so I'm wondering if it's better to leave this out of the public API.

Edit: even if they were supported I don't think mixing change events (business-related) and lifecycle events is intuitive, so I'll remove this. Instead, it might be possible to attach another hook to do side effects (maybe just before filtering), but it's not that important right now.

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/utils/validation.py
```diff
@@ -0,0 +1,84 @@
...
+        data (dict):
+            The data to validate.
+
+    Returns:
+        Any: The validated/coerced model instance.
+
+    Raises:
+        TypeError: If the model type is unsupported.
+        ValueError: If the data cannot be validated against the model.
+    """
```
JeffreyJPZ, line 73: A bit of duplication is necessary here (I think). `validate_message_model` logs exceptions no matter what so I felt that reusing it would result in too much noise, plus they are different semantically (one is for messages, this is for "fields")

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/activations.py
CasperGN: I think this looks good to me

.github/workflows/build.yaml
```diff
@@ -100,9 +100,13 @@ jobs:
...
       run: |
         uv run mypy --config-file mypy.ini
-    - name: Run Tests
+    - name: Run tests
       run: |
         uv run pytest tests -m "not integration"
+    # TODO: combine this with the above step once non-absolute imports are removed from core tests
+    - name: Run extension tests
+      run: |
+        uv run pytest ext -m "not integration"
```
JeffreyJPZ: Will be addressed in a follow-up PR.

.pre-commit-config.yaml
```diff
@@ -85,3 +85,15 @@ repos:
...
+  # TODO: combine this with the above step once non-absolute imports are removed from core tests
+  - repo: local
+    hooks:
+      - id: pytest
+        name: Run extension unit tests
+        entry: uv run --group test pytest
+        args: [ext, -m, "not integration", --tb=short, -q]
+        language: system
+        pass_filenames: false
+        stages: [pre-push]
```
JeffreyJPZ: Same here

ext/dapr-agents-ext-drasi/pyproject.toml
```diff
@@ -0,0 +1,78 @@
...
+# limitations under the License.
+#
+
+[build-system]
+requires = ["setuptools>=61.0", "setuptools_scm"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "dapr-agents-ext-drasi"
+version = "0.1.0"  # TODO: generate this at build time
```
JeffreyJPZ: And here

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/types.py
```diff
@@ -0,0 +1,33 @@
...
+
+from dapr_agents.ext.drasi.schemas import ChangeNotification
+
+DrasiOperation = Literal["i", "u", "d"]
+
+
+class DrasiChangeEvent(ChangeNotification):
+    """A Drasi event that represents a change emitted by a Drasi query."""
+
+    pass
```
CasperGN: What's the goal of this?

pyproject.toml
```diff
@@ -108,10 +111,7 @@ test = [
     "pytest-rerunfailures>=14.0,<17.0",
 ]
 vectorstore = [
-    "torch>=2.7.0",
-    "sentence-transformers>=4.1.0,<6.0.0",
-    "chromadb>=0.4.22,<2.0.0",
-    "nltk>=3.8.0,<4.0.0",
+    "dapr_agents[vectorstore]",
     "redisvl>=0.4.0,<1.0.0",
```
CasperGN, line 119: Why this change?

CasperGN (commented):
Some initial comments on first round of review

pyproject.toml
JeffreyJPZ: No reason really, I was experimenting with nested optional dependencies at some point and it must've slipped in somewhere. I'll revert it since this isn't really in the scope of the PR.

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/types.py
JeffreyJPZ: I wanted it so that consumers don't need to update their code in case the underlying models change for whatever reason. Since we're not doing any mutation/normalization on our end and users shouldn't be creating events themselves I think a type alias might be simpler (and better) - we can revisit this down the line once we get a better understanding of how the API will be used (maybe with the second part of the project).

I could be missing something though.

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/activations.py
```diff
@@ -0,0 +1,390 @@
...
+    client_factory = getattr(ctx.runner, "_client_factory", None)
+
+    try:
+        deduper = TTLDedupeBackend()
+    except ImportError:
+        logger.warning(
+            f"[drasi-trigger]: cachetools not installed; "
+            f"disabling pub/sub message deduplication for pub/sub component {config.pubsub} and topic {config.topic}"
+        )
+        deduper = None
```
CasperGN: This try/catch isn't necessary

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/activations.py
```diff
@@ -0,0 +1,390 @@
...
+        )
+
+        # Resolve user-provided configuration with documented fallback values
+        resolved_pubsub = pubsub or (
+            ctx.agent.pubsub.pubsub_name if ctx.agent.pubsub else None
+        )
+        resolved_topic = topic or f"{_DRASI_TRIGGER_DEFAULT_TOPIC_PREFIX}-{query_id}"
+        resolved_task_mapper = task_mapper or (
+            lambda event, _: TriggerAction(
+                task=f"{_DRASI_TRIGGER_DEFAULT_TASK}: {event.model_dump_json()}"
```
CasperGN: This `:` will leave a stray `:`. Can we perhaps move it up to the constant instead?

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/activations.py
```diff
@@ -0,0 +1,390 @@
...
+                f"matches the agent's (pubsub, topic): "
+                f"('{ctx.agent.message_bus_name}', topic '{ctx.agent.topic_name}'). "
+                f"Please use a different topic to avoid indeterministic behavior."
+            )
+
+        if config.task_mapper is None:
+            logger.warning(
+                "[drasi-trigger]: No task mapper function provided; "
+                "the agent will be instructed to return the serialized Drasi event as-is."
+            )
```
CasperGN: This is dead code

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/schemas/unpacked/ChangeNotification.py
```diff
@@ -0,0 +1,25 @@
...
+from pydantic import Field
+
+from .ChangePayload import ChangePayload
+from .Notification import Notification
+
+
+class Op(Enum):
+    i = "i"
+    u = "u"
+    d = "d"
```
CasperGN, line 19: Why not use the version in `Op.py`?

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/activations.py
```diff
@@ -0,0 +1,390 @@
...
+    event: DrasiChangeEvent,
+    ctx: MessageContext,
+) -> bool:
+    """
+    Call the provided filter function with the validated Drasi event model.
+    Return `True` if the event passes the filter, `False` otherwise.
+    """
+    logger.debug(
+        f"[drasi-trigger]: Applying filter '{filter_fn.__name__}' for handler '{ctx.handler_name}'"
+    )
```
CasperGN, line 81:
```suggestion
    logger.debug(
        f"[drasi-trigger]: Applying filter '{getattr(filter_fn, '__name__', repr(filter_fn))}' for handler '{ctx.handler_name}'"
    )
```

ext/dapr-agents-ext-drasi/tests/test_drasi_trigger.py
```diff
@@ -0,0 +1,1862 @@
...
+        agent,
+        query_id=query_id,
+        pubsub=drasi_pubsub_name,
+        topic=drasi_topic,
+        task_mapper=lambda event, msg_ctx: TriggerAction(task=f"{event.seq}"),
+        operations=operation,
+    )
+
+    # Should fail when the agent is hosted so it doesn't try to process a valid but unsupported operation
+    with pytest.raises(RuntimeError, match=f"operation.*{operation}"):
```
CasperGN: This checks for `RuntimeError` but in `ext/.../dapr_agents/ext/drasi/activations.py:219` & `ext/.../dapr_agents/ext/drasi/activations.py:224` it raises `TypeError`

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/activations.py
```diff
@@ -0,0 +1,390 @@
...
+
+        def _close():
+            nonlocal closed
+            if closed:
+                return
+            closed = True
+            # This is fine for now since we only wire one pub/sub route,
+            # a closer raising shouldn't result in resource leaks
+            for closer in closers:
+                closer()
```
CasperGN: Are there any scenario where calling `closer()` errors?
That would result in early exit

ext/dapr-agents-ext-drasi/README.md
```diff
@@ -0,0 +1,88 @@
...
+runner = AgentRunner()
+try:
+    runner.subscribe(agent)
+    await wait_for_shutdown()
+finally:
+    runner.shutdown(agent)
+```
+
+### Examples
+- [`ext-drasi-trigger-k8s`](../../examples/ext-drasi-trigger-k8s/README.md) — demonstrates how to subscribe an agent to Drasi queries in a Kubernetes environment.
```
CasperGN:
```suggestion
- [`ext-drasi-trigger-k8s`](../../examples/ext-drasi-pubsub-trigger-k8s/README.md) — demonstrates how to subscribe an agent to Drasi queries in a Kubernetes environment.
```

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/activations.py
```diff
@@ -0,0 +1,390 @@
...
+
+        if event.payload.before is not None:
+            change_model_fields.append(("before", event.payload.before.root))
+        if event.payload.after is not None:
+            change_model_fields.append(("after", event.payload.after.root))
+
+        return all(
+            _is_valid_model_field(config.change_model, data, field_name=field_name)
+            for field_name, data in change_model_fields
+        )
```
CasperGN, line 167: Would it make sense to have a guard here in the event both `before` and `after` is `None`? 
In that event this would loop over an empty list.

Could add:
```python
if not change_model_fields:
    logger.debug("[drasi-trigger]: event has no before/after payload; does not match change model")
    return False
```

ext/dapr-agents-ext-drasi/pyproject.toml
CasperGN: There is some mismatch between having a static version and the setuptools. I'm guessing this is also what you're noting here?

CasperGN (changes requested):

ext/dapr-agents-ext-drasi/pyproject.toml
JeffreyJPZ: I'll just remove the setuptools stuff for now since it's a bit confusing.

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/activations.py
JeffreyJPZ: Moved it into a `_resolve_config` helper that runs before validation and and also made sure to check for it in tests.

ext/dapr-agents-ext-drasi/tests/test_drasi_trigger.py
JeffreyJPZ: It's because the runner catches and re-raises exceptions as `RuntimeError`. I think a context manager that asserts on the original exception via `exc.__cause__` could work.

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/activations.py
JeffreyJPZ: I believe stream close errors are swallowed [here](https://github.com/dapr/dapr-agents/blob/e508c9ba97ebf124a535dd5a386ccd679a2af636/dapr_agents/workflow/utils/subscription.py#L453) in `subscription.py`. Subscriptions (should) be daemon threads so they shouldn't raise [here](https://github.com/dapr/dapr-agents/blob/e508c9ba97ebf124a535dd5a386ccd679a2af636/dapr_agents/workflow/utils/subscription.py#L464-L467).

Just to be safe though I'll wrap the closers in a try-catch so they all get to run.

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/activations.py
JeffreyJPZ: Good point - I think this would be a violation of the event model semantics, so we should check it on every event (if we leave the guard in the change model filter it would only run if the `change_model` is given).

I think I'll have it as a "top-level" filter that runs before all the other filters and make sure it logs a warning.

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/schemas/unpacked/ChangeNotification.py
JeffreyJPZ: Changed it so `DrasiOperation` aliases `Op`.

We should be able to keep the `operations` param on `drasi_trigger` as `DrasiOperation | list[DrasiOperation] | None` but accept enums or literals (if users don't want an extra import) by normalizing to the enum type internally.

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/schemas/unpacked/Versions.py
```diff
@@ -0,0 +1,11 @@
+# generated by datamodel-codegen:
+#   filename:  Versions.yaml
+#   timestamp: 2026-06-25T00:21:18+00:00
+
```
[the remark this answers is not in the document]
JeffreyJPZ: I don't think this needs a license since other generated files like protos don't have licenses, plus the schemas aren't owned by Dapr Agents?

mypy.ini
```diff
@@ -71,3 +78,9 @@ ignore_errors = True
 
 [mypy-dapr_agents.observability.*]
 ignore_errors = True
+
+[mypy-dapr_agents.ext.*]
+ignore_errors = True
+
+[mypy-tests.*]
+ignore_errors = True
```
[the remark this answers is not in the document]
JeffreyJPZ: This tech debt can be pushed til later

pyproject.toml
```diff
@@ -134,7 +137,7 @@ Documentation = "https://github.com/dapr/docs"
 Source = "https://github.com/dapr/python-sdk"
 
 [tool.setuptools.packages.find]
-include = ["dapr_agents"]
+include = ["dapr_agents*"]
 exclude = ["tests"]
```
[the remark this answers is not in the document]
JeffreyJPZ: I don't think this is how it works AFAIK since `setuptools` uses the dir path from the root here (since it defaults to `where = ["."]`), so `ext/dapr-agents-ext-drasi/dapr_agents/...` would evaluate to `'ext.dapr-agents-ext-drasi.dapr_agents.(...)'` which doesn't match.

In any case I ran `uv build` and verified it doesn't include extensions inside the core wheel via the build logs.

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+    "drasi_trigger",
+    "DrasiChangeEvent",
+    "DrasiOperation",
+]
+
+try:
+    __version__ = version("dapr-agents-ext-drasi")
+except PackageNotFoundError:
+    # This should only happen during development
+    __version__ = "0.0.0.dev0"
```
JeffreyJPZ, line 29: As long as the package is installed the version will (for now) always be `0.1.0` (development or not).

ext/dapr-agents-ext-drasi/pyproject.toml
```diff
@@ -0,0 +1,70 @@
...
+    "pytest-mock>=3.12.0,<4.0.0",
+]
+
+[project.urls]
+Documentation = "https://github.com/dapr/docs"
+Source = "https://github.com/dapr/dapr-agents"
+
+[tool.setuptools.packages.find]
+include = ["dapr_agents.ext.drasi"]
+exclude = ["tests"]
```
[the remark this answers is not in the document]
JeffreyJPZ: I don't think this is necessary

.github/workflows/build.yaml
```diff
@@ -99,10 +99,10 @@ jobs:
...
-    - name: Check Typing
+        uv sync --frozen --group test --extra drasi
+    - name: Check typing
       run: |
         uv run mypy --config-file mypy.ini
-    - name: Run Tests
+    - name: Run tests
       run: |
-        uv run pytest tests -m "not integration"
+        uv run pytest tests ext -m "not integration"
```
JeffreyJPZ, line 108: If only extension files are modified then it could be possible to skip core tests to cut down on CI time, but there are pros and cons (e.g. if tests mock parts of the core but the core then changes, hidden coupling might not be easily caught).

CasperGN: Imo the CI time is ok. It's not worth the risk of missing a coupling like you also point out

CasperGN (dismissed):

examples/ext-drasi-change-driven-agents-k8s/README.md
```diff
@@ -0,0 +1,209 @@
...
+- [Drasi CLI](https://drasi.io/drasi-kubernetes/reference/command-line-interface/#get-the-drasi-cli)
+- [uv](https://docs.astral.sh/uv/) package manager
+- OpenAI API key (Azure and direct OpenAI are supported)
+- A Bash terminal
+
+## Setup
+
+### Configure OpenAI credentials
+
+NOTE: Azure OpenAI and direct OpenAI are the only two supported providers — the setup script fails if an API endpoint is given and it is not supported by one of the two providers.
```
sicoyle: why only these two providers?

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/schemas/unpacked/__init__.py
```diff
@@ -0,0 +1,3 @@
+# generated by datamodel-codegen:
```
sicoyle, line 1: do you have instructions or a makefile somewhere documenting how to regenerate this?

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/utils/validation.py
```diff
@@ -0,0 +1,92 @@
...
+
+    # Treat sequences (e.g. strings) as scalar values
+    return [value]
+
+
+def is_supported_operation(value: Any) -> bool:
+    """Return `True` if the value is a supported Drasi operation, otherwise `False`."""
+    try:
+        op = DrasiOperation(value)
+        return op in {DrasiOperation.i, DrasiOperation.u, DrasiOperation.d}
```
sicoyle, line 42: what do the i, u, and d stand for? could that be made clearer here?

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/__init__.py
sicoyle: is the release pipeline updated so when I go to cut a release for dapr agents this will get released and pushed to PyPI too?

sicoyle: why do we have to keep the version 0.1.0?

ext/dapr-agents-ext-drasi/AGENTS.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Public API
+
+All public symbols are exported from `dapr_agents.ext.drasi`:
+
+```python
+from dapr_agents.ext.drasi import (
+    drasi_trigger,      # Register Drasi query subscriptions for an agent
+    DrasiChangeEvent,   # Drasi change-event model emitted by a query
+    DrasiOperation,     # Drasi operation literal: i, u, or d
```
sicoyle: any reason there is a mix of camel casing and snake casing and capitalization on the public api here? can this be consistent?

ext/dapr-agents-ext-drasi/PROVENANCE.md
```diff
@@ -0,0 +1,22 @@
...
+
+The files under `./dapr_agents/ext/drasi/schemas/unpacked/` are generated, do not edit them by hand.
+They are derived from Drasi's `output-unpacked` TypeSpec contract (`Drasi.Unpacked`, v1).
+
+Upstream source:        https://github.com/drasi-project/drasi-platform  
+Contract path:          `typespec/output-unpacked/main.tsp`  
+Vendored from commit:   `3edd930`  
+Vendored on:            `2026-06-24`  
+
+## How the vendored schema was produced
```
sicoyle, line 11: this is what i was looking for :) thank you!

sicoyle: do we need a specific version of anything here for this to produce the same generated results? npm version or something?

ext/dapr-agents-ext-drasi/pyproject.toml
```diff
@@ -0,0 +1,70 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+[project]
+name = "dapr-agents-ext-drasi"
+version = "0.1.0"  # TODO: switch to dynamic versioning
```
sicoyle, line 16: can you pls create an issue for this to help track it. Also i saw another TODO in this PR on enabling users to set concurrency fields and such. Would be nice for that issue to be created and linked to this PR as well 🙏

mypy.ini
```diff
@@ -19,9 +19,16 @@ show_error_codes = True
 check_untyped_defs = True
 install_types = True
 non_interactive = True
+namespace_packages = True
+explicit_package_bases = True
 plugins = pydantic.mypy
 
-files = dapr_agents/**/*.py
+# TODO: these should be dynamically generated
```
sicoyle, line 26: capture this too in the issues you create pls 🙏

mypy.ini
```diff
@@ -71,3 +78,9 @@ ignore_errors = True
 
 [mypy-dapr_agents.observability.*]
 ignore_errors = True
+
+[mypy-dapr_agents.ext.*]
+ignore_errors = True
```
sicoyle: if this is a new pkg then is there any way we can not ignore errs here pls?

sicoyle (commented):
hiiiii! 👋 thank you so much for your hard work on this extension and integration for Dapr Agents!!! You rock, and this looks great! 👏 It looked very clean and well tested, so thank you for that! I had a few questions in general and minor comments if you wouldn't mind reviewing them and addressing as needed pls 🙏 :) Pls feel free to tag me when the time comes and I will rereview quickly and we can look to try and merge this today or Monday!! 💪 🙌 ⭐ 

also, did Casper get with you on where in the Dapr Docs to put information on your awesome contribution here? :)

sicoyle:
Thoughts on the docs issue @JeffreyJPZ ? Would you be willing to write them pls? 🙏 https://github.com/dapr/docs/issues/5252

ext/dapr-agents-ext-drasi/pyproject.toml
JeffreyJPZ: I was thinking that concurrency fields could be a default config on the runner set at init time so activations can just read from there - any explicitly provided field on the entrypoints `subscribe`, `serve`, etc. would be able to override for back compat.

Theres a draft config resolution refactor PR that I have rn that could be nice to have too.

ext/dapr-agents-ext-drasi/PROVENANCE.md
JeffreyJPZ: Maybe the regen needs to be made explicit - I can add it to the README.

examples/ext-drasi-change-driven-agents-k8s/README.md
JeffreyJPZ: Moreso laziness then anything 😄 + matching the other examples.

~~The setup script currently builds a conversation component from k8s secrets so users don't need to figure out what fields are required for their LLM provider. What I could do instead is allow other providers but have them use a separate conversation component that users then manually fill out with their credentials, WDYT?~~

Edit: nvm, the docs are a bit unclear on this but all of the generic providers have the same primary specs so it shouldn't be too difficult to support them all.

ext/dapr-agents-ext-drasi/AGENTS.md
JeffreyJPZ: I think they all follow convention? `drasi_trigger` is the main function while the others are types?

mypy.ini
JeffreyJPZ: Your'e probably right, ~~its annoying though b/c it refuses to recognize extension packages for some reason so there'll be a lot of inline ignores.~~

Edit: not quite sure what changed from before but those seem to be gone now, still some errors in the generated files that have to be ignored.

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/utils/validation.py
JeffreyJPZ: The ops are from here: https://drasi.io/drasi-kubernetes/how-to-guides/configure-reactions/configure-post-pubsub-reaction/#unpacked-format. There's also a docstring on the type in `types.py` I believe.

We kept the trigger + types thin intentionally to avoid having to manually redefine the Drasi models since we rely on them for validation.

JeffreyJPZ:
@sicoyle I addressed most of the questions above,

I can take on the docs issue too, but I'm trying to prep for a meetup talk next week about the project so I'll have to see what progress I can make.

ext/dapr-agents-ext-drasi/dapr_agents/ext/drasi/__init__.py
JeffreyJPZ: Re the release questions:

> is the release pipeline updated so when I go to cut a release for dapr agents this will get released and pushed to PyPI too?

Changes should be [here](https://github.com/dapr/dapr-agents/pull/665/changes#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R44-R54)

> why do we have to keep the version 0.1.0?

This kind of relates to the first question - if coordinated releases w/ independent versioning is kept then I think there needs to be an improved release pipeline, but I didn't really want to include that in this PR. As a temp workaround I thought it'd be ok to keep the version in the `pyproject.toml` and manually bump the extension version, one caveat is that the "development version" will always match the "release version".

sicoyle (approved):
lgtm, thank you for your contribution!! and best of luck in your presentation 🎉

PR #676, opened 2026-06-30
fix(embedder): normalize single-chunk OpenAI embeddings

### LINKED ISSUE #675: OpenAIEmbedder.embed does not normalize single-chunk embeddings when normalize=True
### Bug

`OpenAIEmbedder` defaults to `normalize=True` and its docstring promises normalized embeddings, but normalization only happens on the multi-chunk path (inside `_process_embeddings`). The single-chunk branch returns the raw provider vector:

```python
# dapr_agents/document/embedder/openai.py (embed)
if len(embeddings) == 1:
    # If only one chunk, use its embedding directly
    results.append(embeddings[0])   # raw vector, never normalized
```

A chunk only splits when text exceeds `max_tokens` (8191), so virtually every real query / short doc is a single chunk and comes back un-normalized. Net effect: `normalize=True` and `normalize=False` are indistinguishable for ordinary inputs, and long (chunked) docs are normalized while short ones are not, mixing magnitudes and breaking cosine-similarity comparisons in RAG / vector flows.

`NVIDIAEmbedder` (`dapr_agents/document/embedder/nvidia.py`) normalizes every vector unconditionally, so this is an `OpenAIEmbedder`-only gap.

### Expected behavior

With `normalize=True`, every returned embedding has unit norm regardless of chunk count. With `normalize=False`, raw vectors are returned (unchanged).

### Steps to reproduce

Embedding `[3.0, 4.0]` has raw norm `5.0` and unit norm `1.0`. With the OpenAI provider mocked to return `[3.0, 4.0]` (no network):

```python
from dapr_agents.document.embedder.openai import OpenAIEmbedder

emb = OpenAIEmbedder()                 # normalize=True by default
emb.embed("hello")                     # -> norm 5.0 (raw); expected ~1.0
emb.embed("hello") == OpenAIEmbedder(normalize=False).embed("hello")  # True (proves normalize=True is a no-op)
```

### Release / version

main (current `dapr_agents/document/embedder/openai.py`)

I'd like to fix this; please `/assign` me. A small PR mirroring `NVIDIAEmbedder` (normalize the single-chunk branch when `self.normalize` is set) with a focused unit test is ready.

anxkhn:
# Description

`OpenAIEmbedder.embed` only normalized vectors on the multi-chunk path (via `_process_embeddings`). With the default `normalize=True`, single-chunk inputs (any text under `max_tokens`, i.e. the common case) were returned raw, so `normalize=True` and `normalize=False` behaved the same and cosine-similarity comparisons broke. This mirrors `NVIDIAEmbedder` by normalizing the single-chunk branch too; `normalize=False` still returns raw vectors and the multi-chunk path is unchanged. Adds focused unit tests (`create_embedding` mocked, no network).

## Issue reference

Please reference the issue this PR closes: #675

## Checklist

* [x] Created/updated tests
    * [ ] Created the [dapr/docs](https://github.com/dapr/docs) PR: not required (internal correctness fix, no API / feature / config change; behaviour now matches the existing docstring and `NVIDIAEmbedder`)

## AGENTS.md Notes

- AGENTS.md says a dapr/docs PR is required for API / feature / breaking / config changes. Consider adding an explicit line that pure internal correctness fixes (no public API or config change) do not need one, so contributors are not blocked on a no-op docs PR.
- The "Before commit (REQUIRED)" command block is excellent and worked verbatim.
- The Apache 2.0 test-file header convention is enforced in review but only discoverable by reading existing test files; one line in AGENTS.md pointing at it would help first-time contributors.

### DIFF
--- a/dapr_agents/document/embedder/openai.py
+++ b/dapr_agents/document/embedder/openai.py
@@ -160,7 +160,12 @@ class OpenAIEmbedder(OpenAIEmbeddingClient, EmbedderBase):
         for embeddings, tokens in zip(grouped_embeddings, tokenized_inputs):
             if len(embeddings) == 1:
                 # If only one chunk, use its embedding directly
-                results.append(embeddings[0])
+                embedding = embeddings[0]
+                if self.normalize:
+                    embedding = (
+                        np.array(embedding) / np.linalg.norm(embedding)
+                    ).tolist()
+                results.append(embedding)
             else:
                 # Combine chunk embeddings using weighted averaging
                 weights = [
--- a/tests/document/embedder/test_openai.py
+++ b/tests/document/embedder/test_openai.py
@@ -1,0 +1,65 @@
+#
+# Copyright 2026 The Dapr Authors
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from unittest.mock import MagicMock, patch
+
+import numpy as np
+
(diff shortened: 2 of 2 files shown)


CasperGN (approved):
Thank you @anxkhn - and awesome to see the notes made it in. I'll come with an improvement shortly 🚀

PR #687, opened 2026-07-05
fix(memory): record createdAt in UTC instead of mislabeled local time

anxkhn:
`ConversationDaprStateMemory.add_message` stamps every persisted message with a
`createdAt` timestamp:

```python
message.update({"createdAt": datetime.now().isoformat() + "Z"})
```

`datetime.now()` (called with no timezone argument) returns a **naive** datetime
in the host's **local** time. Hard-appending the literal `"Z"` then asserts, per
ISO 8601, that this instant is UTC. So on any host that is not set to UTC, the
stored `createdAt` is wrong by the host's UTC offset while claiming to be UTC,
and the error shifts with daylight saving time.

Example on a host in `America/New_York` (UTC-4 in summer): the code records
`2026-07-03T02:39:22.427252Z`, but the true instant is
`2026-07-03T06:39:22.427263+00:00`, a 4-hour error mislabeled as UTC.

This is the only occurrence of the `isoformat() + "Z"` pattern in the codebase.
The fix uses a timezone-aware UTC timestamp, matching the sibling
`ConversationVectorMemory`, which already writes
`datetime.now(timezone.utc).isoformat()`:

```python
message.update({"createdAt": datetime.now(timezone.utc).isoformat()})
```

`createdAt` is now a correct, timezone-aware UTC ISO-8601 string (`...+00:00`)
regardless of the host timezone.

A regression test (`tests/memory/test_daprstatestore.py`) forces a non-UTC local
clock (4 hours ahead of UTC), calls `add_message`, captures the value written to
the Dapr state store, parses `createdAt`, and asserts it is timezone-aware and
equals the true UTC instant (and is not the naive local wall clock mislabeled as
UTC). It fails on the old code and passes on the fix.

## Issue reference

## Checklist

* [x] Created/updated tests
    * [ ] Created the dapr/docs PR: _not required; internal-only bug fix (no
      public API / feature / config / documented-behavior change)._

## AGENTS.md Notes

- The `AGENTS.md` "Before commit (REQUIRED)" gate (`ruff format` -> `flake8` ->
  `mypy --config-file mypy.ini` -> `pytest tests -m "not integration"`) was clear
  and matched CI exactly; all steps passed locally. Nothing needed to be guessed.
- Suggestion: the "Test Structure" section lists `agents/`, `llm/`, `workflow/`
  as the unit-test dirs; `memory/` (which this PR adds a test under) exists in
  `tests/` too. Listing it, or noting the list is illustrative, would remove a
  small ambiguity for contributors touching `dapr_agents/memory/`.

### DIFF
--- a/dapr_agents/memory/daprstatestore.py
+++ b/dapr_agents/memory/daprstatestore.py
@@ -13,7 +13,7 @@
 
 import json
 import logging
-from datetime import datetime
+from datetime import datetime, timezone
 from typing import Any, Dict, List, Optional, Union
 
 from pydantic import Field
@@ -87,7 +87,7 @@ class ConversationDaprStateMemory(MemoryBase):
         message = self._convert_to_dict(message)
         message.update(
             {
-                "createdAt": datetime.now().isoformat() + "Z",
+                "createdAt": datetime.now(timezone.utc).isoformat(),
             }
         )
 
--- a/tests/memory/test_daprstatestore.py
+++ b/tests/memory/test_daprstatestore.py
@@ -1,0 +1,63 @@
+#
+# Copyright 2026 The Dapr Authors
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


sicoyle (approved):
thank you for keepign the scope on the PR super small. I greatly appreciate it on the reviewing side, esp with your tests 🙏 

also, thank you for your interest in Dapr Agents and for contributing 🚀

PR #689, opened 2026-07-07
docs(tool): fix do_http_request docstring parameter name

### LINKED ISSUE #688: Docstring for DaprHTTPClient.do_http_request uses stale endpoint_url parameter name
The Args section for `DaprHTTPClient.do_http_request` documents a parameter named `endpoint_url`, but the method signature uses `endpoint` and there is no `endpoint_url` parameter.

This is a docstring-only mismatch. The description text remains correct; only the documented parameter name needs to match the actual signature.

anxkhn:
The `Args` section of `DaprHTTPClient.do_http_request` documents a parameter named `endpoint_url`, but the method has no such parameter. The actual parameter is `endpoint`.

`endpoint` is the name used everywhere else: it is the signature parameter (`dapr_agents/tool/http/client.py`) and it is passed through as `self._validate_endpoint_type(endpoint=endpoint, ...)`. The docstring kept the older `endpoint_url` name after the parameter was renamed to `endpoint` in #107. `git grep endpoint_url` returns no other occurrence in the repo, so this is a stale name in the docstring only.

This PR renames that one `Args` entry to match the real parameter. The description text ("The host of the URI to send the request to.") is unchanged and stays accurate.

No behavior change: this is a docstring-only, one-line correction.

## Issue reference

Please reference the issue this PR closes: #688

## Checklist

* [ ] Created/updated tests
    * [ ] Created the [dapr/docs](https://github.com/dapr/docs) PR: n/a - internal docstring correction, no public API / behavior change

This change does not add or alter any public API, feature, configuration option, or documented behavior: it corrects an in-code docstring so the `Args` block names the parameter that actually exists (`endpoint`). No `dapr/docs` content corresponds to it. Tests are not applicable because the change is docstring-only with no observable behavior change, and the repo has no precedent for asserting docstring text.

## AGENTS.md Notes

- The pre-commit gate in `AGENTS.md` ran clean as written (`ruff format` -> `flake8 dapr_agents tests --ignore=E501,F401,W503,E203,E704` -> `mypy --config-file mypy.ini` -> `pytest tests -m "not integration"`): 403 files unchanged, flake8 exit 0, mypy no issues, 938 passed / 15 skipped.


dapr_agents/tool/http/client.py
```diff
@@ -130,7 +130,7 @@ def do_http_request(
         Send a POST request to the specified endpoint with the given input.
 
         Args:
-            endpoint_url (str): The host of the URI to send the request to.
+            endpoint (str): The host of the URI to send the request to.
             payload (dict[str, str]): The payload to include in the request.
```
JeffreyJPZ:
```suggestion
            payload (dict[str, str]): The payload to include in the request.
            endpoint (str): The host of the URI to send the request to.
```

dapr_agents/tool/http/client.py
JeffreyJPZ:
```suggestion
            A tuple with the HTTP status code and response, or a ToolError.
```

dapr_agents/tool/http/client.py
```diff
@@ -130,7 +130,7 @@ def do_http_request(
         Send a POST request to the specified endpoint with the given input.
```
JeffreyJPZ:
```suggestion
        Send an HTTP request to the specified endpoint.
```

JeffreyJPZ (commented):
Thanks for the fix! Since we're fixing this docstring, mind cleaning up a few more things?

I believe there's also a stale log issue around line 150 if you'd like to take that up in a followup, otherwise I can maybe take it up since it requires some code changes.

sicoyle:
thanks for the review @JeffreyJPZ  🙌 @anxkhn would you mind addressing the feedback here pls and you can tag me when it's ready and I can merge :)

sicoyle (approved):
thank you!!

PR #690, opened 2026-07-07
feat(schemas): add caller_claims to TriggerAction for verified user identity propagation

sicoyle:
# Description

What

This change adds an optional caller_claims: Optional[CallerClaims] field to the TriggerAction model in dapr_agents/agents/schemas.py, along with a new CallerClaims Pydantic model that holds verified user identity. The CallerClaims model captures the caller's subject, tenant, scopes, email, issuer_id, and claims_hash.

The key difference from the existing caller_headers field is serialization behavior. Where caller_headers is excluded from serialization, caller_claims is serialized into the workflow input. This gives us three concrete benefits: the signed workflow history now records who triggered the workflow, giving us an audit lineage; PropagateLineage() can carry the caller's identity forward into child workflows; and sub-agents automatically inherit the verified user identity from their parent.

Why

Lifecycle plugins can already verify an inbound credential and attach the caller's identity to the in-memory context, but that context is transient and doesn't persist into workflow state. Without a serialized field on TriggerAction, child workflows have no way to learn the user identity that PropagateLineage() is meant to carry forward. This change adds the missing persistence layer so that identity survives across workflow boundaries.

Security

CallerClaims acts as an explicit allowlist of exactly what is allowed to propagate into workflow state. By design, it has no field for raw tokens — it holds only verified claims. Raw credentials continue to live in caller_headers, which is marked exclude=True and is never serialized. Two tests enforce this boundary: one asserts that CallerClaims has no token-like field, and another asserts that a raw bearer token placed in caller_headers never appears in serialized output, while caller_claims does.

Backwards compatibility

caller_claims defaults to None, so existing publishers and consumers are completely unaffected. Only new consumers that explicitly read or write the field will interact with it.

## Issue reference

## Checklist

* [ ] Created/updated tests


yaron2 (dismissed):

magic-peach (commented):
Checked out the branch, ran tests/agents/test_schemas.py (18/18 pass) and the full tests/agents/ suite (417 pass), flake8 clean. The model itself is solid: sensible fields, clean serialization boundary, good tests proving caller_headers stays excluded while caller_claims is not.

One concern: the description justifies this by saying PropagateLineage() carries caller_claims into child workflows and that lifecycle plugins already populate it. Neither exists in this codebase right now, grep for PropagateLineage/lineage returns nothing, and caller_claims/caller_headers aren't referenced anywhere outside schemas.py and its tests. As written this lands as an inert schema field with no producer and no consumer.

Would suggest either landing this explicitly as a forward-looking schema addition (and dropping the PropagateLineage claim from the description until it exists), or holding it until the actual propagation wiring lands alongside it.

CasperGN (changes requested):
Thanks Sam — the model and the serialization split look good, and the tests pass locally. A few things before this can go in:

1. **Spoofing:** `TriggerAction` is what incoming pub/sub messages are validated against, so a publisher can send `caller_claims={"subject":"admin",...}` and it's accepted and written to signed history as verified identity. The framework has to drop any `caller_claims` that arrive with the message and only let a plugin set it (the same applies to `caller_headers`; `exclude=True` doesn't stop it being read in). A test that sends a message with forged claims and checks they're dropped would lock this in.
2. **Wiring:** agree with @magic-peach. Nothing writes or reads the field, and `agent_workflow` doesn't pass it into child or orchestration inputs. I'd rather it landed together with the plugin that fills it in and the propagation to children, or else drop the `PropagateLineage()` references from the description and the field docs.
3. **`claims_hash`:** a plain SHA-256 can be recomputed by whoever edits the claims, so it isn't tamper detection. Either use an HMAC or signature with a defined canonical form, or call it a fingerprint.
4. **PII:** `email`/`subject` become permanent in workflow history. Worth making `email` opt-in, or documenting it.
5. Nits: make `CallerClaims` frozen, add the docs PR link, and rebase onto main.

PR #693, opened 2026-07-10
fix(mcp): store empty dict when MCP prompt load fails

anxkhn:
# Description

`MCPClient._server_prompts` is a dictionary of per-server prompt dictionaries, but the prompt-loading failure path stored a list. Prompt accessors then called `.values()`, `.keys()`, or `.get()` on that list and raised `AttributeError`. A server without prompt support could also break aggregate prompt lookups across all servers.

This changes the failure value from `[]` to `{}`, matching the declared type and success path. Regression tests cover prompt-loading failure and success paths.

## Issue reference

No issue. This is a small, self-contained bug fix discovered while reviewing the MCP client.

## Checklist

* [x] Created/updated tests

No documentation PR is needed because this fixes an internal container type without changing the public API, configuration, or documented behavior.

## AGENTS.md Notes

- Adding the DCO sign-off requirement beside the commit-format section would help contributors avoid unsigned commits.


anxkhn:
The `E2E Tests (Ollama)` check failed at the "Install Ollama" step with `ollama: command not found`, which means the `curl -fsSL https://ollama.com/install.sh | sh` install step did not succeed on that runner. This is unrelated to this change: the diff only touches `dapr_agents/tool/mcp/client.py` (a one-character container-type fix) plus a unit test under `tests/tool/`, and it adds no integration tests that run in the Ollama e2e path. Lint, all build matrix jobs (3.11-3.14), schema checks, and DCO are green. Could a maintainer re-run the failed job? Happy to rebase if that helps re-trigger it.

sicoyle (approved):
amazing, thank you!!! 🚀

PR #694, opened 2026-07-10
fix(prompt): correct extract_input_variables return type to List[str]

anxkhn:
# Description

`Prompty.extract_input_variables` was annotated as returning `Tuple[List[str], List[str]]`, and its docstring described two lists. The implementation has one return statement that produces a flat, deduplicated `List[str]`. Its caller already treats the result as that list.

This aligns the annotation and docstring with the existing behavior, removes the unused `Tuple` import, and adds regression tests for the return contract and prompt-template input variables. Runtime behavior is unchanged.

## Issue reference

No issue. This is a small, self-contained type-contract correction.

## Checklist

* [x] Created/updated tests

No documentation PR is needed because this corrects an internal annotation and docstring without changing runtime behavior.

## AGENTS.md Notes

- Mentioning that `dapr_agents.prompt.*` is excluded from mypy checks would clarify why runtime contract tests remain useful there.

### DIFF
--- a/dapr_agents/prompt/prompty.py
+++ b/dapr_agents/prompt/prompty.py
@@ -21,7 +21,7 @@ from dapr_agents.prompt.base import PromptTemplateBase
 from dapr_agents.prompt.chat import ChatPromptTemplate
 from dapr_agents.prompt.string import StringPromptTemplate
 from dapr_agents.prompt.utils.prompty import PromptyHelper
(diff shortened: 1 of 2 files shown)


sicoyle (approved):
amazing, thank you for your well scoped and tested PRs @anxkhn. They make reviewing easy on my end which I super appreciate!! And a big thank you for your interest and contributions to Dapr Agents! 🙌 🚀

yaron2 (approved):

PR #696, opened 2026-07-10
fix(prompt): correct extract_variables return type to List[str]

anxkhn:
# Description

`StringPromptHelper.extract_variables` was annotated as returning `Dict[str, Any]`, but it delegates to f-string and Jinja extractors that both return `List[str]`. Its callers also use the result as a list of input variable names.

This corrects the annotation and docstring, updates the typing import, and adds tests covering both template formats, empty templates, unsupported formats, and the return annotation. Runtime behavior is unchanged.

## Issue reference

No issue. This is a small, self-contained type-contract correction.

## Checklist

* [x] Created/updated tests

No documentation PR is needed because this corrects an internal annotation and docstring without changing runtime behavior.

## AGENTS.md Notes

- Mentioning that `dapr_agents.prompt.*` is excluded from mypy checks would clarify why runtime contract tests remain useful there.

### DIFF
--- a/dapr_agents/prompt/utils/string.py
+++ b/dapr_agents/prompt/utils/string.py
@@ -19,7 +19,7 @@ from dapr_agents.prompt.utils.fstring import (
     render_fstring_template,
     extract_fstring_variables,
 )
-from typing import Any, Dict
+from typing import Any, List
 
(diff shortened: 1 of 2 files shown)


sicoyle (approved):
thank you!!

PR #698, opened 2026-07-13
perf(agents): bound DaprInfra etag cache to prevent unbounded growth

anxkhn:
`DaprInfra` caches etags by workflow state key so a `save_state` can reuse the etag loaded by a preceding `get_state`. Read-only state loads and purged instances could leave entries behind indefinitely in the long-lived agent object.

This change:

- bounds the cache with a 1024-entry `LRUCache`
- protects every cache mutation with a lock because workflow activities can execute concurrently in the threadpool
- evicts a cached etag when state is purged, including when deletion fails

LRU plus a fixed bound is the simplest policy here. It guarantees bounded memory without choosing an arbitrary expiry window. Eviction is safe because `save_state` already reloads an etag on a cache miss. The lock covers only short in-memory cache operations, not state-store I/O.

Tests use shared fixtures and a shared `load_with_etag` helper. They cover concurrent threadpool-backed reads, the size bound and LRU behavior, purge cleanup on success and failure, and the paired get/save fast path.

## Issue reference

No linked issue. This PR addresses an internal cache growth problem discovered in the existing workflow state implementation.

## Checklist

* [x] Created/updated tests
    * [ ] Created the [dapr/docs](https://github.com/dapr/docs) PR: not required

This is an internal performance and concurrency fix with no public API, feature, or configuration change, so a documentation PR is not required.

### Tests

- `uv run ruff format`
- `uv run flake8 dapr_agents tests --ignore=E501,F401,W503,E203,E704`
- `uv run mypy --config-file mypy.ini`
- `uv run pytest tests -m "not integration"` (964 passed, 15 skipped)

## AGENTS.md Notes

`AGENTS.md` was clear and sufficient for this change. The required checks, commit convention, license header, and documentation exception all applied directly.


dapr_agents/agents/components.py
```diff
@@ -134,8 +142,13 @@ def __init__(
...
         # (different workflow_instance_ids on the same DurableAgent object) from
-        # overwriting each other's cached etag.
-        self._etag_cache: Dict[str, Optional[str]] = {}
+        # overwriting each other's cached etag. It is bounded with LRU eviction
+        # because a long-lived DaprInfra serves an unbounded stream of instance
+        # ids and read-only get_state callers never pair with a save_state that
+        # would otherwise drain the entry.
+        self._etag_cache: LRUCache[str, Optional[str]] = LRUCache(
+            maxsize=_ETAG_CACHE_MAXSIZE
+        )
```
JeffreyJPZ, line 154: I think this would need to be made thread-safe as the current workflow client implementation is threadpool-backed, so we get race conditions if concurrent workflow activities are running (correct me if I'm wrong).

Also, given that `get_state` and `save_state` are almost always paired together in a single workflow activity, could a `TTLCache` with a reasonably short TTL be more appropriate? Could maybe bump the maxsize too.

JeffreyJPZ (changes requested):

tests/agents/test_components.py
```diff
@@ -0,0 +1,150 @@
...
+    a long-lived DaprInfra this would grow one entry per workflow_instance_id
+    forever, so the cache is bounded with LRU eviction.
+    """
+    infra = _make_infra()
+
+    def fake_load_with_etag(
+        *, key: str, default: Optional[Dict[str, Any]] = None, **_: Any
+    ) -> Tuple[Dict[str, Any], str]:
+        payload = default if isinstance(default, dict) else {}
+        return payload, f"etag-{key}"
```
JeffreyJPZ: Instead of redefining `fake_load_with_etag` in every test, could we make this a shared helper?

tests/agents/test_components.py
```diff
@@ -0,0 +1,150 @@
...
+
+    total = _ETAG_CACHE_MAXSIZE + 500
+    with patch.object(
+        infra.state_store, "load_with_etag", side_effect=fake_load_with_etag
+    ):
```
JeffreyJPZ: Can we have these as fixtures instead?

JeffreyJPZ (changes requested):

sicoyle:
@anxkhn thank you again for your contributions and interest in Dapr Agents 🚀 any chance you could address the feedback here please? I'd be happy to merge this afterwards 😇

dapr_agents/agents/components.py
anxkhn: Added a lock around each cache operation. I kept the fixed-size LRU because it guarantees the memory bound without choosing an expiry window; eviction safely falls back to the existing etag reload.

tests/agents/test_components.py
anxkhn: Moved this to one shared helper.

tests/agents/test_components.py
anxkhn: Converted the infrastructure and patched state-store setup to fixtures.

dapr_agents/agents/components.py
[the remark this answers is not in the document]
anxkhn: Added a lock around all LRU cache mutations and a concurrent threadpool regression test.

anxkhn:
Addressed the review feedback, rebased onto current `main`, and reran the full local gate.

JeffreyJPZ:
LGTM ✅

sicoyle (approved):
amazing!! This is very well scoped and clean - thank you for your contribution and interest in Dapr Agents!! 🙌 🚀

PR #733, opened 2026-08-23
feat(llm): bring Anthropic backend to parity with shared pipeline

### LINKED ISSUE #732: feat(llm): bring Anthropic backend to parity with shared LLM utilities and pipeline
## Summary

A cross-reference of the LLM backends in `dapr_agents/llm` (`openai`, `anthropic`, `dapr`, `litellm`, `nvidia`, `iflytek`, `huggingface`, `mistral`) reveals that while `AnthropicChatClient` supports standard chat completion, Prompty loading, and native `output_config` structured outputs, it is isolated from the core shared LLM utilities (`StreamHandler`, `ResponseHandler`, `StructureHandler`, `RequestHandler`).

This issue tracks the work to bring the Anthropic backend to full parity with the shared pipeline and other providers.

---

### 1. Inconsistent Provider Key (`"anthropic"` vs `"claude"`) in Tool Formatting
- **Issue**: `AnthropicClientBase` sets `self._provider = "anthropic"`, but `to_function_call_definition` and `validate_and_format_tool` in `dapr_agents/tool/utils/function_calling.py` only recognize `"claude"`.
- **Impact**: Calling `ToolHelper.format_tool(tool, tool_format=client.provider)` or `RequestHandler.process_params(..., llm_provider="anthropic")` raises `FunCallBuilderError("Unsupported format type: anthropic")`. `AnthropicChatClient.generate` currently works around this by hardcoding `tool_format="claude"` and bypassing `RequestHandler.process_params`.

### 2. Disconnection from `StreamHandler` and `providers.py`
- **Issue**: `"anthropic"` is omitted from `PROVIDERS_WITH_STREAMING` in `dapr_agents/llm/utils/providers.py`.
- **Impact**: Calling `StreamHandler.process_stream(..., llm_provider="anthropic")` raises `ValueError("Streaming not supported for provider: anthropic")`.

### 3. Streaming Return Type & Callback Signature Discrepancy
- **Issue**:
  - `ChatClientBase.generate(stream=True)` defines `Iterator[LLMChatCandidateChunk]` (implemented by `OpenAIChatClient`, `NVIDIAChatClient`, `HFHubChatClient`, `IFlytekChatClient`, `LiteLLMChatClient`).
  - `AnthropicChatClient.generate(stream=True)` returns `Iterator[LLMChatResponseChunk]`.
  - `AnthropicChatClient.generate` and `iter_stream` do not accept an `on_chunk: Optional[Callable[[LLMChatCandidateChunk], None]]` callback, unlike `ResponseHandler.process_response`.

### 4. Isolation from Shared `StructureHandler` and Lack of Defensive JSON Repair
- **Issue**:
  - `StructureHandler.extract_structured_response` explicitly checks known providers and raises `StructureError: Unsupported LLM provider: anthropic`.
  - `ResponseHandler.process_response` lacks an `anthropic` branch.
  - Anthropic's `parse_json_response` directly calls `StructureHandler.validate_response(block.text, target_format)` without `_repair_json_content` preprocessing, causing unnecessary validation failures if models emit markdown code fences (````json ... ````) or slight prose around JSON.

### 5. Extended Thinking / Reasoning Tokens in Streaming
- **Issue**: For reasoning models (e.g. Claude 3.7 Sonnet with `thinking={"type": "enabled", ...}`), non-streaming captures thinking blocks in `metadata["raw_content"]`, but streaming in `iter_stream` drops `thinking_delta`, `signature_delta`, and `redacted_thinking` events without surfacing them or recording them in metadata.

### 6. Multimodal (Image / Document) Input Normalization
- **Issue**: Generic message normalization in `RequestHandler.normalize_chat_messages` / `split_messages` does not automatically translate standard image dictionaries/URLs into Anthropic's content block format (`{"type": "image", "source": {"type": "base64", ...}}`).

---

## Proposed Changes / Acceptance Criteria

- [ ] **Tool Formatting Aliases**: Update `to_function_call_definition` and `validate_and_format_tool` in `dapr_agents/tool/utils/function_calling.py` to accept `"anthropic"` as an alias for `"claude"`.
- [ ] **Provider Registration**: Add `"anthropic"` to `PROVIDERS_WITH_STREAMING` in `dapr_agents/llm/utils/providers.py`.
- [ ] **Stream Handler Routing**: Add Anthropic stream processing to `StreamHandler.process_stream` and `ResponseHandler.process_response`.
- [ ] **Streaming Signature & Chunk Alignment**:
  - Add `on_chunk` callback support to `AnthropicChatClient.generate` and `iter_stream`.
  - Standardize chunk yields to `LLMChatCandidateChunk` (or support both cleanly across `ResponseHandler`).
- [ ] **Structured Output Integration & JSON Repair**:
  - Add `"anthropic"` to `StructureHandler.extract_structured_response`.
  - Integrate `_repair_json_content` into Anthropic's JSON response parsing to make structured output extraction resilient against markdown-wrapped responses.
- [ ] **Extended Thinking Streaming**: Handle `thinking_delta` and `signature_delta` SSE events in `iter_stream` to capture reasoning tokens in streaming metadata.
- [ ] **Unit Tests**: Add test coverage in `tests/llm/test_anthropic.py` and `tests/llm/test_structure_handler.py` for the unified streaming and tool formatting paths.

lindner:
# Description

This PR brings the Anthropic LLM backend (`AnthropicChatClient`) to full parity with the shared LLM pipeline (`StreamHandler`, `ResponseHandler`, `StructureHandler`, `RequestHandler`) and other LLM backends.

### Key Changes
1. **Tool Formatting Aliases**:
   - Updated `to_function_call_definition` and `validate_and_format_tool` in `dapr_agents/tool/utils/function_calling.py` to accept `"anthropic"` as an alias for `"claude"`.
   - Enabled `ToolHelper.format_tool(tool, tool_format=self.provider)` to succeed when `self.provider == "anthropic"`.
2. **Centralized Streaming Integration**:
   - Registered `"anthropic"` in `PROVIDERS_WITH_STREAMING` within `dapr_agents/llm/utils/providers.py`.
   - Added Anthropic streaming routing in `StreamHandler.process_stream` and `ResponseHandler.process_response`.
3. **Streaming Callbacks (`on_chunk`) & Event Handling**:
   - Added `on_chunk: Optional[Callable[[LLMChatResponseChunk], None]]` callback support in `AnthropicChatClient.generate` and `iter_stream`.
   - Captured `thinking_delta` and `signature_delta` SSE events in streaming chunk metadata for Claude 3.7+ reasoning models.
4. **Structured Outputs & Normalization**:
   - Added `"anthropic"` to `StructureHandler.extract_structured_response`.
   - Enabled `ResponseHandler.process_response` to normalize both raw and structured Anthropic responses.
5. **Multimodal Input Normalization**:
   - Added translation in `split_messages` for OpenAI-style `image_url` content blocks (both base64 data URIs and HTTP URLs) into Anthropic's native `image` content blocks.
6. **Unit Tests**:
   - Added unit tests in `tests/llm/test_anthropic.py` covering tool format aliases, `StreamHandler` streaming, `ResponseHandler` normalization, `on_chunk` callbacks, thinking deltas, and multimodal image block translation.

## Issue reference

Closes #732

## Checklist

* [x] Created/updated tests
* [x] Tested this change against the unit test suite (`uv run pytest tests -m "not integration"`)
    * Internal pipeline alignment and backward-compatible alias additions; no external API changes requiring a docs PR.

## AGENTS.md Notes

- Ensure `uv` executable location in user path (`~/.local/bin`) is documented or handled smoothly during environment initialization.

### DIFF
--- a/dapr_agents/llm/anthropic/chat.py
+++ b/dapr_agents/llm/anthropic/chat.py
@@ -23,14 +23,12 @@ from pydantic import BaseModel, Field, model_validator
 from dapr_agents.llm.anthropic.client import PROVIDER, AnthropicClientBase
 from dapr_agents.llm.anthropic.utils import (
     STRUCTURED_INJECTORS,
-    STRUCTURED_PARSERS,
     assert_json_output_supported,
-    iter_stream,
     split_messages,
-    to_llm_chat_response,
 )
 from dapr_agents.llm.chat import ChatClientBase
 from dapr_agents.llm.utils import RequestHandler
+from dapr_agents.llm.utils.response import ResponseHandler
 from dapr_agents.prompt.base import PromptTemplateBase
 from dapr_agents.prompt.prompty import Prompty
 from dapr_agents.tool import AgentTool
@@ -168,14 +166,34 @@ class AnthropicChatClient(AnthropicClientBase, ChatClientBase):
         logger.info("Calling Anthropic Messages API...")
         logger.debug(f"Anthropic request params: {params}")
         if stream:
-            return iter_stream(self.client, params, on_chunk=on_chunk)
+
+            def _stream_gen() -> Iterator[LLMChatResponseChunk]:
+                try:
+                    raw_stream = self.client.messages.create(stream=True, **params)
+                except Exception:
+                    logger.exception("Anthropic Messages API streaming call failed")
+                    raise
+                yield from ResponseHandler.process_response(
+                    response=raw_stream,
+                    llm_provider=self.provider,
+                    response_format=response_format,
+                    structured_mode=structured_mode,
+                    stream=True,
+                    on_chunk=on_chunk,
+                )
+
+            return _stream_gen()
+
         try:
             raw_resp = self.client.messages.create(**params)
         except Exception:
             logger.exception("Anthropic Messages API call failed")
             raise
 
-        if response_format is not None:
-            parse_structured_response = STRUCTURED_PARSERS[structured_mode]
-            return parse_structured_response(raw_resp, response_format)
-        return to_llm_chat_response(raw_resp)
+        return ResponseHandler.process_response(
+            response=raw_resp,
+            llm_provider=self.provider,
+            response_format=response_format,
+            structured_mode=structured_mode,
+            stream=False,
+        )
--- a/dapr_agents/llm/anthropic/chat.py
+++ b/dapr_agents/llm/anthropic/chat.py
@@ -14,7 +14,7 @@
 
 import logging
 import os
-from collections.abc import Iterable, Iterator
+from collections.abc import Callable, Iterable, Iterator
 from pathlib import Path
 from typing import Any, Literal
 
@@ -106,6 +106,7 @@ class AnthropicChatClient(AnthropicClientBase, ChatClientBase):
         response_format: type[BaseModel] | None = None,
         structured_mode: Literal["json", "function_call"] = "json",
         stream: bool = False,
+        on_chunk: Callable[[LLMChatResponseChunk], None] | None = None,
         **kwargs: Any,
     ) -> Iterator[LLMChatResponseChunk] | LLMChatResponse | BaseModel | list[BaseModel]:
         """Run one chat turn against the Anthropic Messages API.
@@ -131,6 +132,8 @@ class AnthropicChatClient(AnthropicClientBase, ChatClientBase):
         messages_normalized = RequestHandler.normalize_chat_messages(messages)
         system, messages_anthropic = split_messages(messages_normalized)
 
+        on_chunk_callback = on_chunk or kwargs.pop("on_chunk", None)
+
         prompty_params = (
             self.prompty.model.parameters.model_dump(exclude_none=True)
             if self.prompty
@@ -139,7 +142,10 @@ class AnthropicChatClient(AnthropicClientBase, ChatClientBase):
         extras = prompty_params | kwargs
         tools_effective = tools if tools is not None else extras.get("tools")
         tools_formatted = (
-            [ToolHelper.format_tool(t, tool_format="claude") for t in tools_effective]
+            [
+                ToolHelper.format_tool(t, tool_format=self.provider)
+                for t in tools_effective
+            ]
             if tools_effective
             else None
         )
@@ -164,7 +170,7 @@ class AnthropicChatClient(AnthropicClientBase, ChatClientBase):
         logger.info("Calling Anthropic Messages API...")
         logger.debug(f"Anthropic request params: {params}")
         if stream:
-            return iter_stream(self.client, params)
+            return iter_stream(self.client, params, on_chunk=on_chunk_callback)
         try:
             raw_resp = self.client.messages.create(**params)
         except Exception:
--- a/dapr_agents/llm/anthropic/chat.py
+++ b/dapr_agents/llm/anthropic/chat.py
@@ -132,8 +132,6 @@ class AnthropicChatClient(AnthropicClientBase, ChatClientBase):
         messages_normalized = RequestHandler.normalize_chat_messages(messages)
         system, messages_anthropic = split_messages(messages_normalized)
 
-        on_chunk_callback = on_chunk or kwargs.pop("on_chunk", None)
-
         prompty_params = (
             self.prompty.model.parameters.model_dump(exclude_none=True)
             if self.prompty
@@ -170,7 +168,7 @@ class AnthropicChatClient(AnthropicClientBase, ChatClientBase):
         logger.info("Calling Anthropic Messages API...")
         logger.debug(f"Anthropic request params: {params}")
         if stream:
-            return iter_stream(self.client, params, on_chunk=on_chunk_callback)
+            return iter_stream(self.client, params, on_chunk=on_chunk)
         try:
             raw_resp = self.client.messages.create(**params)
         except Exception:
--- a/dapr_agents/llm/anthropic/utils.py
+++ b/dapr_agents/llm/anthropic/utils.py
@@ -76,7 +76,11 @@ def _normalize_content_blocks(content: Any) -> Any:
                                 },
                             }
                         )
-                    except Exception:
+                    except Exception as exc:
+                        logger.warning(
+                            f"Failed to parse base64 data URI in content block; "
+                            f"passing block through unchanged: {exc}"
+                        )
                         blocks.append(item)
                 elif url.startswith(("http://", "https://")):
                     blocks.append(
@@ -364,6 +368,16 @@ def to_llm_chat_response(resp: Any) -> LLMChatResponse:
     return LLMChatResponse(results=[candidate], metadata=metadata)
 
 
+def _snapshot_stream_meta(meta: dict[str, Any]) -> dict[str, Any]:
+    """Create an isolated snapshot of stream metadata, making copies of mutable containers."""
+    snapshot = dict(meta)
+    if "thinking_blocks" in snapshot:
+        snapshot["thinking_blocks"] = list(snapshot["thinking_blocks"])
+    if "thinking_deltas" in snapshot:
+        snapshot["thinking_deltas"] = list(snapshot["thinking_deltas"])
+    return snapshot
+
+
 def process_anthropic_stream(
     raw_stream: Any,
     *,
@@ -372,10 +386,10 @@ def process_anthropic_stream(
 ) -> Iterator[LLMChatResponseChunk]:
     """Translate Anthropic SSE events into `LLMChatResponseChunk`s.
 
-    Each yield gets a `dict(meta)` snapshot so buffered consumers don't all
+    Each yield gets an isolated snapshot so buffered consumers don't all
     see the post-stream state.
     """
-    meta: dict[str, Any] = {"provider": PROVIDER, **(enrich_metadata or {})}
+    meta: dict[str, Any] = {**(enrich_metadata or {"provider": PROVIDER})}
 
     stream_iter = (
         raw_stream.__enter__() if hasattr(raw_stream, "__enter__") else raw_stream
@@ -405,7 +419,9 @@ def process_anthropic_stream(
                         index=0,
                         tool_calls=[tool_call_chunk],
                     )
-                    chunk = LLMChatResponseChunk(result=candidate, metadata=dict(meta))
+                    chunk = LLMChatResponseChunk(
+                        result=candidate, metadata=_snapshot_stream_meta(meta)
+                    )
                     if on_chunk:
                         on_chunk(chunk)
                     yield chunk
@@ -422,7 +438,9 @@ def process_anthropic_stream(
                         content=delta.text,
                         index=0,
                     )
-                    chunk = LLMChatResponseChunk(result=candidate, metadata=dict(meta))
+                    chunk = LLMChatResponseChunk(
+                        result=candidate, metadata=_snapshot_stream_meta(meta)
+                    )
                     if on_chunk:
                         on_chunk(chunk)
                     yield chunk
@@ -436,7 +454,9 @@ def process_anthropic_stream(
                         index=0,
                         tool_calls=[tool_call_chunk],
                     )
-                    chunk = LLMChatResponseChunk(result=candidate, metadata=dict(meta))
+                    chunk = LLMChatResponseChunk(
+                        result=candidate, metadata=_snapshot_stream_meta(meta)
+                    )
                     if on_chunk:
                         on_chunk(chunk)
                     yield chunk
@@ -452,7 +472,9 @@ def process_anthropic_stream(
                     candidate = LLMChatCandidateChunk(
                         finish_reason=event.delta.stop_reason
                     )
-                    chunk = LLMChatResponseChunk(result=candidate, metadata=dict(meta))
+                    chunk = LLMChatResponseChunk(
+                        result=candidate, metadata=_snapshot_stream_meta(meta)
+                    )
                     if on_chunk:
                         on_chunk(chunk)
                     yield chunk
--- a/dapr_agents/llm/anthropic/utils.py
+++ b/dapr_agents/llm/anthropic/utils.py
@@ -65,6 +65,11 @@ logger = logging.getLogger(__name__)
 # ---------------------------------------------------------------------------
 
 
+SUPPORTED_IMAGE_MEDIA_TYPES = frozenset(
+    {"image/jpeg", "image/png", "image/gif", "image/webp"}
+)
+
+
 def _normalize_content_blocks(content: Any) -> Any:
     """Translate OpenAI-style content blocks into Anthropic content blocks."""
     if not isinstance(content, list):
@@ -95,15 +100,27 @@ def _normalize_content_blocks(content: Any) -> Any:
                 if url.startswith("data:"):
                     try:
                         header, base64_data = url.split(",", 1)
-                        mime_type = header.split(";")[0].split(":")[1]
-                        normalized_block = {
-                            "type": "image",
-                            "source": {
-                                "type": "base64",
-                                "media_type": mime_type,  # type: ignore[typeddict-item]
-                                "data": base64_data,
-                            },
-                        }
+                        if ";base64" not in header:
+                            logger.warning(
+                                "Data URI is missing ';base64' encoding; "
+                                "passing block through unchanged"
+                            )
+                        else:
+                            mime_type = header.split(";")[0].split(":")[1]
+                            if mime_type not in SUPPORTED_IMAGE_MEDIA_TYPES:
+                                logger.warning(
+                                    f"Unsupported image media type '{mime_type}' for Anthropic; "
+                                    "passing block through unchanged"
+                                )
+                            else:
+                                normalized_block = {
+                                    "type": "image",
+                                    "source": {
+                                        "type": "base64",
+                                        "media_type": mime_type,  # type: ignore[typeddict-item]
+                                        "data": base64_data,
+                                    },
+                                }
                     except Exception:
                         logger.warning(
                             "Failed to parse base64 data URI in content block; "
@@ -270,36 +287,17 @@ def inject_function_call_request(
     params["tool_choice"] = {"type": "tool", "name": target_model.__name__}
 
 
-def _coerce_to_dict(obj: Any) -> dict[str, Any]:
-    """Convert SDK models, dataclasses, SimpleNamespaces, or dicts to plain dictionaries."""
-    if obj is None:
-        return {}
+def _dump_obj(obj: Any) -> dict[str, Any]:
+    """Serialize SDK model, dataclass, or object to dictionary without silent failure."""
     if isinstance(obj, dict):
         return obj
     if hasattr(obj, "model_dump") and callable(obj.model_dump):
-        try:
-            return obj.model_dump()
-        except Exception:
-            pass
-    if hasattr(obj, "to_dict") and callable(obj.to_dict):
-        try:
-            return obj.to_dict()
-        except Exception:
-            pass
+        return obj.model_dump()
     if dataclasses.is_dataclass(obj):
-        try:
-            return dataclasses.asdict(obj)
-        except Exception:
-            pass
+        return dataclasses.asdict(obj)
     if hasattr(obj, "__dict__"):
-        try:
-            return vars(obj)
-        except Exception:
-            pass
-    try:
-        return dict(obj)
-    except Exception:
-        return {}
+        return vars(obj)
+    return dict(obj)
 
 
 def parse_function_call_response(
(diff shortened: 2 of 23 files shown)


lindner:
Read the generated code, did a separate code review and it all looks good.  Will test this with a real backend shortly.

CasperGN:
Thank you @lindner for this!

Can you handle the merge conflict here, then I'll get it reviewed and merged

lindner:
> Thank you @lindner for this!
> 
> Can you handle the merge conflict here, then I'll get it reviewed and merged

Done!  Thanks for considering this

dapr_agents/llm/anthropic/utils.py
```diff
@@ -42,6 +67,65 @@
...
+                            },
+                        }
+                    )
+                else:
+                    blocks.append(item)
+            else:
+                blocks.append(item)
+        else:
+            blocks.append(item)
+    return blocks
```
JeffreyJPZ:
```suggestion
    for block in content:
        if isinstance(block, str):
            text_block: TextBlockParam = {"type": "text", "text": block}
            blocks.append(text_block)
            continue

        if not isinstance(block, dict):
            blocks.append(block)
            continue

        block_type = block.get("type")
        normalized_block: ContentBlockParam = block

        match block_type:
            case "image_url":
                image_url_dict = block.get("image_url", {})
                url = (
                    image_url_dict.get("url", "")
                    if isinstance(image_url_dict, dict)
                    else str(image_url_dict)
                )

                if url.startswith("data:"):
                    try:
                        header, base64_data = url.split(",", 1)
                        mime_type = header.split(";")[0].split(":")[1]
                        normalized_block: ImageBlockParam = {
                            "type": "image",
                            "source": {
                                "type": "base64",
                                "media_type": mime_type,  # type: ignore[typeddict-item]
                                "data": base64_data,
                            },
                        }
                    except Exception:
                        logger.warning(
                            "Failed to parse base64 data URI in content block; "
                            "passing block through unchanged",
                            exc_info=True,
                        )
                elif url.startswith(("http://", "https://")):
                    normalized_block: ImageBlockParam = {
                        "type": "image",
                        "source": {
                            "type": "url",
                            "url": url,
                        },
                    }
            case _:
                pass  # Pass through unknown block types unchanged

        blocks.append(normalized_block)

    return blocks
```

Control flow got a bit hard to follow, could we maybe cut down on the nesting w/ something like the above (checked out branch and verified it works)? Also changed the log to use `exc_info=True` to preserve the traceback and added some more type hints to bring it in line with other utils, but feel free to update it as you see fit.

EDIT: forgot to add a default case

dapr_agents/llm/anthropic/utils.py
```diff
@@ -13,10 +13,35 @@
...
+    RawMessageStartEvent,
+    RawMessageStreamEvent,
+    RedactedThinkingBlock,
+    SignatureDelta,
+    TextBlock,
+    TextBlockParam,
+    TextDelta,
+    ThinkingBlock,
+    ThinkingDelta,
+    ToolParam,
```
JeffreyJPZ:
```suggestion
```
nit: I don't think this is used anywhere

dapr_agents/llm/anthropic/utils.py
```diff
@@ -13,10 +13,35 @@
...
+    ToolResultBlockParam,
+    ToolUseBlock,
+    ToolUseBlockParam,
+    Usage,
```
JeffreyJPZ:
```suggestion
```
same as above

dapr_agents/llm/anthropic/utils.py
```diff
@@ -270,119 +363,238 @@ def parse_json_response(
...
-                        )
-                        candidate = LLMChatCandidateChunk(
-                            role="assistant",
-                            index=0,
-                            tool_calls=[tool_call_chunk],
-                        )
-                        yield LLMChatResponseChunk(
-                            result=candidate, metadata=dict(meta)
+        for event in stream_iter:
+            event_type = getattr(event, "type", None)
```
JeffreyJPZ: Same concern as with regular responses, referencing the OpenAI equivalent: https://github.com/dapr/dapr-agents/blob/3399f388a22f5c6921700d3d8f71247dc97f5880/dapr_agents/llm/openai/utils.py#L155-L164

dapr_agents/llm/anthropic/utils.py
```diff
@@ -270,119 +363,238 @@ def parse_json_response(
...
+    """Translate an Anthropic Message or SDK-compatible response to LLMChatResponse.
+
+    Note: Non-text/tool_use blocks (e.g. `thinking`) are stored in `metadata['raw_content']`.
+    """
     text_parts: list[str] = []
     tool_calls: list[ToolCall] = []
     raw_content: list[dict[str, Any]] = []
 
-    for block in resp.content or []:
+    content = getattr(resp, "content", None) or []
```
JeffreyJPZ: Not too familiar with the Anthropic SDK, would coercion be necessary here in case responses are not Pydantic models? Referencing the OpenAI equivalent:
https://github.com/dapr/dapr-agents/blob/3399f388a22f5c6921700d3d8f71247dc97f5880/dapr_agents/llm/openai/utils.py#L213-L224

JeffreyJPZ (commented):
First pass from me, some minor nits and some questions.

dapr_agents/llm/anthropic/utils.py
```diff
@@ -13,10 +13,35 @@
...
 
-from anthropic import Anthropic
+from anthropic import Anthropic, Stream
+from anthropic.types import (
+    Base64ImageSourceParam,
+    ContentBlock,
+    ImageBlockParam,
+    InputJSONDelta,
+    Message,
+    MessageParam,
```
JeffreyJPZ:
```suggestion
    ContentBlockParam,
    ImageBlockParam,
    InputJSONDelta,
    Message,
```

nit: could remove unused imports same as above, added `ContentBlockParam` for extra type hinting in `_normalize_content_blocks` but this is completely optional.

dapr_agents/llm/anthropic/utils.py
lindner: thanks, implemented.

dapr_agents/llm/anthropic/utils.py
lindner: fixed.

dapr_agents/llm/anthropic/utils.py
lindner: ditto

dapr_agents/llm/anthropic/utils.py
lindner: Thanks!    Added dictionary/object coercion for responses and SSE stream events (bringing parity with OpenAI utils).

dapr_agents/llm/anthropic/utils.py
lindner: done

dapr_agents/llm/anthropic/utils.py
lindner: ditto

lindner:
Updated.  Thanks for the review!

lindner:
I also tested a usage with a real ANTHROPIC key and it's working well.

pyproject.toml
```diff
@@ -61,7 +61,7 @@ dependencies = [
     "openinference-instrumentation>=0.1.0,<1.0.0",
     "openinference-semantic-conventions>=0.1.0,<1.0.0",
     "mistralai>=2.2.0",
-    "anthropic>=0.98.0,<1.0.0",
+    "anthropic>=0.98.0,<2.0.0",
```
JeffreyJPZ: If we support v1.0 it's gonna result in some breaking changes, mainly the removal of some
[request parameters](https://github.com/anthropics/anthropic-sdk-python/blob/main/MIGRATION.md#removed-deprecated-request-parameters) meaning that we can't just [forward them through anymore](https://github.com/dapr/dapr-agents/blob/a383dd47b11458783123c4e703cdc75ee661e756/dapr_agents/llm/anthropic/chat.py#L169). Maybe we could leave this to a followup?

pyproject.toml
```diff
@@ -131,6 +131,7 @@ dapr = "2026-05-21T00:00:00Z"
 dapr-ext-fastapi = "2026-05-21T00:00:00Z"
 dapr-ext-workflow = "2026-05-21T00:00:00Z"
 durabletask-dapr = "2026-05-21T00:00:00Z"
+anthropic = "2026-08-23T00:00:00Z"
```
JeffreyJPZ: Don't think this is necessary if we go with the above

JeffreyJPZ (changes requested):
One more pass, mainly concerned about the v0 -> v1 breaking changes but lgtm once that's addressed.

@CasperGN can you do a final review?

lindner:
@JeffreyJPZ thanks! Reverted the Anthropic dependency constraint back to `<1.0.0` and removed the exclude-newer package entry here. Created #747 stacked on top of this branch for the Anthropic v1 upgrade as a follow-up.

pyproject.toml
lindner: moved back to <1.0.0 for this PR.

pyproject.toml
lindner: ack

lindner:
Updates?   Have a lot more Anthropic things I'd like to get in so we can move off of `claude-agent-sdk`....

dapr_agents/llm/anthropic/utils.py
```diff
@@ -10,13 +10,36 @@
 # See the License for the specific language governing permissions and
```
sicoyle, line 10: can you add a test file to cover these changes pls?

dapr_agents/llm/anthropic/utils.py
```diff
@@ -186,14 +270,62 @@ def inject_function_call_request(
     params["tool_choice"] = {"type": "tool", "name": target_model.__name__}
 
 
+def _coerce_to_dict(obj: Any) -> dict[str, Any]:
```
sicoyle: any way we can drop this? The SDK always hands back pydantic models, so model_dump() and attribute access covers real input. My worry here is the pass on an exception so a malformed response quietly becomes {}.

dapr_agents/llm/anthropic/utils.py
```diff
@@ -270,119 +411,302 @@ def parse_json_response(
...
+        or (getattr(resp, "content", None) if not isinstance(resp, dict) else None)
+        or []
+    )
+
+    for raw_block in content:
+        block = _coerce_to_dict(raw_block)
+        raw_content.append(block)
+
+        block_type = block.get("type") or getattr(raw_block, "type", None)
+        if isinstance(raw_block, TextBlock) or block_type == "text":
```
sicoyle: this is guarded three ways (isinstance, string compare, getattr) and for real SDK blocks they can't disagree. The old block.type version was doing the job — any input shape you're protecting against here that I'm not thinking of?

dapr_agents/llm/anthropic/utils.py
```diff
@@ -270,119 +411,302 @@ def parse_json_response(
...
-                        candidate = LLMChatCandidateChunk(
-                            role="assistant",
-                            index=0,
-                            tool_calls=[tool_call_chunk],
-                        )
-                        yield LLMChatResponseChunk(
-                            result=candidate, metadata=dict(meta)
-                        )
+        for event in stream_iter:
+            evt_dict = _coerce_to_dict(event)
```
sicoyle: this means a full recursive model_dump() per sse event which is costly, so once per token delta. Before we just read event.type. Can we go back to that on the hot path pls?

dapr_agents/llm/anthropic/utils.py
```diff
@@ -270,119 +411,302 @@ def parse_json_response(
...
+    on_chunk: Callable[[LLMChatResponseChunk], None] | None = None,
 ) -> Iterator[LLMChatResponseChunk]:
     """Translate Anthropic SSE events into `LLMChatResponseChunk`s.
 
-    Each yield gets a `dict(meta)` snapshot so buffered consumers don't all
+    Each yield gets an isolated snapshot so buffered consumers don't all
     see the post-stream state.
     """
-    meta: dict[str, Any] = {"provider": PROVIDER}
+    meta: dict[str, Any] = {**(enrich_metadata or {"provider": PROVIDER})}
```
sicoyle: if a caller forgets the provider key we lose this entirely....
```suggestion
    meta: dict[str, Any] = {**(enrich_metadata or {"provider": PROVIDER, **(enrich_metadata or {})}
```

dapr_agents/llm/anthropic/utils.py
```diff
@@ -42,6 +65,66 @@
...
+
+        match block_type:
+            case "image_url":
+                image_url_dict = block.get("image_url", {})
+                url = (
+                    image_url_dict.get("url", "")
+                    if isinstance(image_url_dict, dict)
+                    else str(image_url_dict)
+                )
+                if url.startswith("data:"):
```
sicoyle, line 100: might be worth checking for ;base64 in the header too — a data:image/svg+xml,<svg…> URI is URL-encoded, not base64, and we'd forward the escaped text as base64 data. Also pls validate media_type against the four types Anthropic accepts.

dapr_agents/llm/anthropic/utils.py
```diff
@@ -270,119 +411,302 @@ def parse_json_response(
...
+                    meta.setdefault("thinking_deltas", []).append(thinking_val or "")
+                elif (
+                    isinstance(delta, SignatureDelta) or delta_type == "signature_delta"
+                ):
+                    sig_val = (
+                        delta_dict.get("signature")
+                        if "signature" in delta_dict
+                        else getattr(delta, "signature", None)
+                    )
+                    meta["thinking_signature"] = sig_val
```
sicoyle, line 712: if there are multiple thinking blocks only the last sig survives. Is that intentional?

dapr_agents/llm/utils/stream.py
```diff
@@ -36,20 +36,20 @@ class StreamHandler:
 
     @staticmethod
     def process_stream(
-        stream: Iterator[ChatCompletionChunk],
+        stream: Any,
```
sicoyle:
```suggestion
        stream: Iterator[Any],
```

dapr_agents/llm/utils/stream.py
```diff
@@ -68,5 +68,13 @@ def process_stream(
                 enrich_metadata={"provider": llm_provider},
                 on_chunk=on_chunk,
             )
+        elif llm_provider in ("anthropic", "claude"):
```
sicoyle, line 70: we handle claude here but PROVIDERS_WITH_STREAMING only lists anthropic. Can we either read the constant here or add claude to it pls so the two don't drift?

dapr_agents/llm/utils/stream.py
```diff
@@ -36,20 +36,20 @@ class StreamHandler:
 
     @staticmethod
     def process_stream(
-        stream: Iterator[ChatCompletionChunk],
+        stream: Any,
         llm_provider: str,
-        on_chunk: Optional[Callable],
-    ) -> Iterator[LLMChatCandidateChunk]:
+        on_chunk: Optional[Callable] = None,
+    ) -> Iterator[Union[LLMChatCandidateChunk, LLMChatResponseChunk]]:
```
sicoyle:
```suggestion
    ) -> Iterator[LLMChatResponseChunk]:
```
process_openai_stream and process_hf_stream are both annotated Iterator[LLMChatCandidateChunk] but actually yield LLMChatResponseChunk. I'd narrow this to Iterator[LLMChatResponseChunk] and fix those two annotations while we're here pls

dapr_agents/llm/anthropic/chat.py
```diff
@@ -164,7 +168,7 @@ def generate(
         logger.info("Calling Anthropic Messages API...")
         logger.debug(f"Anthropic request params: {params}")
         if stream:
-            return iter_stream(self.client, params)
+            return iter_stream(self.client, params, on_chunk=on_chunk)
```
sicoyle: question on scope here pls... generate calls iter_stream and to_llm_chat_response directly and never go through the ResponseHandler and StreamHandler, so the new registrations aren't reachable from the client itself. Is this dual path deliberate, or should the client delegate for consistency here?

tests/llm/test_anthropic.py
```diff
@@ -970,3 +970,511 @@ def test_anthropic_generate_requires_messages_or_input_data(mock_anthropic_class
...
+    assert msg.tool_calls[0].id == "tool_call_1"
+    assert msg.tool_calls[0].function.name == "lookup_weather"
+    assert json.loads(msg.tool_calls[0].function.arguments) == {"city": "Seattle"}
+
+
+def test_anthropic_process_stream_with_dict_events():
+    """process_anthropic_stream handles plain dict event streams."""
+    from dapr_agents.llm.anthropic.utils import process_anthropic_stream
+
+    dict_events = [
```
sicoyle, line 1340: not something this PR introduced — the whole tests/llm/ directory uses SimpleNamespace fixtures. But since this PR adds isinstance checks against anthropic.types models, those branches now never execute in CI. Given your other PR is going to move us across a major SDK version, can you pls add one or two fixtures built from real models here. That's the thing that would actually catch v1 shape changes.

sicoyle (changes requested):
hiiii! Thank you so much for your interest in our open source community, and for all the issues and contributions you've been making — we really do appreciate them 🤗 I just got back from my honeymoon and am working through my review backlog everywhere, so thanks for your patience on my delayed feedback here!

sicoyle:
> Updates? Have a lot more Anthropic things I'd like to get in so we can move off of claude-agent-sdk....

I am happy to help move things along here now that I am back too🙌  Pls feel free to ping me when this is ready for rereview.

dapr_agents/llm/anthropic/utils.py
lindner: Added `tests/llm/test_anthropic_utils.py` covering all the functions and edge cases in `utils.py`.

dapr_agents/llm/anthropic/utils.py
lindner: Ah yes, dropped `_coerce_to_dict` and it's silent exception swallowing.

dapr_agents/llm/anthropic/utils.py
lindner: Simplified to just check `block.type` directly (with a dict fallback).

dapr_agents/llm/anthropic/utils.py
lindner: oops, reverted to reading attributes directly on the streaming hot path.

dapr_agents/llm/anthropic/utils.py
lindner: okay, switched ot `meta = {"provider": PROVIDER, **(enrich_metadata or {})}`

dapr_agents/llm/utils/stream.py
lindner: done

dapr_agents/llm/anthropic/utils.py
lindner: sure, added ;base64 check and restricted to the 4 supported media types (jpeg, png, gif, webp).

dapr_agents/llm/anthropic/utils.py
lindner: now collecting all signatures into a thinking_signatures list while keeping thinking_signature as the latest.

dapr_agents/llm/utils/stream.py
lindner: moved this to  `PROVIDERS_WITH_STREAMING`

dapr_agents/llm/utils/stream.py
lindner: Updated the return type to Iterator[LLMChatResponseChunk] everywhere

dapr_agents/llm/anthropic/chat.py
lindner: Makes sense, routed generate through `ResponseHandler.process_response` for both streaming and non-streaming.

tests/llm/test_anthropic.py
lindner: Added TestRealSDKModelFixtures using real `anthropic.types` classes for both chat responses and stream events!

lindner:
@sicoyle congrats on your marriage/honeymoon!   Addressed a lot of the comments.  I'm getting the feel of the system a little better now.  Have more to come!

sicoyle (approved):
LGTM - thank youuuu!

sicoyle:
> @sicoyle congrats on your marriage/honeymoon! Addressed a lot of the comments. I'm getting the feel of the system a little better now. Have more to come!

Thank youuu! 👰‍♀️ 😊  I'm glad things are clicking more. Pls feel free to tag me anytime. I'm happy to help however I can 😄

PR #747, opened 2026-09-03
chore(deps): upgrade anthropic dependency to <2.0.0

lindner:
# Description

> **Note**: This PR is stacked on top of #733 (`feat/anthropic-llm-parity`). Please review/merge #733 first.

This PR upgrades the `anthropic` dependency constraint from `<1.0.0` to `<2.0.0` (enabling Anthropic Python SDK v1.x) and pins cutoff timestamp to lock 1.0.0 in `uv.lock`.

### Key Changes
- Expanded `anthropic` dependency constraint to `>=0.98.0,<2.0.0` in `pyproject.toml`
- Added `anthropic = "2026-08-23T00:00:00Z"` in `[tool.uv.exclude-newer-package]`
- Updated `uv.lock`

## Issue reference

Follow-up stacked on top of #733

## Checklist

* [x] Created/updated tests
* [x] Tested this change against the unit test suite (`uv run pytest tests -m "not integration"`)
  * Dependency constraint update only; no documentation changes required.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,7 +61,7 @@ dependencies = [
     "openinference-instrumentation>=0.1.0,<1.0.0",
     "openinference-semantic-conventions>=0.1.0,<1.0.0",
     "mistralai>=2.2.0",
-    "anthropic>=0.98.0,<1.0.0",
+    "anthropic>=0.98.0,<2.0.0",
     "opentelemetry-instrumentation-grpc>=0.60b1,<0.61",
     "opentelemetry-exporter-zipkin-json>=1.36.0",
     "opentelemetry-instrumentation-httpx>=0.60b1,<0.61",
@@ -131,6 +131,7 @@ dapr = "2026-05-21T00:00:00Z"
 dapr-ext-fastapi = "2026-05-21T00:00:00Z"
 dapr-ext-workflow = "2026-05-21T00:00:00Z"
 durabletask-dapr = "2026-05-21T00:00:00Z"
+anthropic = "2026-08-23T00:00:00Z"
 
 [project.urls]
 Documentation = "https://github.com/dapr/docs"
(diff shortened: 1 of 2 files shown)


JeffreyJPZ:
> If we support v1.0 it's gonna result in some breaking changes, mainly the removal of some [request parameters](https://github.com/anthropics/anthropic-sdk-python/blob/main/MIGRATION.md#removed-deprecated-request-parameters) meaning that we can't just [forward them through anymore](https://github.com/dapr/dapr-agents/blob/a383dd47b11458783123c4e703cdc75ee661e756/dapr_agents/llm/anthropic/chat.py#L169). Maybe we could leave this to a followup?

As mentioned in #733, we won't be able to forward `temperature`, `top_p`, `top_k` directly to the Anthropic client so we can't just upgrade as-is. Also worth noting that (edit: the big closed-weight) LLM providers are generally moving away from sampling parameters with newer reasoning models, so anything we do here should probably be applied to the others + the Dapr chat client eventually (I think there's been an effort to [remove models from the component spec](https://github.com/dapr/proposals) and move it into the conversation API instead, but for now this is a [known component-level issue](https://github.com/dapr/components-contrib/issues/4533)).

I think the best play for now is to do some model-aware branching. For older models we should check for those params and pass them via `extra_body` like so (taken from the Anthropic v1 migration guide):
```python
# Before
client.messages.create(..., model="claude-sonnet-4-6", temperature=0.2)

# After
client.messages.create(..., model="claude-sonnet-4-6", extra_body={"temperature": 0.2})
```

Then for newer models we can maybe emit a deprecation warning for those params and just ignore them, that way the client API can remain consistent for now, WDYT?

CasperGN (approved):

CasperGN:
Thank you @lindner

PR #749, opened 2026-09-04
fix: register newly added pub/sub routes on repeated wire calls

### LINKED ISSUE #707: Second register_routes/subscribe call silently skips pub/sub registration due to instance-wide _wired_pubsub guard
After the first successful pub/sub wiring, an `AgentRunner` will silently skip registering pub/sub routes on any later call, even when the agent's pubsub config has changed or a different config is passed. Messages Dapr accepts for the newly configured topics are then never handled, with no error and no log line to explain why.

**Where**

`dapr_agents/workflow/runners/agent.py`, in `_wire_pubsub_routes`:

https://github.com/dapr/dapr-agents/blob/aa5b87889223662e703415988359ed54a0bcede3/dapr_agents/workflow/runners/agent.py#L430-L461

```python
specs = self._build_pubsub_specs(agent, config)
if not specs:
    return

self._ensure_dapr_client()
if self._wired_pubsub or self._dapr_client is None:
    return
# ... register_message_routes(...) ...
self._wired_pubsub = True
```

`self._wired_pubsub` is set once (L461) and never reset. Both public entry points call this method: `register_routes` (L339) and `subscribe` (L508). So the specs are recomputed on every call, but the early return at L435 means only the first call actually registers routes.

**How it manifests**

1. Wire an agent whose `pubsub` config has `agent_topic` but no `broadcast_topic`, via `register_routes` or `subscribe`. Routes register, `_wired_pubsub` becomes `True`.
2. Set `agent.pubsub.broadcast_topic` to a valid topic and call `subscribe` (or `register_routes`) again.
3. `_build_pubsub_specs` now returns both the direct and broadcast subscriptions, but the guard at L435 returns early, so `register_message_routes` is never called again.
4. The sidecar subscription metadata still lists only the original topic. Broadcast messages are accepted by Dapr but never delivered to the agent, with no error.

**Expected behavior**

A later call with a changed set of specs should register the new routes. One option is to track the currently registered specs and re-register (and close the stale closers in `self._pubsub_closers`) when they change, rather than returning unconditionally once `_wired_pubsub` is set. If re-wiring is intentionally unsupported, calling it a second time with different specs should at least emit a clear warning instead of silently doing nothing.

**Additional context**

This surfaced while reviewing #699 (the fix for #691, "hosting an agent with a pubsub config fails when broadcast topic is not configured"). The `_wired_pubsub` guard predates and is untouched by that change, but the broadcast-topic scenario makes the re-wire gap easy to hit.

Found while running [Ito](https://ito.ai) (AI code review, free for open source) against recently merged PRs. Full analysis: https://app.ito.ai/share/2957941b-bab2-45db-aee1-12f0f92bc2b7.

magic-peach:
# Description

AgentRunner._wire_pubsub_routes only registered pub/sub routes on the first call for a given runner instance. The guard `self._wired_pubsub` was set to True after the first successful registration and never reset, so a later call to register_routes or subscribe with a broader pubsub config for example a broadcast_topic added after the agent was already wired once was silently skipped in full. The sidecar subscription metadata kept only the original topics and messages sent to the newly configured topic were accepted by Dapr but never delivered to the agent, with no error or log line explaining why.

This PR tracks which (pubsub_name, topic) pairs have already been registered on the runner instance and only registers the specs that are new on each call, instead of skipping registration entirely once anything has been wired. unwire_pubsub now also forgets the previously wired topics so a rewire after teardown still works as it did before this change.

## Issue reference

Please reference the issue this PR closes: #707

## Checklist

* [x] Created/updated tests

Notes:
- Tests: added tests/workflow/test_agent_runner_pubsub.py with a regression test that fails against the pre-fix code (verified locally) and passes with the fix, plus a no-op case and an unwire/rewire case.
- Quickstarts: not run in this environment since it requires a live Dapr sidecar; the fix is covered by the new unit tests instead.
- Documentation: this is an internal bug fix with no change to any public API or configuration surface, so no dapr.io documentation update is needed for it.

### DIFF
--- a/dapr_agents/workflow/runners/agent.py
+++ b/dapr_agents/workflow/runners/agent.py
@@ -165,6 +165,11 @@ class AgentRunner(WorkflowRunner):
             client_factory=client_factory,
         )
         self._default_http_paths: set[str] = set()
+        # (pubsub_name, topic) pairs already registered via _wire_pubsub_routes,
+        # so a later call with additional specs (e.g. a broadcast topic added
+        # after the first wiring) still registers what's new instead of being
+        # skipped entirely by the instance-wide `_wired_pubsub` guard.
+        self._wired_pubsub_topics: set[tuple[str, str]] = set()
 
         # In-memory store of managed agents - used for handling shutdown
         self._managed_agents: List[DurableAgent] = []
@@ -802,7 +807,15 @@ class AgentRunner(WorkflowRunner):
             return
 
         self._ensure_dapr_client()
-        if self._wired_pubsub or self._dapr_client is None:
+        if self._dapr_client is None:
+            return
+
+        new_specs = [
+            spec
+            for spec in specs
+            if (spec.pubsub_name, spec.topic) not in self._wired_pubsub_topics
+        ]
+        if not new_specs:
             return
 
         try:
@@ -815,7 +828,7 @@ class AgentRunner(WorkflowRunner):
             deduper = None
 
         closers = register_message_routes(
-            routes=specs,
+            routes=new_specs,
             dapr_client=self._dapr_client,
             delivery_mode=delivery_mode,
             queue_maxsize=queue_maxsize,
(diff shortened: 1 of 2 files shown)


CasperGN (approved):
Thank you @magic-peach!

PR #752, opened 2026-09-05
feat: forward LLM response metadata to after_llm_call hooks

### LINKED ISSUE #731: DurableAgent discards Anthropic token-usage metadata — not exposed via AgentRunner.run() or any hook
## Summary

`LLMChatResponse.metadata["usage"]` (input/output/cache-read/cache-creation token counts) is computed by the Anthropic chat client but is silently dropped inside `DurableAgent`'s workflow loop before it reaches either the `AgentRunner.run()` return value or the `after_llm_call` hook. There's currently no supported way for calling code to capture per-call token usage for a DurableAgent run.

## Version

dapr-agents 1.0.5

## Where it's dropped
- `dapr_agents/llm/anthropic/utils.py:296-307` (`to_llm_chat_response()`) correctly builds `metadata["usage"] = resp.usage.model_dump()` — the full   Anthropic `Usage` object is present at this point.
- `dapr_agents/agents/durable.py`, the `call_llm` workflow activity (~line 2075-2139), calls `self.llm.generate(**generate_kwargs)` and then only reads `response.get_message()` (line ~2109) to build the returned assistant-message dict. `response.metadata` (containing `usage`) is never  read again and is not included in what `call_llm` returns.
- The `after_llm_call` hook (`dapr_agents/hooks.py:252-255`) is documented to receive `(ctx, assistant_message_dict)` — at the call site (`durable.py:2120-2129`) the second argument is `dict(assistant_message)`, the message only. There is no hook that receives the full `LLMChatResponse` or its `metadata`.
- `AgentRunner.run()` (`dapr_agents/workflow/runners/agent.py:114-167`) awaits the workflow and returns only the serialized final message string.

## Impact
There's no supported way to do per-call cost/usage accounting for a DurableAgent run without monkey-patching/wrapping the LLM client's `generate()` method yourself before constructing the agent. This is awkward and fragile compared to just having the data flow through.

Notably, the usage data clearly *is* recoverable at the `generate()` boundary — `DaprAgentsInstrumentor`'s OTel wrapper (`dapr_agents/observability/wrappers/llm.py:459-530`) already extracts `usage` from the response metadata into span attributes
(`gen_ai.usage.input_tokens` etc.) via the same class-level wrap. So the data exists right up until `call_llm` throws it away — it's just not
propagated to callers who aren't running the OTel instrumentor and inspecting spans.

## Suggested fix
Either:
1. Include `metadata`/`usage` alongside the assistant message in `call_llm`'s return value (and therefore in `AgentRunner.run()`'s
result), or
2. Pass the full `LLMChatResponse` (or at least its `metadata`) to `after_llm_call` hooks instead of just the message dict, so consumers
can capture usage without needing OTel wired up.

magic-peach:
# Description

LLMChatResponse.metadata built by each provider client for example token usage counts on the Anthropic backend was silently dropped inside the call_llm workflow activity. There was no supported way for calling code to capture per call token usage for a DurableAgent run without wrapping or monkeypatching the LLM client's generate method.

The after_llm_call hook already receives a payload dict built for context (a copy of the generate kwargs). This PR adds the LLM response's metadata to that payload under a new response_metadata key whenever the call produced an LLMChatResponse, so a hook can read ctx.payload["response_metadata"]["usage"] for per call usage accounting without needing to touch the LLM client at all. The persisted assistant message and the chat history sent back to the model are unchanged.

## Issue reference

Please reference the issue this PR closes: #731

## Checklist

* [x] Created/updated tests

Notes:
- Tests: added tests/agents/durableagent/test_llm_hooks.py::TestAfterLLMCallResponseMetadata, covering both the case where response_metadata is forwarded and the case where the LLM returns something other than an LLMChatResponse (payload must not gain the key). Verified the new test fails against the pre-fix code and passes with the fix.
- Quickstarts: not run in this environment since it requires a live Dapr sidecar.
- Documentation: the after_llm_call docstring in dapr_agents/hooks.py was updated to describe the new payload key, but before_llm_call/after_llm_call hooks are not currently documented anywhere on the dapr.io docs site, so there is no existing page to extend for this. If the hooks system gets a dedicated docs.dapr.io page in the future, this new payload key would need to be mentioned there as a follow-up.

### DIFF
--- a/dapr_agents/agents/durable.py
+++ b/dapr_agents/agents/durable.py
@@ -2372,6 +2372,7 @@ class DurableAgent(AgentBase):
             # `response_format`, or `tool_choice` from generate_kwargs.
             generate_kwargs = {**generate_kwargs, **before_llm_decision.payload}
 
+        response: Any = None
         if synthesized_message is not None:
             assistant_message = synthesized_message
         else:
@@ -2478,6 +2479,11 @@ class DurableAgent(AgentBase):
             after_payload: Dict[str, Any] = dict(generate_kwargs)
             if "messages" in after_payload:
                 after_payload["messages"] = list(after_payload["messages"])
+            # Surface provider response metadata (token usage, finish_reason, …)
+            # so hooks can do per-call usage accounting without having to
+            # wrap/monkeypatch the LLM client themselves.
+            if isinstance(response, LLMChatResponse) and response.metadata:
+                after_payload["response_metadata"] = dict(response.metadata)
             after_ctx = LLMHookContext(payload=after_payload)
             for hook in self._hooks.after_llm_call:
                 result = hook(after_ctx, dict(assistant_message))
--- a/dapr_agents/hooks.py
+++ b/dapr_agents/hooks.py
@@ -260,4 +260,6 @@ class Hooks:
     after_llm_call: List[AfterLLMHook] = field(default_factory=list)
     """called after every llm response. return Mutate(payload=<assistant_message dict>)
     to replace the message that gets persisted and returned. Skip / Deny / RequireApproval
-    are no-ops on this slot (the LLM has already produced output)."""
+    are no-ops on this slot (the LLM has already produced output). ctx.payload includes
+    "response_metadata" (the provider's LLMChatResponse.metadata, e.g. token usage) when
+    the call produced one."""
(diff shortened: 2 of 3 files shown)


CasperGN (approved):
Thank you @magic-peach!

PR #753, opened 2026-09-07
fix(memory): return most recent messages, not oldest, from get_messages

magic-peach:
# Description

add_message always appends, so stored messages in ConversationDaprStateMemory are ordered oldest to newest. get_messages sliced with raw_messages[:limit], which returns the oldest messages instead of the most recent ones. Once a conversation grows past the limit, this means it forever returns the same first N messages and never surfaces anything new. This slices from the end instead so the limit keeps the most recent messages.

Found by reading the code, not tied to an existing issue.

## Issue reference

Please reference the issue this PR closes: none, self discovered

## Checklist

* [x] Created/updated tests
    * [ ] Created the dapr/docs PR: not applicable, no user facing behavior to document beyond the bug fix itself

### DIFF
--- a/dapr_agents/memory/daprstatestore.py
+++ b/dapr_agents/memory/daprstatestore.py
@@ -201,7 +201,7 @@ class ConversationDaprStateMemory(MemoryBase):
         if response and hasattr(response, "data") and response.data:
             raw_messages = json.loads(response.data)
             if raw_messages:
-                messages = raw_messages[:limit]
+                messages = raw_messages[-limit:]
                 logger.info(
                     f"Retrieved {len(messages)} messages for workflow instance {workflow_instance_id}",
                 )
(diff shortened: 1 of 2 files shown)


dapr_agents/memory/daprstatestore.py
```diff
@@ -201,7 +201,7 @@ def get_messages(
         if response and hasattr(response, "data") and response.data:
             raw_messages = json.loads(response.data)
             if raw_messages:
-                messages = raw_messages[:limit]
+                messages = raw_messages[-limit:]
```
sicoyle:
```suggestion
                messages = raw_messages[-limit:] if limit > 0 else []
```

otherwise if limit = 0 then it would now return the entire list

sicoyle (changes requested):
one minor comment and then this is g2g. Thank you for your interest in the Dapr Open Source community and the Dapr Agents project. I really appreciate your contributions (here and python sdk 🤗 ) and the fact that they are easy to review and well tested - you rock 🚀

magic-peach:
@sicoyle I made the changes could you please merge it if nothing else is needed?

magic-peach:
@CasperGN could you please take a look at this pr and let me know if anything else is needed from my side?

sicoyle (approved):
amazing - thank youuu!!!

PR #755, opened 2026-09-08
fix(executors): fix missing package probe in LocalCodeExecutor

### LINKED ISSUE #754: LocalCodeExecutor's missing-package probe always reports every import as missing
## What happened

`LocalCodeExecutor._get_missing_packages` (`dapr_agents/executors/local.py`) checks whether each import extracted from a code snippet is already installed by spawning `python3` with a probe script. The probe script is passed as a single `argv` string in heredoc form:

```python
proc = await asyncio.create_subprocess_exec(
    str(python),
    "- <<PY\nimport importlib.util, sys;"
    f"sys.exit(importlib.util.find_spec('{pkg}') is None)\nPY",
    ...
)
```

`create_subprocess_exec` never goes through a shell, so this heredoc syntax is not interpreted, it is passed to the Python interpreter as a literal argument. The interpreter treats the leading `-` as an option flag it does not recognize.

## Impact

Every extracted import, including stdlib modules, is reported as missing. This forces a redundant `pip install` attempt on every code execution, and can hang or fail hard if one of those installs fails (for example, trying to `pip install sys`).

## How to reproduce

Run `LocalCodeExecutor._get_missing_packages` (or the underlying `probe` coroutine) against any installed package name, including a stdlib module like `sys`. The probe subprocess exits non-zero (or hangs, depending on platform) instead of exiting 0 for an installed package.

## Suggested fix

Use `python3 -c "<script>"` instead of the heredoc-as-argv form, matching the `-c` invocation style already used elsewhere in the same file.

magic-peach:
# Description

`_get_missing_packages` ran `python3` with a heredoc string as a single `argv` entry instead of a shell, so the probe subprocess always failed (or hung) instead of exiting 0 for an installed package. Every extracted import, including stdlib modules like `sys`, was always reported as missing. That forces an unnecessary `pip install` attempt for every import on every execution, and any of those failing (for example trying to `pip install` a stdlib name) raises and aborts the whole snippet run.

Use `python3 -c` directly instead of the shell heredoc string, matching how `-c` is used everywhere else in this file.

## Issue reference

Please reference the issue this PR closes: #754

## Checklist

* [x] Created/updated tests

Notes:
- Tests: added tests/executors/test_local.py (none existed before), confirmed both cases fail against the pre-fix code (one hangs rather than failing cleanly) and pass against the fix.
- Quickstarts: not run in this environment since it requires a live Dapr sidecar; the fix is covered by the new unit tests instead.
- Documentation: this is an internal bug fix with no change to any public API or configuration surface, so no dapr.io documentation update is needed for it.

### DIFF
--- a/dapr_agents/executors/local.py
+++ b/dapr_agents/executors/local.py
@@ -241,8 +241,8 @@ class LocalCodeExecutor(CodeExecutorBase):
         async def probe(pkg: str) -> str | None:
             proc = await asyncio.create_subprocess_exec(
                 str(python),
-                "- <<PY\nimport importlib.util, sys;"
-                f"sys.exit(importlib.util.find_spec('{pkg}') is None)\nPY",
+                "-c",
+                f"import importlib.util, sys; sys.exit(importlib.util.find_spec('{pkg}') is None)",
                 stdout=asyncio.subprocess.DEVNULL,
                 stderr=asyncio.subprocess.DEVNULL,
             )
--- a/tests/executors/test_local.py
+++ b/tests/executors/test_local.py
@@ -13,6 +13,7 @@
 
 """Unit tests for ``LocalCodeExecutor``."""
 
+import asyncio
 import sys
 from pathlib import Path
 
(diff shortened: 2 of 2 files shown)


magic-peach:
Good catch, pushed a fix: both tests now wrap the probe call in asyncio.wait_for with a 5 second timeout, so they fail fast instead of hanging if the subprocess ever blocks again.

sicoyle (approved):
LGTM - thank you for your great contribution!! 🤗

PR #757, opened 2026-09-10
fix(service): surface FastAPIServerBase start failures instead of hanging

magic-peach:
# Description

FastAPIServerBase.start() waits for the server to come up by polling self.server.started in a loop, with no way to exit that loop if the uvicorn server task fails before it ever sets started, for example when the configured port is already in use. The task's exception went unretrieved and start just spun forever instead of raising. This checks server_task.done() on each iteration and calls server_task.result(), so a startup failure is raised immediately instead of hanging.

## Issue reference

No existing issue tracks this; found while reading the FastAPI service startup path.

## Checklist

* [x] Created/updated tests

This is an internal bug fix with no user-facing API or documented behavior change, so no docs PR is needed.

### DIFF
--- a/dapr_agents/service/fastapi/base.py
+++ b/dapr_agents/service/fastapi/base.py
@@ -126,6 +126,9 @@ class FastAPIServerBase(APIServerBase, SignalMixin):
 
             # Wait for startup to complete
             while not self.server.started:
+                if server_task.done():
+                    server_task.result()
+                    return
                 await asyncio.sleep(0.1)
 
             # Extract the real port from the bound socket
--- a/tests/service/__init__.py
+++ b/tests/service/__init__.py
@@ -1,0 +1,12 @@
+#
+# Copyright 2026 The Dapr Authors
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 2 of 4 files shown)


magic-peach:

CasperGN:
@magic-peach could you pls look at the lint failure?
Also: THANK YOU for all the PRs 🚀

dapr_agents/service/fastapi/base.py
```diff
@@ -126,6 +126,9 @@ async def start(self, log_level=None):
 
             # Wait for startup to complete
             while not self.server.started:
+                if server_task.done():
+                    server_task.result()
```
JeffreyJPZ, line 130:
```suggestion
                    # Re-raise if serve() failed (e.g. port in use, startup/lifespan error)
                    server_task.result()
```
nit: could use a small comment here

dapr_agents/service/fastapi/base.py
```diff
@@ -126,6 +126,9 @@ async def start(self, log_level=None):
...
+                    return
```
JeffreyJPZ, line 131: Should we raise here in case it gets a shutdown signal or something and exits before startup completes?

JeffreyJPZ (commented):
Much appreciated @magic-peach, have a few comments here but lgtm overall!

PR #758, opened 2026-09-10
perf(llm): reduce serialization overhead for OpenAI-compatible and HF streaming/non-streaming chat completions

JeffreyJPZ:
# Description

OpenAI-compatible and HuggingFace streaming and non-streaming chat completions now match the Anthropic pattern by using typed attributes on SDK objects directly instead of serializing responses/packets into dicts. This reduces CPU overhead, memory churn, and chunk latency especially for streaming chat completions.

Also:
- Handle LiteLLM separately in response and stream processors. IMO this way is cleaner as the OpenAI processing path now expects `ChatCompletion` and `ChatCompletionChunk` which LiteLLM just doesn't use (even if its "structurally compatible"), you'd need to use `getattr()` in multiple places which bleeds in provider-specific logic and makes stuff a bit harder to reason about.
- Add guard for unsupported streaming provider in stream processor
- Add tests for OpenAI and HuggingFace chat completion, LiteLLM utils tests, stream utils tests
- Move LLM tests to match package structure
- Fix HuggingFace `api_key` normalization bug while writing tests (`api_key` was never set)

## Issue reference

Complements #733 which should be pushed through first

## Checklist

* [x] Created/updated tests


JeffreyJPZ:
This stuff should build on #764

PR #759, opened 2026-09-11
fix(utils): schedule async graceful_shutdown() from signal handler instead of dropping it

magic-peach:
# Description

SignalHandlingMixin._handle_shutdown_signal is invoked from a real signal.signal() OS callback, not the event loop. When graceful_shutdown is a coroutine function, which it is by default since the class's own implementation is `async def graceful_shutdown`, the handler only logged a debug message and never scheduled or awaited it. So the default cleanup path (`await self.stop()`) silently never ran on a real SIGINT/SIGTERM; only the shutdown event got set.

This class is still exported as public API from `dapr_agents.utils` (`SignalHandlingMixin` in `__all__`), even though internal code has moved to the newer `dapr_agents.utils.signal.mixin.SignalMixin`, which already schedules graceful shutdown correctly via `loop.call_soon_threadsafe` + `loop.create_task`. This PR ports the same fix into the older, still-public class: capture the event loop in `setup_signal_handlers`, then use it to properly schedule the coroutine in `_handle_shutdown_signal`.

Had no test coverage before this change.

## Issue reference

No existing issue tracks this; found while reading the signal-handling utilities.

## Checklist

* [x] Created/updated tests

### DIFF
--- a/dapr_agents/utils/signal_mixin.py
+++ b/dapr_agents/utils/signal_mixin.py
@@ -39,6 +39,7 @@ class SignalHandlingMixin:
         super().__init__(*args, **kwargs)
         self._shutdown_event: Optional[asyncio.Event] = None
         self._signal_handlers_setup = False
+        self._loop: Optional[asyncio.AbstractEventLoop] = None
 
     def setup_signal_handlers(self) -> None:
         """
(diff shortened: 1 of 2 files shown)


magic-peach:

CasperGN (approved):

PR #760, opened 2026-09-11
fix(document): stop TextSplitter from silently dropping content at chunk boundaries

magic-peach:
# Description

SplitterBase._merge_splits, the chunk-merging logic behind every TextSplitter call including split_documents (used for RAG document ingestion), silently dropped the split that triggered finalizing the current chunk. When a split didn't fit in the in-progress chunk, the code finalized that chunk and reset to the overlap-seeded new chunk, but never added the triggering split anywhere before the loop moved on to the next item. Any split landing exactly on a chunk-size boundary simply vanished from the output.

Reproduced directly: `splits=['A'*100, 'B'*100, 'C'*3000]` with `max_size=150` produced output missing the `B` chunk entirely.

Fix: move the append outside the size-check branch so the triggering split is always added to the new chunk. This also makes the old "standalone oversized chunk" branch redundant, since an oversized split with an empty current_chunk now naturally becomes its own chunk once appended, so that branch is removed.

Had no test coverage before this change.

## Issue reference

No existing issue tracks this; found while reading the document splitting pipeline.

## Checklist

* [x] Created/updated tests

### DIFF
--- a/dapr_agents/document/splitter/base.py
+++ b/dapr_agents/document/splitter/base.py
@@ -116,43 +116,42 @@ class SplitterBase(BaseModel, ABC):
         for split in splits:
             split_size = self._get_chunk_size(split)
 
-            # If adding the current split exceeds max_size, finalize the current chunk
-            if current_size + split_size > max_size:
-                if current_chunk:
-                    # Finalize the current chunk
-                    full_chunk = "".join(current_chunk)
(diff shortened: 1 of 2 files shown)


magic-peach:

CasperGN (approved):

PR #765, opened 2026-09-13
fix(prompt): stop pre_fill_variables from crashing on StringPromptTemplate

magic-peach:
# Description

PromptTemplateBase.pre_fill_variables always rebuilt the instance with messages=self.messages and template_format=self.template_format, but messages only exists on ChatPromptTemplate. Calling pre_fill_variables on a StringPromptTemplate raised AttributeError. This is a reachable path: agents/prompting.py calls pre_fill_variables on the configured prompt template whenever instructions or style_guidelines are set, regardless of which template type is in use.

Switched to model_copy(update=...), which carries over every other field as-is regardless of subclass, so it works for any PromptTemplateBase subclass instead of only ones shaped like ChatPromptTemplate, and does so without re-validating fields like messages.

Fixing that surfaced a second bug right behind it: StringPromptTemplate.format_prompt checked for missing variables against the raw kwargs passed to the call, before merging in pre_filled_variables, so a variable that had actually been pre-filled was still reported as missing unless it was passed again. Merge first, then validate against the merged set, matching how ChatPromptTemplate.format_prompt already does it.

## Issue reference

No existing issue tracks this; found while reading the prompt template base class.

## Checklist

* [x] Created/updated tests


magic-peach:

CasperGN (approved):

PR #766, opened 2026-09-13
fix(storage): stop add_documents dropping metadata for later documents

magic-peach:
# Description

VectorStoreBase.add_documents decided whether to forward metadata to add() by checking only documents[0].metadata. If the first document had no metadata but a later one did, the whole batch's metadata was silently dropped instead of just that first document's. Now checks whether any document in the batch has metadata.

Added tests/storage/vectorstores/test_base.py covering the batch-metadata decision, including the case that previously dropped metadata.

## Issue reference

No existing issue tracks this; found while reviewing the vector store base class.

## Checklist

* [x] Created/updated tests


dapr_agents/storage/vectorstores/base.py
```diff
@@ -121,7 +121,9 @@ def add_documents(self, documents: List[Document]) -> List[str]:
         """
         texts = [doc.text for doc in documents]
         metadatas = (
-            [doc.metadata for doc in documents] if documents[0].metadata else None
+            [doc.metadata for doc in documents]
+            if any(doc.metadata for doc in documents)
+            else None
```
JeffreyJPZ:
```suggestion
            metas
            if any(metas := [doc.metadata if doc.metadata else {} for doc in documents])
            else None
```

Just a tiny fix - we should default to empty dicts as this is what `add` expects (I believe the vector stores we support either raise on null values or discourage it).

JeffreyJPZ (changes requested):
Thanks @magic-peach! One small thing from me.

magic-peach:
@JeffreyJPZ I made the changes

CasperGN (approved):

PR #778, opened 2026-09-16
fix(tool): stop infer_func_schema naming the model after the last parameter

magic-peach:
# Description

`ToolHelper.infer_func_schema` iterated `sig.parameters.items()` using `name` as its loop variable, shadowing the method's own `name: Optional[str] = None` argument. For any function with parameters, the generated Pydantic args model ended up named after the function's last parameter instead of the caller-supplied name or the `f"{func.__name__}Model"` default. This is used by every `AgentTool.from_func` call, so every tool built from a plain function got a mis-named schema.

Fixed by renaming the loop variable to `param_name` so it no longer collides with the `name` argument.

## Issue reference

Please reference the issue this PR closes: N/A, self-discovered while reading the tool schema inference code, no existing issue filed.

## Checklist

* [x] Created/updated tests


CasperGN (approved):

magic-peach:
@CasperGN I would love to contribute more to the dapr-agents and python-sdk repo. Asking out of curiosity what is the process to become a member of the repo?

